PR #2, opened 2017-06-14
Feature/Experimenter - Manage Studies

pattisdr:
# Purpose

Add manage studies page

# TODO
- [x]  Adapt html from original experimenter studies page
- [x] Add font-awesome
- [x]  Add filtering on study state
- [x] Highlight study state if active
- [x] Add filtering in search bar on submit - needs to filter title and description
- [x] Match state filters to filters in db
- [x] add dateModified to study model
- [x] Change "blocks" to "structure" on study model
- [x] Add exit_url and display_full_screen fields to study model.
- [x] Make beginDate and endDate properties on the model, pulling from study logs
- [x] Create migration for study model
- [x] Add sorting capabilities
- [x] Figure out sorting on property if null values in list
- [x] Figure out how to fetch "my studies"
- [x] Add completed field to response model + migration
- [x] Handle multiple query parameters simultaneously - register a templateTag to help with this?
- [x] How to calculate completed responses versus incomplete responses
- [x] Prevent form submission from clearing text box

studies experimenter


exp/views/study.py
```diff
@@ -29,9 +31,40 @@ class StudyListView(LoginRequiredMixin, generic.ListView):
...
+        match = request.get('match')
+        if match:
+            queryset = queryset.filter(Q(name__icontains=match) | Q(short_description__icontains=match))
+
+        sort = request.get('sort')
+        if sort:
+            if 'name' in sort:
+                queryset = queryset.order_by(sort)
+            elif 'beginDate' in sort:
+                queryset = sorted(queryset, key=lambda t: t.begin_date or timezone.now(), reverse=True if '-' in sort else False)
```
cwisecarver: This is probably fine for now. Once this list gets long it's going to use a lot of memory and be very slow. Sorting outside of SQL is a bit of an anti-pattern as it's going to evaluate the entire queryset in order to sort it. There's some SQL magic we can do that would allow you to find the beginning date and the end date in a subquery and then sort by that all inside of the database. I might add a TODO to optimize this and the one below it using a subquery or an annotation.

studies/models.py
```diff
@@ -72,6 +76,32 @@ class Meta:
             ('can_view_demographics', 'Can View Demographics'),
         )
 
+    @property
+    def begin_date(self):
+        active_logs = self.logs.filter(action='active')
+        if active_logs:
```
cwisecarver: I could be misremembering but I think querysets are always truthy. I think you want to do `if active_logs.exists():`

exp/views/study.py
```diff
@@ -29,9 +31,40 @@ class StudyListView(LoginRequiredMixin, generic.ListView):
...
+            elif 'endDate' in sort:
+                queryset = sorted(queryset, key=lambda t: t.end_date or timezone.now(), reverse=True if '-' in sort else False)
+
+        return queryset
+
+    def get_context_data(self, **kwargs):
+        context = super(StudyListView, self).get_context_data(**kwargs)
```
cwisecarver: In python3 you can do `super().get_context_data(**kwargs)` without passing the class and self. I just found this out the other day.

exp/views/study.py
```diff
@@ -29,9 +31,40 @@ class StudyListView(LoginRequiredMixin, generic.ListView):
...
+        context['state'] = self.request.GET.get('state') or 'all'
```
cwisecarver: `self.request.GET.get('state', 'all')` I think `get` takes a default value.

studies/models.py
```diff
@@ -72,6 +76,32 @@ class Meta:
...
+        begin_date = self.begin_date
+        deactivated_logs = self.logs.filter(action='deactivated')
+        if deactivated_logs and begin_date:
+            end_date = deactivated_logs.latest('action').created_at
+        else:
+            return None;
+        return end_date if end_date > begin_date else None
+
+    @property
+    def completed_responses(self):
```
cwisecarver: I might rename this to `completed_responses_count` or `num_completed_responses` so it doesn't sound like it's going to actually return the responses. Same with the one below.

studies/models.py
```diff
@@ -72,6 +76,32 @@ class Meta:
...
+        if active_logs:
+            return active_logs.latest('action').created_at
+        else:
+            return None
+
+    @property
+    def end_date(self):
+        begin_date = self.begin_date
+        deactivated_logs = self.logs.filter(action='deactivated')
+        if deactivated_logs and begin_date:
```
cwisecarver: I think this might need an `.exists()` too.

exp/templatetags/exp_extras.py
```diff
@@ -0,0 +1,22 @@
...
+    state = kwargs.get('state')
+    if state:
+        updated['state'] = state
+
+    if kwargs.get('match'):
+        updated['match'] = kwargs.get('match')
+
+    if kwargs.get('sort'):
+        sort_value = kwargs.get('sort')
+        previous_sort_value = request.GET.get('sort') or ''
```
cwisecarver: I think `.get('sort', '')` should work.

pattisdr:
Thanks for the review @cwisecarver!

cwisecarver:
Enjoy your weekend @pattisdr. We'll pick it up on Monday.

pattisdr:
Changes made @cwisecarver

PR #3, opened 2017-06-15
Fix signup and demo-data creation process

cwisecarver:


pattisdr:
@cwisecarver, maybe unrelated, have you gotten a JSONDecodeError at /admin/accounts/demographicdata/add/ when trying to add demographic-data to a user in the admin app? I was wondering if the formatting of child_birthdays might have something to do with it.

cwisecarver:
@pattisdr Yeah, it has to be `1980-02-23, 1980-02-18,`

accounts/models.py
```diff
@@ -319,7 +319,7 @@ class DemographicData(models.Model):
         ('rural', _('rural')),
     )
     user = models.ForeignKey(
-        User, on_delete=models.CASCADE,
+        User, on_delete=models.CASCADE, null=True,
```
pattisdr: What does null=True here do?

cwisecarver: I added that because I couldn't figure out how to get the active user and set it on the form prior to saving the demographic data. So it gets saved once and then I add the user when I have it in scope and save it again. Not the way I wanted to do it, but it's done.

cwisecarver:
Fixed the styling to match the rest of the site and and removed the clean method on the form.

PR #4, opened 2017-06-15
Revert to basic DRF JSON API

cwisecarver:
This also adds routes for demo-data and makes studies able to be looked up by their UUID.


cwisecarver:
Just realized that pk ids were being exposed while uuids were still appearing in urls... I'll fix and resubmit.

PR #5, opened 2017-06-19
Feature/Researchers List, Detail, Create

pattisdr:
# Purpose

Add to Researcher List, Detail, and Create Pages so they match the mockups

# TODO
- [x] Move over base html for three pages
- [x] Get X-editable working for researcher detail page
- [x]  In navbar on researcher-detail, user is current user selected, not logged-in user
- [x] When creating a researcher, it is not created with the same organization as the logged in user
- [x]  Add breadcrumbs to create researcher page
- [x] Hook up sort by name alphabetically
- [x] Hook up filtering on name
- [x] Add delete researcher confirmation modal?
- [x] Creating researcher with email already in db
- [x] Display permissions on researchers list page (what permissions are we showing, and how to access with django guardian??
- [x] Enable editing permissions
- [x] Figure out how to display user groups (for permissions)
- [x] Check permissions before researcher delete
- [x] Check permissions before researcher update
- [x] Check permissions before researcher create
- [ ] Checks before deleting a researcher - user was associated with logs.
- [] Delete researcher behavior not very proper.  (on hold)
- []Send password reset email (on hold)
- [] resend confirmation email (on hold)

screen shot 2017-06-21 at 12 42 13 am

screen shot 2017-06-20 at 10 53 07 pm
screen shot 2017-06-21 at 12 42 27 am

screen shot 2017-06-20 at 10 54 40 pm


exp/views/user.py
```diff
@@ -147,20 +207,28 @@ class ResearcherCreateView(LoginRequiredMixin, generic.CreateView):
...
-        # implicitly add them to their creator's organization
-        query_dict.update(is_active=True, is_superuser=False, is_staff=False, password=self.user_password, organization=self.request.user.organization)
+        query_dict.update(is_active=True, is_superuser=False, is_staff=False, password=self.user_password)
         form.data = query_dict
         if form.is_valid():
+            # implicitly add them to their creator's organization
+            form.instance.organization = self.request.user.organization
             return self.form_valid(form)
         else:
+            self.object = None
```
cwisecarver: What does this do?

exp/templates/exp/_navigation.html
```diff
@@ -12,23 +12,23 @@
...
+        {% endif %} {# /if request.user.is_superuser #}
         <li role="separator" class="divider"></li>
         <li class="dropdown">
             <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
-                {{ user.identicon_small_html }} {{ user.get_short_name }} <span class="caret"></span>
+                {{ request.user.identicon_small_html }} {{ request.user.get_short_name }} <span class="caret"></span>
             </a>
             <ul class="dropdown-menu">
                 <li><a href="#">OSF Profile</a></li>
                 <li><a href="{% url 'logout' %}">Logout</a></li>
```
cwisecarver: I know you didn't do this, but after pulling this branch and loading experimenter I get a NoReverseMatch trying to render any page with this partial template in it. Changing it to `'web:logout'` makes it work fine.

exp/views/user.py
pattisdr: This was for when there was an error in the researcher create form.

Adding to the code you already had here, if you used an email that had been used before then `self.form_invalid` would get called.  That method then called `self.get_context_data` which called  `self.object`.  This would fail because there was no object, I think because none had been created?

exp/views/user.py
```diff
@@ -70,10 +76,37 @@ class ResearcherListView(LoginRequiredMixin, generic.ListView):
...
+        if match:
+            queryset = queryset.filter(Q(family_name__icontains=match) | Q(given_name__icontains=match)| Q(middle_name__icontains=match))
+        sort = self.request.GET.get('sort')
+        if sort:
+            if 'family_name' in sort:
+                queryset = queryset.order_by(Lower('family_name').desc()) if '-' in sort else queryset.order_by(Lower('family_name').asc())
+        return queryset
+
+    def post(self, request, *args, **kwargs):
+        retval = super(ResearcherListView, self).get(request, *args, **kwargs)
```
pattisdr: This is a ListView so there isn't a post method on it.  It's a list view, but the mockups have the delete button on this page. If I add a deleteview to this, then I get  `object has no attribute 'object'` error. i needed to return something here.

exp/views/user.py
```diff
@@ -70,10 +76,37 @@ class ResearcherListView(LoginRequiredMixin, generic.ListView):
...
+        if 'delete' in self.request.POST:
```
pattisdr: Does delete work with an html form?

pattisdr:
@cwisecarver, responded to everything I can.  The things that have not changed are how the DELETE behavior works.  I am having trouble with deleting things on a ListView.  If I add a delete method to the form instead of post, the delete method is never called. Do html forms have `delete`?

cwisecarver:
@pattisdr You're right. No DELETE on a form because DELETE can't have a request body. Sorry for the misdirection.

pattisdr:
@cwisecarver I am not sure how to handle deleting a researcher where an IntegrityError is raised, for example, deleting a researcher whose id is referenced elsewhere, like in study logs.  I am trying to catch the exception and throw an error, but "catch" isn't getting hit

```
try:
    User.objects.get(pk=self.request.POST['delete']).delete()
except IntegrityError as e:
     ...
```

cwisecarver:
Maybe don't delete them but instead mark them as `is_active = False`. This will prevent them from logging in. Then maybe display this in the manage org UI? @pattisdr

cwisecarver:
If you really want to delete them, you'd have to change the on_delete attribute of all the fields referencing them to [DO_NOTHING](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ForeignKey.on_delete).

cwisecarver:
But deleting things is scary.

cwisecarver:
By DO_NOTHING, I mean SET_NULL...

pattisdr:
Yes, I guess that was more my question, should we even be deleting them at all.  Marking as inactive seems safer.

PR #8, opened 2017-06-28
JSON API with full relationships

cwisecarver:
- UUIDS appear in place of primary keys
- Nested urls e.g. `/api/v1/users/{user-id}/children/`
- Relationship urls
- Serializers and views for missing models


pattisdr:
Hooray! Will look at this shortly.

cwisecarver:
Thanks! I started to write tests for the nested filtering `/api/v1/users/{user-id}/children/`, etc. but I couldn't get the APIClient in DRF to work in what I thought was a respectable amount of time. I functionally tested it and did find some bugs that I fixed so I might give that an extra look. Thanks!

api/renderers.py
```diff
@@ -6,16 +6,16 @@
...
                                         Serializer)
 from rest_framework.settings import api_settings
 
 import inflection
-from rest_framework_json_api import renderers, utils
+from rest_framework_json_api import utils
 
 
-class JSONRenderer(renderers.JSONRenderer):
+class JSONAPIRenderer(renderers.JSONRenderer):
```
pattisdr: around [L164](https://github.com/CenterForOpenScience/lookit-api/blob/e8eb8e7093d94b33c5a341eccd290beddc0dfdb2/api/renderers.py#L162) 
```
relation_data = {
       'data': resource.get(field_name)
}
```
- Do we need to be including the id/type of every related resource?  These items aren't paginated, so if you have 200 responses to a study, the type/id of 200 responses will be included in the study > responses relationship. Relationships aren't required to have a "data" key.
screen shot 2017-06-29 at 3 16 53 pm

pattisdr: Should [meta > count ](https://github.com/CenterForOpenScience/lookit-api/blob/e8eb8e7093d94b33c5a341eccd290beddc0dfdb2/api/renderers.py#L174)information have parallel structure?

screen shot 2017-06-29 at 3 19 25 pm

studies/serializers.py
```diff
@@ -1,10 +1,90 @@
...
-
-class StudySerializer(serializers.ModelSerializer):
+class StudySerializer(UUIDSerializerMixin, ModelSerializer):
+    url = serializers.HyperlinkedIdentityField(
+        view_name='study-detail',
+        lookup_field='uuid'
+    )
+    organization = UUIDResourceRelatedField(
+        queryset=Organization.objects,
+        many=False,
```
cwisecarver: Yeah, there's only one organization per study. It's a FK. There's no documented way to do this in drf-jsonapi, at least not that I can find. It is a list of one. Changing it from a list view to a detail view didn't work because the detail view expects you to pass the UUID in the url, which it doesn't do.

studies/serializers.py
```diff
@@ -1,10 +1,90 @@
...
+
+class ResponseSerializer(UUIDSerializerMixin, ModelSerializer):
+    url = serializers.HyperlinkedIdentityField(
+        view_name='response-detail',
+        lookup_field='uuid'
+    )
+    study = UUIDResourceRelatedField(
+        queryset=Study.objects,
+        many=False,
+        related_link_view_name='response-studies-list',
```
cwisecarver: Yeah, again, detail requires a uuid to be passed to it, `responses/resp_id/studies/study_id/`.

api/views.py
```diff
@@ -1,25 +1,74 @@
...
+    resource_name = 'demographics'
+    queryset = DemographicData.objects.filter(user__is_active=True)
+    serializer_class = DemographicDataSerializer
+    filter_fields = [('user', 'user'), ]
+
+
+class UserViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     lookup_field = 'uuid'
     resource_name = 'users'
     queryset = User.objects.filter(demographics__isnull=False)
```
cwisecarver: It returns all users that have demographics that are marked is_active = True. I think we decided that when oauth is implemented we can flag researchers with a boolean. I don't know why we would need access to the researchers in the api but I could be wrong.

All organizations do share the same participant pool, participants do not belong to an organization.

accounts/serializers.py
```diff
@@ -30,51 +47,79 @@ class Meta:
...
-    included_serializers = {
-        'user': UserSerializer,
-    }
+    # included_serializers = {
+    #     'user': UserSerializer,
+    # }
+    user = UUIDResourceRelatedField(
+        queryset=User.objects,
+        many=False,
+        related_link_view_name='child-users-list',
```
cwisecarver: I can probably do `users/user_id` for all of these if that works for you. I think it makes more sense anyway since there's only one.

cwisecarver:
@pattisdr I'm pretty sure, one way or another, I've fixed everything you noticed in here. I'm going to merge it and if there are other things that we find I will fix them in a separate PR.

PR #10, opened 2017-06-28
[WIP] Celery experiment deployment

cwisecarver:


project/settings/defaults.py
```diff
@@ -190,7 +204,7 @@
 
 LANGUAGE_CODE = 'en-us'
 
-TIME_ZONE = 'UTC'
+TIME_ZONE = 'America/New_York'
```
cwisecarver: This is the [default "display" timezone](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-TIME_ZONE). Everything is always stored in UTC in the database. Django just uses this setting to change times before they're shown to the user unless the user has a custom timezone set. We're pulling one over from the API for them if they login through oauth. I will double check that we're actually using that to show the user times in their local tz.

project/settings/defaults.py
```diff
@@ -207,7 +221,7 @@
 # https://docs.djangoproject.com/en/1.9/howto/static-files/
 
 # base url for experiments, should be s3 bucket in prod
-EXPERIMENT_BASE_URL = os.environ.get('EXPERIMENT_BASE_URL', 'http://google.com/')  # default to ember base url
+EXPERIMENT_BASE_URL = os.environ.get('EXPERIMENT_BASE_URL', 'https://storage.googleapis.com/io-osf-lookit-staging2/experiments/')  # default to ember base url
```
cwisecarver: I agree. I'll kill it before merging. It was mostly so I could test it locally.

project/settings/defaults.py
```diff
@@ -283,3 +300,5 @@
     messages.WARNING: 'alert-warning',
     messages.ERROR: 'alert-danger',
 }
+EMBER_EXP_PLAYER_REPO = 'https://github.com/pattisdr/ember-frame-player'
```
cwisecarver: @pattisdr Would you work with @icereval to transfer your ember-frame-player repo to COS?

web/views.py
```diff
@@ -333,4 +331,9 @@ def dispatch(self, request, path, *args, **kwargs):
             # requesting user doesn't belong to that child
             raise PermissionDenied()
 
+        request.META['HTTP_X-EXPERIMENT-CHILD-ID'] = path.split('/')[1]
```
cwisecarver: I need to delete that. I forgot for a minute that ember could actually look at the URL in the browser and was trying to find a way to pass the child id to the app, but it already had it.

project/settings/defaults.py
```diff
@@ -12,20 +12,23 @@
...
+EMBER_BUILD_ROOT_PATH = os.path.join(BASE_DIR, '../ember_build')
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.9/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = 'c*si4oji5r=#)5+6e$bi@an%v)(n2be2+1(s985uh4mu3jq!qt'
 
-# SECURITY WARNING: don't run with debug turned on in production!
+    # SECURITY WARNING: don't run with debug turned on in production!
```
cwisecarver: It's for good luck

studies/tasks.py
```diff
@@ -0,0 +1,188 @@
...
+    print(f'Downloading {addons_zip_path}...')
+    unzip_file(requests.get(addons_zip_path).content, os.path.join(local_repo_destination_folder, 'lib'))
+
+    return repo_destination_folder
+
+
+def build_docker_image():
+    # this is broken out so that it can be more complicated if it needs to be
+    print(f'Running docker build...')
+    subprocess.run(['docker', 'build', '-t', 'ember_build', '.'], cwd=settings.EMBER_BUILD_ROOT_PATH)
```
cwisecarver: I'll add the logging. I think we can do this in the study log model now.

web/views.py
cwisecarver: deleted

PR #17, opened 2017-07-10
Feature/Participant List/Detail, Response Pages [LEI-422][LEI-423]

pattisdr:
https://openscience.atlassian.net/browse/LEI-422
https://openscience.atlassian.net/browse/LEI-423

Wrapping up @hmoco's work on these pages...there are several TODO's that have been reticketed in JIRA.

- [X] Participant List
- [x] Participant Detail
- [x] Session Data
- [x] All Session Data
- [x] All Attachments

participants experimenter

screencapture-127-0-0-1-8000-exp-participants-8-1499873563163
screencapture-127-0-0-1-8000-exp-studies-14-responses-1499873607392
screencapture-127-0-0-1-8000-exp-studies-14-responses-1499873623991
screencapture-127-0-0-1-8000-exp-studies-14-responses-1499873642944


accounts/models.py
```diff
@@ -1,6 +1,7 @@
 import base64
 import hashlib
 import uuid
+from datetime import date
```
cwisecarver: `timezone.now().date`

exp/views/study.py
```diff
@@ -89,10 +93,6 @@ def get_queryset(self, *args, **kwargs):
                 # TODO optimize using subquery
                 queryset = sorted(queryset, key=lambda t: t.end_date or timezone.now(), reverse=True if '-' in sort else False)
```
cwisecarver: There's a way to do this with annotations, I was working on it before I got distracted. Something like this:
```
Study.objects.all().annotate(begin_date=models.Max(models.Case(models.When(logs__action='deactivate', then='logs__created_at'), default='logs__created_at', output_field=models.DateTimeField()))).values('begin_date')
```

or 

```
Study.objects.all().annotate(begin_date=models.Subquery(StudyLog.objects.filter(action='activate', study=models.OuterRef('pk')).values('created_at').annotate(begin_date=models.Max('created_at')).values('begin_date'), output_field=models.DateTimeField()))
```

accounts/templates/accounts/participant_detail.html
```diff
@@ -1,42 +1,289 @@
...
+                            <div class="col-xs-7">
+                                 {{child.age}}
+                            </div>
+                        </div>
+                        <div class="row">
+                            <div class="col-xs-5">
+                                <b> Extra: </b>
+                            </div>
+                            <div class="col-xs-7">
+                                 {{child.extra}}
```
pattisdr: This "extra" field is incorrectly named.  It's actually the additional_information field on the child which is a string, and was renamed to "extra" in get_context_data.  We don't need the <pre> here

pattisdr:
CR comments addressed in https://github.com/CenterForOpenScience/lookit-api/pull/23

exp/views/study.py
pattisdr: Putting aside for now, follow-up ticket in JIRA created https://openscience.atlassian.net/browse/LEI-434

PR #19, opened 2017-07-14
Add docker, update wsgi & requirements to match

binoculars:
Docker for deployment

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,63 @@
+FROM python:3.6.1-slim
+
+RUN apt-get update \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+        ca-certificates \
+        gcc \
+        git \
+        libev4 \
+        libev-dev \
+        libevent-dev \
+        libxml2-dev \
+        libxslt1-dev \
+        libffi-dev \
+        # psycopg2
```
cwisecarver: This is a python library

PR #22, opened 2017-07-17
Fix Response POST requests

cwisecarver:
- Create Writeable serializer for Response objects
    - Use the ids instead of the actual objects so DRF doesn't try to create things.
- Swap in that serializer when creating responses
- Grab the latest demographic data off the specified child's user and use that id

### DIFF
--- a/studies/serializers.py
+++ b/studies/serializers.py
@@ -81,7 +81,7 @@ class ResponseSerializer(UUIDSerializerMixin, ModelSerializer):
         many=False,
         related_link_view_name='demographicdata-detail',
         related_link_lookup_field='uuid',
-        required = False
+        required=False
     )
 
     class Meta:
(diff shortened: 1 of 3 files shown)


studies/serializers.py
```diff
@@ -98,3 +98,18 @@ class Meta:
...
+class ResponseWriteableSerializer(ResponseSerializer):
+    def create(self, validated_data):
+        """
+        Use the ids for objects so django rest framework doesn't
+        try to create new objects out of spite
+        """
+        study = validated_data.pop('study')
+        validated_data['study_id'] = study.id
+        # implicitly set the demographic data because we know what it will be
+        validated_data['demographic_snapshot_id'] = validated_data.get('child').user.latest_demographics.id
```
pattisdr: `latest_demographics` (outside of this PR) might be pulling the first demographics info instead of the last?

pattisdr:
Looking good to me @cwisecarver - can do POST /responses/ without having demographic_data relationship in request body, and then it is added.  Just the one comment about latest_demographics.

studies/serializers.py
cwisecarver: I double check that and fix it if it is. Judging from the name of the property it should always be the one with the largest datetime.

cwisecarver:
I fixed the latest thing, you were right @pattisdr. It was basically a double-negative. Default ordering was -created and last got the oldest not the newest.

PR #26, opened 2017-07-20
Feature/Build Study [LEI-421]

pattisdr:
# Ticket 
https://openscience.atlassian.net/browse/LEI-421

# Purpose
Adds page to edit study structure (json block describing frames and sequence).  Gets adding json to study on Create Study working. Validates json.  Makes sure display of json is in correct format. Proxies preview route to ember app to preview study.


pattisdr:
@cwisecarver, started connecting the `preview` route - proxying `http://127.0.0.1:8000/exp/studies/<study_uuid>/preview` to the ember app.  Just the study id needs to be sent, not a child here.

With this PR, I am getting `dispatch() missing 1 required positional argument: 'path'`.  It works if I move the exp urls last in `project/urls.py`, but then all other experimenter routes will redirect to the ember app and return not found.  Any ideas since you're more familiar with this before I do more digging?

exp/urls.py
```diff
@@ -41,5 +42,7 @@
     url(r'studies/create/$', StudyCreateView.as_view(), name='study-create'),
     url(r'studies/(?P<pk>\d+)/$', StudyDetailView.as_view(), name='study-detail'),
     url(r'studies/(?P<pk>\d+)/edit/$', StudyUpdateView.as_view(), name='study-edit'),
+    url(r'studies/(?P<pk>\d+)/edit/build/$', StudyBuildView.as_view(), name='study-build'),
+    url(r'^studies/(?P<study_uuid>[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89ab][0-9a-fA-F]{3}-[0-9a-fA-F]{12})/preview/?$', PreviewProxyView.as_view(), name='preview-proxy'),
```
cwisecarver: The `ProxyView` wants a path kwarg in the URL. If you look at the [one for taking the experiment](https://github.com/CenterForOpenScience/lookit-api/blob/develop/web/urls.py#L16) it wraps the entire url after `/studies/` in a named capture group called path. I think the other two kwargs, study_uuid and child_uuid are also available but I'm not sure.

Also, I'm not sure that you need the `^`. That indicates the beginning of a line and I can't remember if django's URL inheritance fixes that for you or not, just that the others above this one are working without it.

Make sense? If adding path in here gives you trouble let me know and I'll pull this down and step through it with ipdb.

exp/views/study.py
```diff
@@ -276,3 +300,10 @@ def get_context_data(self, **kwargs):
             }, indent=4) for resp in context['responses']]
         context['all_responses'] = ', '.join(context['response_data'])
         return context
+
+
+class PreviewProxyView(ProxyView, LoginRequiredMixin):
+    upstream = settings.EXPERIMENT_BASE_URL
+
+    def dispatch(self, request, path, *args, **kwargs):
```
cwisecarver: If you're not doing anything but calling super you can just remove this method and it will inherit from the parent.

pattisdr: I didn't have it there at first, but put it in to avoid `dispatch() got an unexpected keyword argument 'uuid'` errors.

cwisecarver: That makes sense. :+1:

web/templates/web/study-detail.html
```diff
@@ -36,6 +37,11 @@
         function childSelected(selected) {
             var participateButton = document.getElementById('participate-button');
             selected.value === 'none' ? participateButton.disabled = true : participateButton.disabled = false;
+            participateButton.value = selected.value;
+        }
+
+        function portToParticipate(selected) {
```
cwisecarver: Might be worth doing this in a `window.open` so they can preview while having their experiment in another tab for editing but just an idea.

pattisdr: i changed the preview route so it opens in a new tab. this was for the participate route - not sure if we should change that one or not. the player is supposed to have things to allow you to exit early and then it directs you back to the exitUrl that's defined in the study.

cwisecarver: My mistake. I thought this was the preview for some reason.

exp/urls.py
pattisdr: thanks i missed the <path> in the URL!

PR #29, opened 2017-07-31
[RFR] Oauth2 login through OSF

cwisecarver:
This integrates oauth2 login into experimenter.

- [ ] Go to http://localhost:8000/accounts/login/
- [ ] Click on Open Science Framework
- [ ] Login with a staging account
- [ ] You should then have a user created with
    - [ ] Your username as your email address
    - [ ] Your first and last name populated
    - [ ] A SocialAccount instance with your full API metadata in one of the extra fields
- [ ] If you login in a different browser/session with an OrgAdmin account
    - [ ] Go to manage organization in experimenter
    - [ ] See your other account and it says you have no permissions
    - [ ] Edit your other account and set yourself an permission of some sort
    - [ ] Your user should be assigned to the correct organization and have the correct permissions

@pattisdr Not sure about the super-secret second login page but ... it'll work for now. Also, it's a little ugly but it does seem to do the job.

# TODO I need to remove that oauth token from staging when we're done.


osf_oauth2_adapter/views.py
```diff
@@ -0,0 +1,76 @@
...
+        extra_data = requests.get(self.profile_url, headers={
+            'Authorization': 'Bearer {}'.format(access_token.token)
+        })
+        return self.get_provider().sociallogin_from_response(
+            request,
+            extra_data.json()
+        )
+
+oauth2_login = OAuth2LoginView.adapter_view(OSFOAuth2Adapter)
+oauth2_callback = OAuth2CallbackView.adapter_view(OSFOAuth2Adapter)
```
pattisdr: I don't have an exact series of steps, but I kept getting myself into a state where I couldn't login, it was like the credentials were not getting passed back to the app. I would click on "Open Science Framework" to login, but would be directed to a CAS screen saying I was logged into the OSF, but wouldn't direct me back to Lookit.  I would still be logged out of Lookit.  Can't duplicate it at the moment, but this happened to me three times today.

EDIT:  I think this is happening when I am using `http://127.0.0.1:8000/` instead of localhost:8000/

project/urls.py
```diff
@@ -24,10 +24,14 @@
...
 from project import settings
 from web import urls as web_urls
+from osf_oauth2_adapter import views as osf_oauth2_adapter_views
 
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
     url(r'^api/', include(api_urls)),
+    url(r'^accounts/social/login/cancelled/', osf_oauth2_adapter_views.login_errored_cancelled),
+    url(r'^accounts/social/login/error/', osf_oauth2_adapter_views.login_errored_cancelled),
+    url(r'^accounts/', include('allauth.urls')),
```
pattisdr: LoginRequiredMixin is on almost all views except maybe Study List / Study Detail Lookit-Side.  If you're experimenter side, should you be redirected to `accounts/login` and if you're lookit-side, be redirected to `login`.  Right now everyone is getting directed to the participant login which won't work for researchers.

exp/views/user.py
```diff
@@ -101,7 +101,7 @@ class ResearcherListView(LoginRequiredMixin, DjangoPermissionRequiredMixin, gene
     '''
     template_name = 'accounts/researcher_list.html'
     # TODO needs to change once oauth in
```
pattisdr: Can remove this.

exp/views/user.py
```diff
@@ -101,7 +101,7 @@ class ResearcherListView(LoginRequiredMixin, DjangoPermissionRequiredMixin, gene
...
-    queryset = User.objects.filter(demographics__isnull=True, is_active=True)
```
pattisdr: Speaking of "demographics__isnull" which is not really relevant for distinguishing between participant/researcher any longer, some of the querysets in the API are still using this filter.

project/urls.py
```diff
@@ -24,10 +24,14 @@
 from exp import urls as exp_urls
 from project import settings
 from web import urls as web_urls
+from osf_oauth2_adapter import views as osf_oauth2_adapter_views
 
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
     url(r'^api/', include(api_urls)),
+    url(r'^accounts/social/login/cancelled/', osf_oauth2_adapter_views.login_errored_cancelled),
+    url(r'^accounts/social/login/error/', osf_oauth2_adapter_views.login_errored_cancelled),
```
pattisdr: No templates for these URL's?

osf_oauth2_adapter/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class LoginErroredCancelledView(TemplateView):
+    template_name = ("allauth/login_errored_cancelled.html")
```
pattisdr: where is this?

osf_oauth2_adapter/views.py
```diff
@@ -0,0 +1,76 @@
...
+        already exists, is not a responsibility.
+        """
+        username = data.get('username')
+        first_name = data.get('first_name')
+        last_name = data.get('last_name')
+        email = username
+        name = data.get('name')
+        time_zone = data.get('time_zone')
+        locale = data.get('locale')
+        gravatar = data.get('profile_image_url')
```
pattisdr: should we use this instead of the identicon for a researcher in experimenter?

osf_oauth2_adapter/provider.py
```diff
@@ -0,0 +1,91 @@
...
+        :return: A populated instance of the `SocialLogin` model (unsaved).
+        """
+        # NOTE: Avoid loading models at top due to registry boot...
+        from allauth.socialaccount.models import SocialLogin, SocialAccount
+
+        adapter = get_adapter(request)
+        uid = self.extract_uid(response)
+        extra_data = self.extract_extra_data(response)
+        common_fields = self.extract_common_fields(response)
+        socialaccount = SocialAccount(extra_data=extra_data,
```
pattisdr: Can you explain to me the purpose of this SocialAccount model? Looks like it links back to the User object and has an extra field with a lot of the data from the API. What are you doing with this?

accounts/templates/account/login.html
```diff
@@ -0,0 +1,32 @@
+{% extends "exp/base.html" %}
+
+{% load i18n %}
+{% load account socialaccount %}
+{% load bootstrap3 %}
+
+{% block head_title %}{% trans "Sign In" %}{% endblock %}
```
pattisdr: What are we going to do eventually about this secret login page? First, I think experimenter should have its own landing page instead of defaulting to Manage Studies (since this page has specific permissions). 

Then, Experimenter pages have an experimenter login button which takes them to experimenter login (through oauth), Lookit pages have Lookit login button. If you're logged out on experimenter page you get experimenter login. Logged out on Lookit page get Lookit login.

If researcher logs in first time through oauth, he is redirected to new Experimenter landing page instead of leaving him in lookit land. Thoughts?

pattisdr:
Also, @cwisecarver there is a OSF Profile link in the dropdown top-right when you click on your logged in user. That could probably be hooked up now with the info you're pulling from the API.

pattisdr:
And maybe we can use the timezone information you're getting from the API to set the times displayed (logs for example) in the right timezone!

osf_oauth2_adapter/migrations/0002_application.py
```diff
@@ -0,0 +1,41 @@
...
+            client_id=os.environ.get('OSF_OAUTH_CLIENT_ID', '3518b74e12584abf9e48565ff6aee6f3'),
+            secret=os.environ.get('OSF_OAUTH_SECRET', 'vYlku3raTL5DnHZlkqCIaShmPVIl1nifsFJCNLxU'),
+        )
+        app.sites.add(Site.objects.get(domain='osf.io'))
+
+
+def create_site(apps, schema_editor):
+    Site = apps.get_model('sites', 'site')
+    if not Site.objects.filter(domain='osf.io').exists():
+        Site.objects.create(domain='osf.io', name='OSF')
```
cwisecarver: It's because our SITE_ID in defaults.py is `1`, which is example.com, because it gets created first. I'll change this migration to make a lookit site and change our SITE_ID to be `2`. That should fix this initial fail.

accounts/templates/account/login.html
cwisecarver: I think that makes sense. There's a view at `/exp/` that just redirects to `/exp/studies/`. That could be changed to conditionally redirect based on the user having permissions and then toss a template in that says you've not got any permissions and you'll be notified when someone gives you some. Make sense?

accounts/templates/account/login.html
```diff
@@ -0,0 +1,32 @@
...
+      {% get_providers as socialaccount_providers %}
+
+      {% if socialaccount_providers %}
+      <p>{% blocktrans with site.name as site_name %}Please sign in to the Open Science Framework or
+      <a href="https://osf.io/register">register</a>
+      for an account and sign in below:{% endblocktrans %}</p>
+
+      <div class="socialaccount_ballot">
+        <ul class="socialaccount_providers">
+          {% include "socialaccount/snippets/provider_list.html" with process="login" %}
```
cwisecarver: :+2:

osf_oauth2_adapter/provider.py
cwisecarver: It's just part of the django-allauth package. I claim no responsibility for it other than installing it. We use it in SHARE and I think a couple of labs projects.

osf_oauth2_adapter/provider.py
```diff
@@ -0,0 +1,91 @@
...
+        self.cleanup_email_addresses(common_fields.get('email'),
+                                     email_addresses)
+        sociallogin = SocialLogin(account=socialaccount,
+                                  email_addresses=email_addresses)
+        user = sociallogin.user = adapter.new_user(request, sociallogin)
+        user.set_unusable_password()
+        user.is_researcher = True
+        user.identicon
+        adapter.populate_user(request, sociallogin, common_fields)
+        return sociallogin
```
cwisecarver: I'll dig into this.

project/settings/defaults.py
```diff
@@ -226,7 +233,7 @@
 
 
 # Configuration for cross-site requests
-CORS_ORIGIN_ALLOW_ALL = True  # Needs to be changed before production!
+CORS_ORIGIN_ALLOW_ALL = True  # TODO Needs to be changed before production!
```
cwisecarver: No worries. I just always do a TODO search before prod launch and if not fix them at least decide they're not terrifying.

accounts/models.py
```diff
@@ -157,18 +158,22 @@ def studies(self):
 
     @cached_property
     def is_org_admin(self):
-        if not self.organization:
+        if not self.organization_id:
```
cwisecarver: `organization_id` is the field in the studies table in SQL that represents the ForeignKey to the organization. If you have an instance of the study object you know the value of that field without having to talk to the database or really do anything other than the value comparison in python.

The second you reference `self.organization` django's lazy ORM is going to do a query for an organization with an `id` of `organization_id`, then it's going to create an instance of the `Organization` model with all of that data it gets back. Just unnecessary for what you're trying to accomplish. Django's not smart enough to know that you just want to null-check that field, not use the org name or any other field.

pattisdr: got it!

osf_oauth2_adapter/views.py
cwisecarver: Makes sense. The application in the OSF has the url as localhost:8000. I think oauth does some sort of validation around that to make sure you're coming from the right domain.

osf_oauth2_adapter/views.py
cwisecarver: It's built into the Django allauth package. We can override it if we need to by putting a template  with the same name inside of `templates/allauth/` inside one of our applications.

project/urls.py
cwisecarver: It's using the templates from django-allauth, which might not be optimal.

project/urls.py
cwisecarver: I think there's a class-level variable to set the login page url. I might make a mixin for exp's loginrequiredmixin and set that variable there.

exp/views/user.py
cwisecarver: I thought I did an `ag demographics__isnull` but I'll double-check.

cwisecarver:
I noticed the timestamps were all wrong while working on my celery PR. I fixed them in there. It's just setting the default timezone to US/New_York instead of UTC.

cwisecarver:
I'll look into setting the user's locale on login. It might just be a switch that we have to flip or a template tag around datetimes.

pattisdr:
Yeah, I wasn't sure what to do about the timezones.  If i did `{% get_current_timezone as TIME_ZONE %}` it was saying my timezone was UTC. that's what showing up for me.  I wanted to convert it to my current time zone. Then I saw this on [getting user's timezone](https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/#selecting-the-current-time-zone):

> The current time zone is the equivalent of the current locale for translations. However, there’s no equivalent of the Accept-Language HTTP header that Django could use to determine the user’s time zone automatically. 

So I thought we could use what you're pulling in from the API for the user to get their timezone.

osf_oauth2_adapter/views.py
cwisecarver: Fixed this. Pulled the gravatar url, base64ed it and threw it in _identicon. I also added fields for time_zone and locale on the user and made sure those got where they were going.

osf_oauth2_adapter/provider.py
cwisecarver: Calling `user.identicon` saved the user which made allauth very very angry. I removed that and replaced it with the value of their gravatar from the OSF and now logging in multiple times doesn't create more than one user.

exp/views/user.py
cwisecarver: Killed these from the api

project/urls.py
cwisecarver: I made a template for these two that inherits from our base exp template.

project/urls.py
cwisecarver: did this.

cwisecarver:
@pattisdr I think I've covered everything. If you have time give it a look. If not it can wait until after the demo. I'm going to go CR #31 Thanks! All good catches.

pattisdr:
it'll be the first thing I do after the demo @cwisecarver

pattisdr:
Reviewing now...

exp/views/user.py
```diff
@@ -186,9 +184,10 @@ class ResearcherDetailView(LoginRequiredMixin, DjangoPermissionRequiredMixin, ge
     def get_queryset(self):
         """
         Restrict queryset so org admins can only modify users in their organization
+        or unaffiliated researchers.
         """
         qs = super().get_queryset()
-        return qs.filter(organization=self.request.user.organization)
+        return qs.filter(Q(Q(organization=self.request.user.organization) | Q(is_researcher=True, groups__isnull=True, organization__isnull=True)))
```
pattisdr: I think this needs a distinct() at the end like Researcher List View.  Get errors sometimes depending on the user `MultipleObjectsReturned at /exp/researchers/2/
get() returned more than one User -- it returned 6!`

cwisecarver:
@pattisdr I didn't try to fix `Duplicate messages telling me my login was successful:` but I think that was happening because as you had exceptions after you were getting logged in the messages were getting pushed into the queue and they only get pulled out when they're displayed. 

I think everything else is fixed.

pattisdr:
ok! i'll take a look!

pattisdr:
@cwisecarver now I cannot login twice - getting your `Login Unsuccessful - A Lookit study participant exists with the email address for your OSF user account` with my second attempt.

cwisecarver:
I missed a `is_researcher` check. It's there now.

pattisdr:
ok, testing again..

pattisdr:
@cwisecarver thanks now I can login multiple times with the is_researcher added. Still seeing `Duplicate messages telling me my login was successful`

pattisdr:
This is working great, nice job.

PR #30, opened 2017-07-31
[WIP] Feature/Download CSV and Force Resubmission of Study if Edited[LEI-429]

pattisdr:
# Ticket 
https://openscience.atlassian.net/browse/LEI-429

# Purpose
1. Allow researcher to toggle between JSON/CSV display of individual and all responses.
2. Allow researcher to download individual responses/all responses as JSON or CSV.
3.  If certain study fields are edited after the study has been approved, force the study back into the "rejected" state, and leave comments explaining why.  User must resubmit for the study to be approved again.

** Note, I am giving the action a log - when it is moved back to rejected, but which user to attach to this log?  I attached the study's creator, but a system user or no log would make sense to me too.

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -1,5 +1,7 @@
 import operator, json
 import uuid
+import io
+import csv
 from functools import reduce
 
 from django import forms
@@ -347,6 +349,21 @@ class StudyResponsesList(LoginRequiredMixin, PermissionRequiredMixin, generic.De
             'demographic_id': resp.demographic_snapshot.id
             }, indent=4) for resp in responses]
 
+    def build_csv(self, responses):
+        headers = ['sequence', 'conditions', 'exp_data', 'participant_id', 'global_event_timings',
+        'child_id', 'completed', 'study_id', 'response_id', 'demographic_id']
+        
+        csv_responses = []
+        for resp in responses:
+            output = io.StringIO()
+            writer = csv.writer(output, quoting=csv.QUOTE_NONNUMERIC)
+            writer.writerow(headers)
+            writer.writerow([resp.sequence, resp.conditions, resp.exp_data, resp.child.user.id,
+            resp.global_event_timings, resp.child.id, resp.completed, resp.study.id, resp.id,
+            resp.demographic_snapshot.id])
+            csv_responses.append(output.getvalue())
+        return csv_responses
+
     def get_context_data(self, **kwargs):
         """
         In addition to the study, adds several items to the context dictionary.  Study results
@@ -359,6 +376,7 @@ class StudyResponsesList(LoginRequiredMixin, PermissionRequiredMixin, generic.De
         responses = study.responses.filter(completed=True).order_by(orderby) if orderby else study.responses.all()
         context['responses'] = self.paginated_queryset(responses, page, 10)
         context['response_data'] = self.build_responses(context['responses'])
+        context['csv_data'] = self.build_csv(context['responses'])
         context['all_responses'] = ', '.join(self.build_responses(responses))
         return context
 
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -341,7 +341,7 @@ class StudyResponsesList(LoginRequiredMixin, PermissionRequiredMixin, generic.De
         orderby = self.request.GET.get('sort', 'id')
         page = self.request.GET.get('page', None)
         study = context['study']
-        responses = study.responses.filter(completed=True).order_by(orderby) if orderby else study.responses.all()
+        responses = study.responses.filter(completed=True).order_by(orderby)
         context['responses'] = self.paginated_queryset(responses, page, 10)
         context['response_data'] = self.build_responses(context['responses'])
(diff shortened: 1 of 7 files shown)


studies/models.py
```diff
@@ -249,6 +262,22 @@ def add_study_created_log(sender, instance, created, **kwargs):
...
         )
 
+@receiver(post_save, sender=Study)
+def check_modification_of_approved_study(sender, instance, created, **kwargs):
+    """
+    Puts study back in "rejected" state if study is modified after it's already been approved.
+    Leaves comment for user with explanation.
+    """
+    approved_states = ['approved', 'active', 'paused', 'deactived']
+    if instance.important_fields_changed() and instance.state in approved_states:
```
cwisecarver: I'd put the approved_states check first, shorter list.

studies/models.py
```diff
@@ -249,6 +262,22 @@ def add_study_created_log(sender, instance, created, **kwargs):
             user=instance.creator
         )
 
+@receiver(post_save, sender=Study)
+def check_modification_of_approved_study(sender, instance, created, **kwargs):
+    """
+    Puts study back in "rejected" state if study is modified after it's already been approved.
+    Leaves comment for user with explanation.
+    """
+    approved_states = ['approved', 'active', 'paused', 'deactived']
```
cwisecarver: I worry about going around the workflow engine here but I also can't think of a better way to do it. 

I guess we could add another trigger that can only be used by the system. What do you think?

We should also probably warn the user, prior to saving, that their study will be deactivated and they will have to wait to be re-approved.

studies/models.py
```diff
@@ -60,10 +60,23 @@ def __init__(self, *args, **kwargs):
...
 
     def __str__(self):
         return f'<Study: {self.name}>'
 
+    def important_fields_changed(self):
+        """
+        Check if fields in self.__monitoring fields have changed.
+        """
+        for field in self.__monitoring_fields:
+            orig = '__original_%s' % field
```
cwisecarver: I had no idea this was there. We use django-dirtyfields in the osf, which does basically the same thing, but it feels like overkill compared to your implementation. 

I would probably change this to be the new python 3.6 string formatting `f'__original_{field}'`or at least `'__original_{}'.format(field)`. I think the `%` formatting has been frowned upon for a while. 

You also probably don't need the intermediate variable `orig` just do the interpolation in the getattr since it's only used once. 

Same things with the `__init__` usage.

exp/views/study.py
```diff
@@ -330,6 +332,23 @@ class StudyResponsesList(LoginRequiredMixin, PermissionRequiredMixin, generic.De
...
+    def get_context_data(self, **kwargs):
+        """
+        In addition to the study, adds several items to the context dictionary.  Study results
+        are paginated.
+        """
+        context = super().get_context_data(**kwargs)
+        orderby = self.request.GET.get('sort', 'id')
+        page = self.request.GET.get('page', None)
+        study = context['study']
+        responses = study.responses.filter(completed=True).order_by(orderby) if orderby else study.responses.all()
```
cwisecarver: If `orderby` then filter by `completed` else give them everything? Orderby has a default of `id` so unless it's set to something falsey it's always going to be truthy. Right?

exp/views/study.py
```diff
@@ -330,6 +332,23 @@ class StudyResponsesList(LoginRequiredMixin, PermissionRequiredMixin, generic.De
...
+        context['responses'] = self.paginated_queryset(responses, page, 10)
+        context['response_data'] = self.build_responses(context['responses'])
```
cwisecarver: I worry not having this in a separate view might slow things down. Doing it every time someone lists study responses in multiple formats even though they might only want one of them. It's pretty well broken down into pieces so we can split it up later if we need to.

studies/models.py
pattisdr: In my opinion, we don't need another trigger.  In your current workflow engine, a study can be rejected from many states  ['submitted', 'active', 'paused', 'deactivated'], and ['approved'] .  We are likewise moving the study to rejected from one of these states, so it seems allowed to me. The only confusion is who is doing the moving, so I think we need to create a "system" user or something so the logs can reflect it was moved to rejected state by the system.

I can add a modal that appears when the user in certain states and trying to modify certain fields/structure.

exp/views/study.py
pattisdr: yes missed this in my cleanup last week, thank you

exp/views/study.py
pattisdr: Do you want this addressed now @cwisecarver?

cwisecarver: I'll merge it as is. Let's get some data up on staging and see how it fares.

cwisecarver: It's a pretty big change and if I'm wrong then it'd be a lot of wasted time.

pattisdr: ok, still need to make some changes - confirmation modal if saving something that's going to be put back in rejected state. will let you know when ready.

pattisdr:
Comments addressed @cwisecarver - either left a response to your comment or made the change.

PR #31, opened 2017-08-02
Feature/API Queryset/Permissions Cleanup

pattisdr:
# Purpose

Audit API endpoints, making sure that data is properly restricted.  API may be accessed by both participants (who are taking the study) and researchers (who are previewing a study).  This is why a lot of the querysets are either 1) items attached to studies you have permission to view or 2) your own items

- [x] Restrict all endpoints to GET/OPTIONS methods only except for Responses which gets GET/POST/PUT/PATCH/DELETE/OPTIONS.  
- [x] Enforce need to be authenticated on all api views
- [x] Studies are either active, public. studies or studies you have permission to edit (so researchers can preview studies).  Study Detail view will allow private studies, so you can get a private, view-only link.
- [x] Organizations - show all
- [x] Responses -shows responses you have permission to view or your own responses.
- [x] Children - show children attached to studies whose responses you can view or your own children
- [x] Users - show users attached to studies whose responses you can view or your own user object
- [x] demographic data - show demographic data attached to responses you can view or your own demographic versions

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 from rest_framework import status
+from rest_framework.permissions import IsAuthenticated
 from guardian.shortcuts import get_objects_for_user
 from accounts.models import Child, DemographicData, Organization, User
 from accounts.serializers import (ChildSerializer, DemographicDataSerializer,
@@ -39,6 +40,7 @@ class OrganizationViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     serializer_class = OrganizationSerializer
     filter_fields = [('study', 'study'), ('user', 'user'), ]
     http_method_names = [u'get', u'head', u'options']
+    permission_classes = [IsAuthenticated]
 
 
 class ChildViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
@@ -48,6 +50,7 @@ class ChildViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     lookup_field = 'uuid'
     filter_fields = [('user', 'user'), ]
     http_method_names = [u'get', u'head', u'options']
+    permission_classes = [IsAuthenticated]
 
 
 class DemographicDataViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
@@ -57,6 +60,7 @@ class DemographicDataViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     serializer_class = DemographicDataSerializer
     filter_fields = [('user', 'user'), ]
     http_method_names = [u'get', u'head', u'options']
+    permission_classes = [IsAuthenticated]
 
 
 class UserViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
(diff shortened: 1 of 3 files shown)


pattisdr:
@cwisecarver - wanted to get your opinion on what studies are being shown through the API - right now it's studies you have permission to view, or active studies.  It was originally active and private studies, but then no one can use the "private" link that shows up on the detail page.  What are your thoughts?

api/views.py
```diff
@@ -35,9 +37,13 @@ def get_queryset(self):
 class OrganizationViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     resource_name = 'organizations'
     lookup_field = 'uuid'
+    # TODO - maybe we should restrict more, but I think it's fine to show
+    # all orgs to everyone through the API
     queryset = Organization.objects.all()
     serializer_class = OrganizationSerializer
     filter_fields = [('study', 'study'), ('user', 'user'), ]
+    http_method_names = [u'get', u'head', u'options']
```
cwisecarver: There's no more `u'strings'` in Py3. All strings are unicode by default. So while this won't break anything it also doesn't do anything.

pattisdr: Ah.

api/views.py
```diff
@@ -46,31 +52,78 @@ class ChildViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
...
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        """
+        Shows studies that are either 1) active and public or 2) studies you have permission to edit.
 
+        "can_edit_study" permissions allows the researcher to preview the study before it has been made active/public
+        """
+        return (super().get_queryset() | get_objects_for_user(self.request.user, 'studies.can_edit_study')).distinct()
```
pattisdr: I wasn't sure how to write this when I wanted to combine the queryset with using the guardian shortcut. Also, did you see this [question](https://github.com/CenterForOpenScience/lookit-api/pull/31#issuecomment-319964588)?

pattisdr:
@cwisecarver changes made I just wanted to make sure you saw my earlier question: https://github.com/CenterForOpenScience/lookit-api/pull/31#issuecomment-319964588

api/views.py
cwisecarver: oh, it returns a query set.

cwisecarver:
@pattisdr For the question above, I think you'd want private studies to appear on the detail view but not on the list view.

pattisdr:
got it. how hard is it to separate out this StudyViewSet into a list and detail view?

cwisecarver:
I'd keep it the same and put conditional logic in the `get_queryset` method based on whether it's a detail or a list request. I just spent like 20m looking during a meeting, haven't found a way to determine if it's detail or list request yet.

cwisecarver:
The best thing I can see so far is overriding the [`list` method](http://www.cdrf.co/3.6/rest_framework.mixins/ListModelMixin.html#list) on the viewset, reimplementing most of it's behavior, and filtering the queryset to exclude private things.

pattisdr:
good idea @cwisecarver, will try this

cwisecarver:
If you'd rather do it in `get_queryset` you might be able to find an indication of the view type in a debugger. Worst case, you could see if there's a uuid or pk in `request.kwargs` but that would be an absolute last resort.

pattisdr:
Used self.get_view_name() @cwisecarver to see if it was List View in get_queryset. Thanks for the tip!

pattisdr:
have responded to all comments @cwisecarver.

PR #32, opened 2017-08-04
Feature/Save Confirmation + QA Finds  + Demographic Data/Children Overhaul

pattisdr:
# Purpose

Main intent of PR addresses some of Todd's and Kim's finds. Also more confirmation when things are saved.  

- [x] Allow user to resubmit once retracted
- [x] When adding/searching researchers, anchor results to researcher section
- [x] Change publicly discoverable phrasing to "once activated" instead of "once approved"
- [x] Make signup page only ask for email/username from participants
- [x]  Do not show last name of participants on Participant List/Participant Detail pages, only username
- [x] Give confirmation when changes are saved:  when 1) creating a study 2) editing a study 3) editing study json 4) editing researcher 5) deleting researcher 6) cloning study 7) adding researcher to study 8) editing researcher study permissions 9) deleting researcher from study 10) changing status of study
- [x] Add more javascript to demographic data create form - certain fields hidden and only reveal based on answers to other questions. For example, state field appears when US selected.
- [x] Demographic data forms need to be similar to old Lookit - question phrasing should be similar.
- [x] Children forms need to be similar to old Lookit - field phrasing similar. 
- [x] Validate study metadata on study edit form - if try to add pdf on study edit for example, will fail
- [x] Label to keep file size less than 1 MB


studies/templates/studies/study_edit.html
```diff
@@ -37,7 +61,7 @@
                 <h1 class="panel-title">Description and Discoverability</h1>
               </div>
               <div class="panel-body">
-                <form action="" method="post" enctype="multipart/form-data">{% csrf_token %}
+                <form action="{% url 'exp:study-edit' study.id %}#details" method="post" enctype="multipart/form-data">{% csrf_token %}
```
pattisdr: Didn't try that, will look into it subsequent PR

PR #36, opened 2017-08-07
Feature/More Cleanup

pattisdr:
Visible changes:
- [x] If # of children not selected or 0, don't show child birthdays field
- [x] Confirmation modal before changing study state
- [x] Researcher breadcrumbs change if researcher details edited
- [x] Study state save buttons disable after submitting form so can't resubmit
- [x] Participant page responsiveness
- [x] Add user id to participant page and remove last name from xs view
- [x] Show both complete and incomplete responses on study responses page
- [x] Paginate researchers
- [x] Cancel button on create study page
- [x] Disable create study button after form submitted/edit study button
- [x] Only show osf profile link if logged in through oauth 
- [x] Remove button for study contributors is a minus sign on larger screens and Remove on xs screen. Larger screen needed more space for edit button.
- [x] Remove rejection comments when saving study
- [x] Prepopulate initial study structure with frames and sequence
- [x] Add number of guardians explanation field back
- [x] WIP solution for trailing slashes -


pattisdr:
thanks @cwisecarver, will address above comments in follow-up PR

cwisecarver:
👍

exp/views/dashboard.py
```diff
@@ -14,6 +14,10 @@ class ExperimenterDashboardView(ExperimenterLoginRequiredMixin, generic.Template
     template_name = 'exp/dashboard.html'
 
     def dispatch(self, request, *args, **kwargs):
-        if self.request.user.groups.exists():
-            return redirect(reverse_lazy('exp:study-list'))
-        return super().dispatch(request, *args, **kwargs)
+        if self.request.path.endswith('/'):
```
pattisdr: @cwisecarver so this PR also added similar logic to the proxy view.   The reason I added things to the dashboard view is that experimenter side, too many things were getting redirected to manage studies from the dashboard view if there was a problem.  I know it's not so proper, but the trailing slash thing seems to be working well across all views with this and the other fix.

PR #38, opened 2017-08-10
Feature/Cleanup #4 [LEI-442]

pattisdr:
Tickets
https://openscience.atlassian.net/browse/LEI-442

QA Finds
- [x]  Completed Responses/Incomplete responses are one combined link on manage studies page
- [x] Page number is reflected in URL when filtering studies
- [x] If sorting, or filtering on study state, deliver page 1 of search results and reflect in URL
- [x] If search, previous search text shows up in input bar and can be deleted.

### DIFF
--- a/accounts/templates/accounts/participant_list.html
+++ b/accounts/templates/accounts/participant_list.html
@@ -8,11 +8,18 @@
 {% endblock %}
 {% block content %}
   <div class="container mb-lg">
+    <script>
+        $(document).ready(function() {
+            if ("{{ match }}" !== '') {
+                document.getElementById("search-participants").value = "{{ match }}";
+            }
+        });
+    </script>
     <h1>Participants</h1>
     <div class="row text-center mt-xl">
        <div class="col-xs-12 col-sm-offset-1 col-sm-10 col-md-offset-2 col-md-8">
            <form method="GET">
-               <input class="form-control" name="match" placeholder="{{ match|default:'Filter by name' }}" size="50" type="text">
+               <input id="search-participants" class="form-control" name="match" placeholder="Filter by name" size="50" type="text">
                <input type="hidden" name="sort" value={{ sort }}>
                <input type="hidden" name="page" value="1">
            </form>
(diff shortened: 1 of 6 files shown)


accounts/templates/accounts/participant_list.html
```diff
@@ -44,16 +52,16 @@
    <div class="row mt-xl visible-md visible-lg visible-xl">
        <div class="col-md-3 col-md-offset-1">
            <strong> Username </strong>
-           <a class="pl-sm" aria-label="Sort ascending by username" href="{% url 'exp:participant-list' %}?{% query_transform request sort='given_name' %}"><i aria-hidden="true" class="fa fa-chevron-up" role="button" name="name"></i></a>
-           <a aria-label="Sort descending by username" href="{% url 'exp:participant-list' %}?{% query_transform request sort='-given_name' %}"><i aria-hidden="true" class="fa fa-chevron-down" role="button" name="name"></i></a>
+           <a class="pl-sm" aria-label="Sort ascending by username" href="{% url 'exp:participant-list' %}?{% query_transform request page='1' sort='given_name' %}"><i aria-hidden="true" class="fa fa-chevron-up" role="button" name="name"></i></a>
```
cwisecarver: Could you just change the `query_transform` tag to implicitly go to page one if sort is specified instead of passing it every time?

cwisecarver: If you think it's good the way it is lmk and I'll merge it or if you want to change it in another PR lmk and I'll merge it.

pattisdr: Then paginating won't work right.  If "sort" is specified as "name", but I am actively paginating my results, going to page 2, page 3, I need it to stay at page one.

This query transform tag sets the url before you even click it.  It's not reacting to a button you just clicked, it just builds the url based on what the current parameters of the page are.

cwisecarver: 👍 Makes sense. I'll merge it now.

pattisdr: Maybe a better explanation, passing in the page='1' is an override.  The query transform tag I wrote allows you to build a URL using whatever the current query params are, plus any new query params you pass in.   So passing in page=1 allows me to override the page, forcing 1 instead of the current page. (And sometimes I want the current page, like when I'm actively paginating).

PR #40, opened 2017-08-14
Feature/Cleanup #5 (Demographic Data, Children, and Study Permissions) [LEI-445]

pattisdr:
# Ticket
https://openscience.atlassian.net/browse/LEI-445

# Purpose
PR responds to some of Kim's feedback plus some other polishes like save confirmation Lookit-side, and enforcing @ least one study admin at all times.

- [x] Adds contact_name field to user model so on signup, participants see username, email, contact name, and password fields
- [x] Save confirmation messages Lookit-side
- [x] Only show explanation field if participant selects "varies" for number of guardians
- [x] All demographic data questions are now optional
- [x] United States is first country listed on country dropdown
- [x] Text correction on child update form "This lets us figure out exactly how old your child <b>is </b> when he or she participates in a study. We never publish children's birthdates or information that would allow a reader to calculate the birthdate."
- [x] Make it so parents cannot update birthday after child has been created
- [x] Change datepicker on child create form when picking birthday. Display rough age of child above field when selected
- [x] Enforce that there must be one study admin at all times


accounts/forms.py
```diff
@@ -130,19 +131,20 @@ class Meta:
...
-            'number_of_guardians_explanation': forms.Textarea(attrs={'rows':2})
+            'number_of_guardians_explanation': forms.Textarea(attrs={'rows':2}),
+            'lookit_referrer': forms.Textarea(attrs={'rows':2})
         }
 
 
 class ChildForm(forms.ModelForm):
-    birthday = forms.DateField(widget=DateInput(attrs={'type': 'date'}), help_text="This lets us figure out exactly how old your child when he or she participates in a study. We never publish children\'s birthdates or information that would allow a reader to calculate the birthdate.")
-
+    birthday = forms.DateField(widget=forms.DateInput(attrs={'class': 'datepicker'}), help_text="This lets us figure out exactly how old your child is when he or she participates in a study. We never publish children\'s birthdates or information that would allow a reader to calculate the birthdate.")
```
cwisecarver: Maybe "they" instead of he or she based on the options in the gender choices?

studies/templates/studies/study_edit.html
```diff
@@ -118,7 +150,7 @@ <h1 id="manage-researchers" class="panel-title">Manage Researchers</h1>
                 {% if messages %}
                     {% for message in messages %}
                         {% if message.extra_tags == 'user_added' or message.extra_tags == 'user_removed' %}
-                            <div id="add-researcher-messages" class="alert alert-success alert-dismissable">
+                            <div id="add-researcher-messages" class="alert {{ message.tags }} alert-dismissable">
```
cwisecarver: Will there ever be multiple types of messages? What happens then?

web/templates/web/base.html
```diff
@@ -18,11 +18,13 @@
     {% block head %}
     {% endblock %}
     <script src="https://use.fontawesome.com/41c082d3c2.js"></script>
+    <link rel="stylesheet" href="http://code.jquery.com/ui/1.10.3/themes/smoothness/jquery-ui.css" />
```
cwisecarver: Might be worth pulling these from https://cdnjs.com. I think that's OSF's default place to borrow things from.

web/templates/web/children-list.html
```diff
@@ -13,6 +12,43 @@
...
+                   var x = document.createElement("p");
+                   x.setAttribute("id", "age_calc");
+                   x.setAttribute("class", "age_format")
+                   var t = document.createTextNode(`Age: ${age}`);
+                   x.appendChild(t);
+                   document.getElementsByClassName("form-group")[1].insertBefore(x,document.getElementById("id_birthday"));
+               }
+           });
+        });
+        function getAge(dateValue) {
```
cwisecarver: @icereval Is it worth pulling in moment.js for this kind of thing? I know how much we all love leap seconds. Not sure if it's worth the extra kb of js.

studies/templates/studies/study_edit.html
pattisdr: That's what the message.tags was for.  Before it could only show success messages, now it can show any type.

pattisdr:
I'll make these changes right now @cwisecarver before merge.

studies/templates/studies/study_edit.html
pattisdr: I'm normally not looping through messages, most of the site just does {% bootstrap_messages %} but I wanted more control over where certain success messages show on this page.  The researcher messages I needed to show directly over the researchers.  Also, when researchers are edited, I am using x-editable, so I had to build the success messages myself since the page isn't reloading.

web/templates/web/children-list.html
pattisdr: How this is working, if younger than a month, I am showing age in days, if younger than a year, i show age in months, rounded down, if younger than two years, I show age as 1 year x months, and older than two years, show age in years.  Not exact but pretty close.

pattisdr:
made changes/responded to comments @cwisecarver

studies/templates/studies/study_edit.html
cwisecarver: Ah, okay. Misunderstood. 👍

web/templates/web/children-list.html
cwisecarver: @pattisdr I would use moment.js but keep the same functionality you have now. Just use it for the calculations so you don't have to worry about leap years or any of that nonsense.

pattisdr:
@cwisecarver adding momentjs now

pattisdr:
@cwisecarver this is ready, using momentjs, thanks for the tip. Way cleaner now.

PR #41, opened 2017-08-16
Feature/Email Participants [LEI-446]

pattisdr:
# Ticket
https://openscience.atlassian.net/browse/LEI-446

# Feature Requirements
- Researchers can send emails to participants that have taken part in or are assigned to their study
- Researchers cannot view individual participants’ email addresses
- Researchers can send one email to all participants in their study
- Researchers can send an email to an individual participant in their study

# Changes
- [x] url added studies/study_id/email/ 
- [x] StudyParticipantsEmail view added
- [x] On Study Detail Page, there is a link to the study email page
- [x] Build email form
- [x] Default email is a no-reply email defined on server, but can override with your own email
- [x] Can multiselect from all participants that have responded to your study, only username and id shown for now 
- Show completed status in dropdown **maybe not, participants can respond to study multiple times**
- [x] Add subject and message
- [x] All email fields required
- [x] When sending email, send button is disabled and says "Sending"
- [x] After email sent, directed to study detail page with confirmation message
- [x] Emails are logged
- [x] Make sure participants are bcc (i think this should technically work!?)
- [x] Made edits to send_email helper to support custom emails and custom "from" fields
- [x] Need permission to edit the study to send the email - so just study admins and org admins

screen shot 2017-08-15 at 7 15 52 pm
screen shot 2017-08-15 at 7 16 13 pm


exp/views/study.py
```diff
@@ -200,6 +202,64 @@ def get_context_data(self, **kwargs):
...
+        '''
+        return User.objects.filter(Q(children__response__study__id=self.get_object().id) & Q(email_personally=True)).distinct()
+
+    def post(self, request, *args, **kwargs):
+        """
+        Post form for emailing participants.
+        """
+        retval = super().get(request, *args, **kwargs)
+        email_form = self.request.POST
+        researcher = User.objects.get(id=self.request.user.id)
```
cwisecarver: You already have `self.request.user` you shouldn't need to do the extra query.

exp/views/study.py
```diff
@@ -200,6 +202,64 @@ def get_context_data(self, **kwargs):
...
+        context['sender'] = settings.EMAIL_FROM_ADDRESS
+        context['participants'] = self.get_study_participants()
+        return context
+
+    def get_study_participants(self):
+        '''
+        Restricts list to participants that have responded to this study as well as participants
+        that have given their permission to be emailed personally
+        '''
+        return User.objects.filter(Q(children__response__study__id=self.get_object().id) & Q(email_personally=True)).distinct()
```
cwisecarver: You can just do `Q(children__response__study=self.get_object())`. The object is already queried and built when you call get_object so doing the id doesn't really save anything.

exp/mixins/participant_mixin.py
```diff
@@ -4,7 +4,9 @@
 from django.contrib.auth.mixins import PermissionRequiredMixin as DjangoPermissionRequiredMixin
 
 class ParticipantMixin(DjangoPermissionRequiredMixin, object):
```
cwisecarver: Don't need to inherit from object in py3.

cwisecarver:
@pattisdr It could wait until morning too. Go enjoy some of that Texas heat. 😃

studies/helpers.py
```diff
@@ -5,7 +5,7 @@
 
 
 # TODO: celery taskify
-def send_mail(template_name, subject, to_addresses, cc=None, bcc=None, **context):
+def send_mail(template_name, subject, to_addresses, cc=None, bcc=None, custom_message=None, from_email=None, **context):
```
pattisdr: well that is a much better way to do it :)

exp/mixins/participant_mixin.py
pattisdr: Pulling from develop fixed - made this change on the other PR!

pattisdr:
Changes made @cwisecarver, I also stepped through and am pretty sure the bcc would work too.

cwisecarver:
We'll test on staging.

PR #42, opened 2017-08-17
Feature/Download Participant Data + Past Studies Page + API Tests [LEI-447]

pattisdr:
# Purpose 

New features/QA Finds/Asks from Kim/Tests
Includes fix for https://openscience.atlassian.net/browse/LEI-447

# Changes 

- [x] Add Demographic Snapshots page - tab to view all demographic snapshots that correspond with each response
- [x] Add Past Studies page Lookit side that shows all the past studies you've participated in.  For each study, shows your responses, plus feedback from potentially multiple researchers.
- [x] Model changes: create feedback model, link to researcher and response
- [x] Modify UI to display multiple pieces of feedback for study response on past studies page
- [x] Allow creation of Feedback through the API
- [x] Add Feedback permissions - GET (must have can_edit_study permissions on the associated study, or viewing feedback left on your own study. POST (must have can_edit_study permissions)
- [x] Tests for POST feedback
- [x] Add basic API test coverage for studies, responses, children, users, organization, feedback, demographic data
- [x] Allow users to edit their email address.  Email address cannot already exist.
- [x] Participant signup fields are now email and nickname, remove username and contact_name
- [x] If child is too old or too young, show warning message on study detail page but allow them to take study anyway
- [x]  When Account Information edited, when form reloaded, focus on username field
- [x] After filling out demographic survey, redirect to child-list page if no children
- [x] Allow bulk downloading of just consent videos
- [x] Can now have 255 char length study titles


pattisdr:
Ok, done with commits, just snuck in one more merging migrations.

exp/mixins/study_responses_mixin.py
```diff
@@ -44,8 +104,8 @@ def build_responses(self, responses):
                     'id': resp.study.id
                 },
                 'participant': {
-                    'id': resp.child.user_id,
-                    'username': resp.child.user.given_name
+                    'id': resp.child.user.id,
```
cwisecarver: `resp.child.user.id` will cause a join where `resp.child.user_id` will not.

exp/views/study.py
```diff
@@ -556,12 +585,11 @@ def get_study_attachments(self, study, orderby, match):
         return sorted(attachments, key=lambda x: getattr(x, sort), reverse=True if '-' in orderby else False)
 
     # TODO move to celery task
-    def download_all_files(self):
+    def download_multiple_files(self, all, folder_name):
```
cwisecarver: I'd rename `all` to `files` or `files_to_download` or something like that.

exp/views/study.py
```diff
@@ -556,12 +585,11 @@ def get_study_attachments(self, study, orderby, match):
...
          """
          Downloads all attachments associated with study and puts into zipfile
          """
-         all = self.get_study_attachments(self.get_object(), 'last_modified', '')
-         zip_subdir = "study_attachments"
+         zip_subdir = folder_name
```
cwisecarver: You could remove this intermediate variable.

exp/views/study.py
```diff
@@ -556,12 +585,11 @@ def get_study_attachments(self, study, orderby, match):
...
          zip_filename = "%s.zip" % zip_subdir
```
cwisecarver: Python 3 string interpolation or at least `.format`

exp/views/study.py
cwisecarver: `folder_name` is technically a filename, right?

cwisecarver: I might add the study uuid to the filename of the zip just so it's easier for the researchers.

web/templates/web/study-detail.html
```diff
@@ -31,23 +30,85 @@
...
+                    document.getElementById('too-old').style.display = 'inline-block';
+                } else {
+                    document.getElementById('too-young').style.display = 'inline-block';
+                }
+            }
         }
 
         function portToParticipate(selected) {
-            window.location.href = "/studies/" + "{{ study.uuid }}" + '/' + selected.value;
+            window.location.href = "/studies/" + "{{ study.uuid }}" + '/' + selected.value.split('|')[0];
```
cwisecarver: You could use the `url` template tag for this just to future proof it in case someone changes the studies URL.

api/views.py
```diff
@@ -79,15 +82,16 @@ def get_queryset(self):
...
         study_ids = studies.values_list('id', flat=True)
-        return DemographicData.objects.filter(Q(response__study__id__in=study_ids) | Q(user__id=self.request.user.id)).distinct()
+        return DemographicData.objects.filter((Q(response__study__id__in=study_ids) | Q(user__id=self.request.user.id)), (Q(id__in=qs_ids))).distinct()
 
 
 class UserViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
     lookup_field = 'uuid'
     resource_name = 'users'
-    queryset = User.objects.filter(is_researcher=False).distinct()
+    queryset = User.objects.all()
```
pattisdr: I had tried to remove the queryset before and got `AssertionError: `base_name` argument not specified, and could not automatically determine the name from the viewset, as it does not have a `.queryset` attribute.` I tried to replace it here with `model = User` and got the same thing.

cwisecarver: Alright, I might be wrong. Not a big deal. I'd just leave it the way that works.

web/templates/web/study-detail.html
pattisdr: I moved building the url into a javascript function because the "Participate" button needed to be conditionally disabled and I don't think you can disable links, just buttons.

pattisdr:
Either responded/made changes to your comments @cwisecarver. Thanks for the review!

web/templates/web/study-detail.html
cwisecarver: Right, I just meant you could still use the template engine like:
`window.location.href = '{% url 'study-detail' study.uuid %}' + '/' + selected.value.split('|')[0];`

There might be too many slashes in the middle, don't remember if `url` will put a slash on the end or not. Also, I just found out that this needs a trailing slash on the URL, at least in my celery branch, or it gets a 403 from Google.

pattisdr: Ok, thanks, got it working @cwisecarver

PR #46, opened 2017-08-25
Feature/Make Send Email a Celery Task + Cleanup # 7 [LEI-451]

pattisdr:
# Purpose

Continued cleanup

# Changes
- [x] Change number of books default to None instead of 0 in case researcher tries to submit null response
- [x]  Remove autofocus on all fields on Account page https://openscience.atlassian.net/browse/LEI-451
- [x] Return 403 on Researcher List page if user does not belong to an organization (scenario only possible for superuser)
- [x] Modify all usages of send_mail and associated templates so complex objects are not being passed to celery
- [x] Make sending email a celery task
- [x] Change language of notify_researchers_of_approval_decision template to reflect that you don't click "Start study" to begin collecting data, but rather, choose "Activate" from change status dropdown
- [x] Reset selected email type when navigating to Email Participants page
- [x] Make deploying status capitalized


pattisdr:
@cwisecarver, if I remove this line then emails start showing up in celery https://github.com/CenterForOpenScience/lookit-api/blob/5fe584287014420f46c6db5235c2658ad7174c4d/project/settings/defaults.py#L319

accounts/utils.py
```diff
@@ -30,7 +30,8 @@ def build_study_group_name(org_name, study_name, study_pk, group):
     'retracted': 'Study has been withdrawn',
     'paused': 'Study is not collecting data',
     'deactivated': 'Study is not collecting data',
-    'archived': 'Study has been archived and removed from search.'
+    'archived': 'Study has been archived and removed from search.',
+    'deploying': 'Study is being built and deployed to S3'
```
cwisecarver: It's Google Cloud Storage not s3.

pattisdr:
Thanks @cwisecarver, change made.

PR #48, opened 2017-08-29
Feature/Study Type [LEI-440]

pattisdr:
# Purpose

Implement an experiment type which specifies a build process and a JSON schema for the parameters needed by the experiment builder. It should also include the necessary template partials to fill out the parameters.

# Ticket 
https://openscience.atlassian.net/browse/LEI-440

# Changes
- [x] Create Study page has a dropdown where you can select study type, and additional fields pulled from configuration.metadata.fields.  Study type must be selected to create a study.
- [x] When creating a study, configuration.metadata.fields are saved into a json blob under study.metadata
- [x] Cloning a study copies both study type and metadata to clone
- [x]  If modify the study type or metadata after the study has been approved, move the study to rejected
- [x] Study Build page modified to include place to edit study type and study metadata
- [x] Study Build page type/metadata defaults to current saved data
- [x] Try experiment button moved to original position
- [x] Add success message if study type/metadata saved.
- [x] Build URL's properly in emails where study state has changed (extra bracket was present)

screen shot 2017-08-28 at 9 03 53 pm


exp/views/study.py
```diff
@@ -434,12 +444,30 @@ def get_initial(self):
...
         return initial
 
+    def post(self, request, *args, **kwargs):
+        """
+        Form for allowing user to change study type and study metadata fields
+        """
+        study = self.get_object()
+        if 'structure' not in self.request.POST:
+            study.metadata = self.extract_type_metadata()
+            study.study_type = StudyType.objects.get(id=self.request.POST.get('study_type'))
```
cwisecarver: You could do:
```
study.study_type_id = self.request.POST.get('study_type')
```
Assuming that you're passing the PK as study_type in the POST.

cwisecarver: Your way does validate that the study type exists tho. If you still want to do that you could `StudyType.objects.filters(id=self.request.POST.get('study_type')).values_list('id', flat=True)[0]`

exp/views/study.py
```diff
@@ -434,12 +444,30 @@ def get_initial(self):
...
         """
-        In addition to the study, adds whether save confirmation is needed to study.
+        In addition to the study, adds whether save confirmation is needed to study, as well as all study_types
+        and current metadata
         """
         context = super().get_context_data(**kwargs)
         context['save_confirmation'] = self.object.state in ['approved', 'active', 'paused', 'deactivated']
+        context['study_types'] = StudyType.objects.all()
+        context['study_metadata'] = self.object.metadata
+        context['types'] = [type.configuration['metadata']['fields'] for type in context['study_types']]
```
cwisecarver: Shouldn't use type as a variable name.

studies/templates/emails/notify_admins_of_study_action.html
```diff
@@ -13,10 +13,10 @@
   {% elif action == 'paused' %}
     {{ researcher_name }} has paused the study {{ study_name }}. {{base_url}}{% url 'exp:study-detail' study_id %}
   {% elif action == 'deactivated' %}
-    {{ researcher_name }} has deactivated the study {{ study_name }}. {{base_url}}{{% url 'exp:study-detail' pk=study_id %}
+    {{ researcher_name }} has deactivated the study {{ study_name }}. {{base_url}}{% url 'exp:study-detail' pk=study_id %}
   {% elif action == 'deployed' %}
-     {{ study_name }} has been deployed. {{base_url}}{{% url 'exp:study-detail' pk=study_id %}
+     {{ study_name }} has been deployed. {{base_url}}{% url 'exp:study-detail' pk=study_id %}
```
cwisecarver: For both this and previewed I linked to the study detail page. Do you think we should link to the page where the study can be taken too? As of right now, at least for preview, this isn't exposed anywhere. I changed your button that took you to the preview to actually trigger the preview job, now there's no way to get to take the preview without URL hacking.

studies/templates/emails/notify_admins_of_study_action.txt
```diff
@@ -11,9 +11,9 @@ Dear {{ org_name }} Admin,
 {% elif action == 'paused' %}
     {{ researcher_name }} has paused the study {{study_name}}. {{base_url}}{% url 'exp:study-detail' pk=study_id %}
 {% elif action == 'deactivated' %}
-    {{ researcher_name }} has deactivated the study {{study_name}}. {{base_url}}{{% url 'exp:study-detail' pk=study_id %}
+    {{ researcher_name }} has deactivated the study {{study_name}}. {{base_url}}{% url 'exp:study-detail' pk=study_id %}
 {% elif action == 'deployed' %}
-    {{ study_name }} has been deployed. {{base_url}}{{% url 'exp:study-detail' pk=study_id %}
+    {{ study_name }} has been deployed. {{base_url}}{% url 'exp:study-detail' pk=study_id %}
```
cwisecarver: Thanks for fixing these by the way. It was hard to see the extra curly in the console backend printout.

exp/views/study.py
pattisdr: i don't understand why you want to set study_type_id instead of study_type. does this save time?

studies/templates/emails/notify_admins_of_study_action.html
pattisdr: Yes, i'll add this.

exp/views/study.py
cwisecarver: `study.study_type` is really just Django ORM magic, in the database there's a field on `studies_study` table called `study_type_id` which is the ForeignKey to the `studies_studytype` table. You can bypass instantiating the model, doing the query to get it: `StudyType.objects.get(id=self.request.POST.get('study_type'))` and just set `study.study_type_id` to the PK of the StudyType and it will work. Make sense?

pattisdr: yes, that's fancy :)

pattisdr:
thanks @cwisecarver, changes made

PR #51, opened 2017-08-30
Feature/Lookit Cleanup 9

pattisdr:
# Purpose 

Fix QA Finds

# Changes
- [x] Don't show deleted children in dropdown when trying to participate in study on Lookit Study Detail Page
- [x] Use https instead of http
- [x] Messages if no attachments - letting researcher know there are delays

### DIFF
--- a/accounts/templates/emails/resend_confirmation.html
+++ b/accounts/templates/emails/resend_confirmation.html
@@ -6,7 +6,7 @@
 </p>
 <p>
     After verification, use your OSF credentials to <a href="{{ base_url }}{{ login_url }}">login to Experimenter</a>.
-</p>x
+</p>
 <p>
     Best,
     <div> {{ org_name }} Admin </div>
--- a/accounts/templates/emails/resend_confirmation.html
+++ b/accounts/templates/emails/resend_confirmation.html
@@ -2,10 +2,10 @@
   Dear {{ researcher_name }},
 </p>
 <p>
-    Here is a link to verify your email address on the OSF: {{ osf_url}}resend/.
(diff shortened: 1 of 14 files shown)


accounts/templates/emails/resend_confirmation.html
```diff
@@ -2,11 +2,11 @@
   Dear {{ researcher_name }},
 </p>
 <p>
-    Here is a link to verify your email address on the OSF: {{ osf_url}}/resend/.
+    Here is a link to verify your email address on the OSF: {{ osf_url}}resend/.
```
cwisecarver: I'm sure 99.999% of email clients will turn this and the one below it into a link, but shouldn't it be inside of an anchor tag?

accounts/templates/emails/resend_confirmation.html
```diff
@@ -2,11 +2,11 @@
...
 </p>
 <p>
     After verification, use your OSF credentials to login to Experimenter: {{ base_url }}{{ login_url }}.
-</p>
+</p>x
```
cwisecarver: I don't think you need this x

web/views.py
```diff
@@ -319,7 +319,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         if self.request.user.is_authenticated:
             context['has_demographic'] = self.request.user.latest_demographics
-            context['children'] = self.request.user.children.all()
+            context['children'] = self.request.user.children.filter(deleted = False)
```
cwisecarver: It really doesn't matter but if you're fixing the x and the anchor thing would you kill the spaces around the `=`? Thanks!

cwisecarver:
Thanks @pattisdr !

PR #58, opened 2017-09-06
Feature/Lookit API Read the Docs

pattisdr:
# Purpose

-  use sphinx to generate documentation to be stored on read the docs
- Add definitions to docs
- Add API documentation
- Tweak API so you can't add response with child id that is not your own.
- Remove hardcoded client_id and secret and make project settings
- Use relative urls for static images


docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
sloria: It would be nice to have a management command to build and optionally open the docs, e.g. `python manage.py docs --browse`. 

That can be done in a separate PR, though.

pattisdr:
Thank you @sloria!

PR #64, opened 2017-09-13
Add django celery beat and make celery beat the default scheduler

### LINKED ISSUE #55: Use Django Celery Beat Scheduler
Consider using https://github.com/celery/django-celery-beat so we do not have to mount a "special" persistent disk for the database file beat generates by default.

http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#using-custom-scheduler-classes - Item `#4`

Reference: https://github.com/CenterForOpenScience/SHARE

cwisecarver:
Resolves #55


cwisecarver:
Tests failed because I broke develop before I PRed off it. The merge should fix this.

PR #72, opened 2017-09-15
Feature/Add Documentation for local development of frames

pattisdr:
# Purpose

Add docs that explain how to get the whole setup running locally so you can make changes to frames and test them in your local environment.


pattisdr:
@cwisecarver documentation for local development of frames done. I need to copy some old documentation over from Experimenter to these docs, but that's separate and can be done later. I'd like to get these instructions on read the docs.

PR #74, opened 2017-09-18
Feature/Pull in Relevant Docs from Old Experimenter

pattisdr:
# Purpose

Experimenter has changed a lot, but the process for creating frames has not.  Pull over and modify the old frame docs from Experimenter v1 regarding creating frames, mixins in the exp-player and randomization in frames as they are still relevant to the new application.

- Also adds link to custom email where users can change email preferences

### DIFF
--- a/docs/source/frames.md
+++ b/docs/source/frames.md
@@ -1,0 +1,291 @@
+# Development: Custom Frames
+
+### Overview
+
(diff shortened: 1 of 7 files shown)


pattisdr:
@cwisecarver email preferences link added to this PR

PR #87, opened 2017-11-16
Migration changes

kimberscott:
Included here are some suggested minor changes before migration - here for concreteness, merge only if useful to you.

1. Transfer old 'name' field to new nickname field when available

2. Show some useful fields in the admin interface so we can find researchers / participants more easily if manual changes are needed (e.g. giving permissions). I'm not attached to the specific format here - e.g., if it'd be better to primarily display user uuids and order the user list by uuid, that's fine. (I just don't want it ordered by (hidden) username when I need to use a dropdown menu to select among thousands of options.) Or if you'd rather not include/allow sorting by certain fields in the admin interface.

### DIFF
--- a/accounts/admin.py
+++ b/accounts/admin.py
@@ -5,11 +5,11 @@ from accounts.models import DemographicData, Organization, Child, User
 
 
 class ChildAdmin(GuardedModelAdmin):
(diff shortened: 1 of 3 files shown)


pattisdr:
thanks @kimberscott, I'm going to close this and add some more changes on top of yours before merging

PR #96, opened 2018-03-09
Quick fixes: Update Travis repo and fix age calculation bug

kimberscott:
* Switch Travis repo owner centerforopenscience -> lookit

* Fix bug in age calculation for study eligibility. Previously a child's age was being calculated by first converting to a moment 'duration', and then converting that back to days. This means that if the child was born, say, February 1st, then on March 1st they would be considered "1 month old" which is 30 days. Actually the child is 28 days old. This was leading to confusion for participants in the Baby Euclid study w/ a narrow age range. It's now fixed & we also display when the child WILL be eligible, so that e.g. someone who's just a few days out of the age range can easily see when they might want to check back.

### DIFF
--- a/web/templates/web/study-detail.html
+++ b/web/templates/web/study-detail.html
@@ -43,10 +43,12 @@
             document.getElementById('too-young').style.display = 'none';
             var birthday = selected.value.split('|')[1];
             var age = calculateAgeInDays(birthday);
-            if (!isEligible(age)) {
-                if (isOldEnough(age)) {
+            var ineligible = ageCheck(age);
+            if (ineligible) {
+                if (ineligible > 0) {
                     document.getElementById('too-old').style.display = 'inline-block';
                 } else {
+                	$('#wait-until').text(moment.duration(-ineligible, "days").humanize() + ' until ' + moment(moment()._d).add(-ineligible, 'days').format("dddd, MMMM Do YYYY") + ' when ');
                     document.getElementById('too-young').style.display = 'inline-block';
                 }
             }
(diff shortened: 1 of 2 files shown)


web/templates/web/study-detail.html
```diff
@@ -148,7 +154,7 @@ <h4>Would you like to participate in this study?</h4>
                                 <option onemptied="" value="{{child.uuid}}|{{child.birthday}}">{{child.given_name}}</option>
                                 {% endfor %}
                             </select>
-                            <p class="text-warning" style="display:none" id='too-young'>Your child is still younger than the recommended age range for this study. If you can wait until he or she is old enough, we'll be able to use the collected data in our research! </p>
+                            <p class="text-warning" style="display:none" id='too-young'>Your child is still younger than the recommended age range for this study. If you can wait until <span id='wait-until'></span> he or she is old enough, we'll be able to use the collected data in our research! </p>
```
pattisdr: I would remove this "until" here, so you don't have " f you can wait until X days until ...." days

web/templates/web/study-detail.html
```diff
@@ -57,16 +59,20 @@
         }
 
         function calculateAgeInDays(birthday) {
-            var birthdayDate = new Date(birthday);
-            return moment.duration(moment()._d - birthdayDate).asDays();
+            // Warning: do NOT use moment.duration in the calculation of age! Use diffs
+            // instead to get ACTUAL time difference, without passing through an 
+            // approximation where each month is 30 days and each year is 365.
+            return moment(moment()._d).diff(new Date(birthday), 'days');
```
pattisdr: Nice 👍

web/templates/web/study-detail.html
```diff
@@ -43,10 +43,12 @@
             document.getElementById('too-young').style.display = 'none';
             var birthday = selected.value.split('|')[1];
             var age = calculateAgeInDays(birthday);
-            if (!isEligible(age)) {
-                if (isOldEnough(age)) {
```
pattisdr: If you are not going to use the isEligible and isOldEnough methods since you need the `ineligible` count, I would remove those methods below.

web/templates/web/study-detail.html
kimberscott: Thanks for catching that! "Until" now just in the span as a default so the sentence still makes sense if that text isn't replaced.

web/templates/web/study-detail.html
kimberscott: Done - thanks!

PR #98, opened 2018-04-06
Edit demographic form for GDPR compliance

kimberscott:
MIT's office of general counsel advised us stop collecting any demographic information from EU participants for the time being. I made some quick changes to the demographic survey so that if someone hasn't selected a non-EU country no personal data is collected. Have tested on staging and it works as intended.

### DIFF
--- a/web/templates/web/demographic-data-update.html
+++ b/web/templates/web/demographic-data-update.html
@@ -74,23 +74,25 @@
             }
             
             $('#id_country').on('change', function() {
+            	// First show/hide state field based on USA or not
                 if ($(this)[0].value === 'US') {
                     showField('state');
                 } else {
                     hideField('state');
                     $(`#id_state`)[0].value = '';
-                    if (gdpr_country_list.includes($('#id_country')[0].value)) {
-						$.each(gdprFieldsToHide, function(i, fieldName) {
-							hideField(fieldName);
-							$(`#id_${fieldName}`)[0].value = '';
-							$(`#id_${fieldName} input[type=checkbox]`).prop("checked", false );
-						});
-					} else {
-						$.each(gdprFieldsToHide, function(i, fieldName) {
-							showField(fieldName);
-						});
-					}
                 }
+                // Then show/hide most demographic fields based on EU or not
+				if (gdpr_country_list.includes($('#id_country')[0].value)) {
+					$.each(gdprFieldsToHide, function(i, fieldName) {
+						hideField(fieldName);
+						$(`#id_${fieldName}`)[0].value = '';
+						$(`#id_${fieldName} input[type=checkbox]`).prop("checked", false );
+					});
+				} else {
+					$.each(gdprFieldsToHide, function(i, fieldName) {
+						showField(fieldName);
+					});
+				}
             });
 
             $('#id_number_of_children').on('change', function() {
@@ -143,7 +145,7 @@
                       multiple languages or having older siblings affect children's learning. </p>
                       <p> Even if you allow your study videos to be published
                       for scientific or publicity purposes, your demographic information is never published in conjunction with your video. </p>
-                      <p> As we figure out the details of General Data Protection Regulation (GDPR) compliance, we're restricting demographic information collection to families outside the EU. </p>
+                      <p> We're currently restricting demographic information collection to families outside the EU to ensure GDPR compliance. </p>
                   </h1>
 
                 </div>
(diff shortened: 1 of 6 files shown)


web/templates/web/demographic-data-update.html
```diff
@@ -12,25 +12,89 @@
...
                 } else {
                     hideField('state');
                     $(`#id_state`)[0].value = '';
                 }
+                // Then show/hide most demographic fields based on EU or not
+				if (gdpr_country_list.includes($('#id_country')[0].value)) {
+					$.each(gdprFieldsToHide, function(i, fieldName) {
+						hideField(fieldName);
+						$(`#id_${fieldName}`)[0].value = '';
+						$(`#id_${fieldName} input[type=checkbox]`).prop("checked", false );
```
pattisdr: Minor: I would consolidate these three lines into a function that can be called both here and [here](https://github.com/lookit/lookit-api/pull/98/files#diff-b8d298b6a4a06a533d94e07abb55422cR70)

web/templates/web/demographic-data-update.html
```diff
@@ -12,25 +12,89 @@
...
+            
+            if ($('#id_country')[0].value !== 'US') {
+                hideField('state');
+            }
+            
+            if (gdpr_country_list.includes($('#id_country')[0].value)) {
+            	$.each(gdprFieldsToHide, function(i, fieldName) {
+            		hideField(fieldName);
+            		$(`#id_${fieldName}`)[0].value = '';
+            		$(`#id_${fieldName} input[type=checkbox]`).prop("checked", false );
```
pattisdr: Nice job removing data that might be in this field 👍

web/templates/web/demographic-data-update.html
```diff
@@ -54,10 +119,12 @@
         });
         function hideField(field) {
             $(`#id_${field}`).hide().prop('required',false);
+            $(`#id_${field}`).parent().hide(); //hide the parent form group
             $(`label[for=id_${field}]`).hide();
         }
         function showField(field) {
             $(`#id_${field}`).show();
+            $(`#id_${field}`).parent().show(); // show the parent form group
```
pattisdr: What is the parent form group?

web/templates/web/demographic-data-update.html
```diff
@@ -12,25 +12,89 @@
 {% block content %}
     <script>
         $(document).ready(function() {
-            if ($('#id_country')[0].value !== 'US') {
-                hideField('state');
-            }
+        
+			const gdpr_country_list = [
```
pattisdr: Would there be any benefit to pulling this list from a third-party source that updates if this country list changes?  I see you're also adding "", NO, IS, LI, so maybe not.

web/templates/web/demographic-data-update.html
```diff
@@ -12,25 +12,89 @@
...
+            if (gdpr_country_list.includes($('#id_country')[0].value)) {
+            	$.each(gdprFieldsToHide, function(i, fieldName) {
+            		hideField(fieldName);
+            		$(`#id_${fieldName}`)[0].value = '';
+            		$(`#id_${fieldName} input[type=checkbox]`).prop("checked", false );
+            	});
+            }
+            
             $('#id_country').on('change', function() {
+            	// First show/hide state field based on USA or not
```
pattisdr: Just for my understanding, why do you have to show/hide state field first?

kimberscott: Sorry, bad comment. What I meant was just do state fields and everything else *separately* for clarity. (It was briefly structured as "are you in the US? if show, show state. If not, are you in the EU? Show/hide personal data fields accordingly." so that fields weren't appropriately showing back up if someone selected US.)

web/templates/web/demographic-data-update.html
kimberscott: Probably, I just didn't find a good source. I suspect as GDPR takes effect this will be a fairly common need and there will be a good option, so I'll keep that in mind.

web/templates/web/demographic-data-update.html
kimberscott: Fixed, thanks!

web/templates/web/demographic-data-update.html
```diff
@@ -119,12 +115,17 @@
         });
         function hideField(field) {
             $(`#id_${field}`).hide().prop('required',false);
-            $(`#id_${field}`).parent().hide(); //hide the parent form group
+            $(`#id_${field}`).parent('.form-group').hide(); // hide the parent form-group div
             $(`label[for=id_${field}]`).hide();
         }
+        function clearField(field) { // hide field and remove any stored data
+            hideField(fieldName);
```
pattisdr: Drive-by comment, variables need to be consistent.  I would change "field" argument to fieldName.

kimberscott: Darnit thank you! that's what I get for not testing "trivial" changes locally...

PR #101, opened 2018-06-11
[Do not merge into prod] Feature/Download Endpoints for Response Data

pattisdr:
❗️ Do not merge into production yet
# Purpose

Large Downloads are failing with 'Failed - Network error' in Chrome.  This is mostly happening on the All Responses page where you can download all responses to your study in JSON and CSV.  We are currenly showing a preview of all the json or csv responses to your study on the page, and then pushing that code block into a json file or csv file for download.  As the number of study responses grows, this is becoming unwieldy and slow. 

# Changes

Added four endpoints for downloading data. You must have permission to view study responses to hit these endpoints.
- `/studies/<study_id/responses/all/download_json/`
- `/studies/<study_id/responses/all/download_csv/`
- `/studies/<study_id/responses/demographics/download_json/`
- `studies/<study_id/responses/demographics/download_csv/`

The download buttons on the All Responses page now send requests to these endpoints and the data is downloaded to your machine.  The data preview has been removed.

screen shot 2018-06-10 at 9 02 10 pm

# Testing Notes
This is a potential solution for the Failed - Network error in Chrome @kimberscott.  It improved download speed quite a bit for me locally. I've removed the preview sections here - the page load is going to get more and more laggy if we do not do this. Alternatively, if you want to leave the preview data, you can still hit these endpoints directly to download, without navigating to the preview pages.  You should make sure that only researchers can hit these download links, and that you must have proper permissions to the study to download the data.  Would verify that the downloaded data is what you expect.

### DIFF
--- a/exp/urls.py
+++ b/exp/urls.py
@@ -22,7 +22,10 @@ from exp.views import (ExperimenterDashboardView, ParticipantDetailView,
                        StudyAttachments, StudyBuildView, StudyCreateView,
                        StudyDemographics, StudyDetailView, StudyListView,
                        StudyParticipantEmailView, StudyPreviewBuildView,
-                       StudyResponsesAll, StudyResponsesList, StudyUpdateView)
+                       StudyResponsesAll, StudyResponsesAllDownloadJSON,
+                       StudyResponsesAllDownloadCSV, StudyResponsesList,
+                       StudyDemographicsDownloadJSON, StudyDemographicsDownloadCSV,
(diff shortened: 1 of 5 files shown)


kimberscott:
This is lovely, and much faster for me locally as well. Thank you so much, Dawn!  It makes sense to me to remove the previews and just go to file download. 

I verified locally that non-researchers and researchers without read access to a particular study can't access any of the four endpoints. Merging into develop now to verify on staging.

PR #113, opened 2018-07-06
[feature] docker image cleanup

icereval:
A nightly task to remove dangling docker images.


studies/tasks.py
```diff
@@ -283,6 +283,14 @@ def cleanup_checkouts(older_than=None):
     cleanup_old_directories(checkouts, older_than)
 
 
+@app.task
+def cleanup_docker_images():
+    logger.debug('Cleaning up docker images...')
+    images = subprocess.run(['docker', 'images', '--quiet', '--filter', 'dangling=true'], stdout=subprocess.PIPE)
+    for image in images.stdout.splitlines():
+        return subprocess.run(['docker', 'rmi', '--force', image])
+
```
kimberscott: When there are multiple images, this looks like it'll clean up only the first one - is this the intended behavior? Or just subprocess.run(...) in the loop, and return afterwards?

icereval: :+1:, lol I had fixed this locally when validating in a shell but forgot to apply the change. Thanks for the review!

PR #195, opened 2018-11-30
Vastly improved dockerfile

Datamance:
Cheating the layering system and doing some dirty hacks to get the final build size down.


Datamance:
Here, we have improved Experiment build vastly by leveraging Docker Image Layer Caching to get faster incremental builds and instantaneous repeat builds. Rather than suffer a storage penalty for all the caching we're doing, however, we've done some clever things in the Dockerfile to mitigate space consumption per build.

PR #211, opened 2019-01-03
Bugfix/210 hide build

### LINKED ISSUE #210: Separate 'build' from 'deploy' processes
Building dependencies should not automatically make participation possible at /studies/[STUDYID]/ - that URL should only proxy to the built study on Google Cloud if the study is active. That way researchers can actually take down a study and not accept participants even if they have already shared a link (e.g., because they've found a bug they need to fix first, or because they've finished collecting data and don't want to keep paying people). 

* The email text upon successful dependency build should change from "[Study] has been deployed. This study is located [here]." to "Dependencies have been built for [Study]. When this study is approved and activated, participants will be able to access it [here]."

* The study participation proxy view should dispatch a 404 if the study is not currently in "Active" state.

Datamance:
Fixes #210

### DIFF
--- a/studies/templates/emails/notify_researchers_of_deployment.html
+++ b/studies/templates/emails/notify_researchers_of_deployment.html
@@ -3,10 +3,10 @@
 </p>
 <p>
   {% if action == 'previewed' %}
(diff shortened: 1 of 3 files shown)


Datamance:
fixed #210

PR #221, opened 2019-01-07
Move back to old build style to use pipe env vars

### LINKED ISSUE #220: Pipe configs not propagating into new ember builds
It looks like the mistargeting of staticfile/ember builds related to #209 is uncovering some issues we weren't seeing previously.

Here you can see the [old build.sh](https://github.com/lookit/lookit-api/blob/8df19c804cd0d46fb202de92a450d8b3633697b4/ember_build/build.sh) pulling in lots of things that were probably related to/necessary for proper pipe configuration:

```bash
sed -i "s/prepend: ''/$REPLACEMENT/g" /checkout-dir/ember-cli-build.js
sed -i "s/VideoRecorder.swf/$RECORDER_REPLACEMENT/g" /checkout-dir/lib/exp-player/addon/services/video-recorder.js

# Copy in required files
cp /environment /checkout-dir/.env
cp /VideoRecorder.swf /checkout-dir/public/
```

All that stuff is gone because when I was testing my earlier changes, the actual (fresh) experiment builds were going into a nested directory, so anything other than JSON changes weren't actually reflected in the new builds. 

We're going to fix this by adding these instructions back into our image builds.

Datamance:
This will fix #220 in staging.

### DIFF
--- a/ember_build/build.sh
+++ b/ember_build/build.sh
@@ -1,6 +1,32 @@
 #!/bin/sh
-echo "*** BEGINNING BUILD ***"
-mkdir -p deployments/$STUDY_UUID && cd deployments
-cp -Rv /checkouts/frameplayer_exp_bundle/dist/* $STUDY_UUID
-chown -Rv www-data:www-data $STUDY_UUID
-echo "*** DONE ***"
+# copy git checkout into local env for faster development
+# could probably be removed in production
+cp -r $CHECKOUT_DIR /checkout-dir/
+
+sed -i "s/prepend: ''/$REPLACEMENT/g" /checkout-dir/ember-cli-build.js
+sed -i "s/VideoRecorder.swf/$RECORDER_REPLACEMENT/g" /checkout-dir/lib/exp-player/addon/services/video-recorder.js
+
+# Copy in required files
+cp /environment /checkout-dir/.env
+cp /VideoRecorder.swf /checkout-dir/public/
+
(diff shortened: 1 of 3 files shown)


ember_build/build.sh
```diff
@@ -1,6 +1,32 @@
...
-mkdir -p deployments/$STUDY_UUID && cd deployments
-cp -Rv /checkouts/frameplayer_exp_bundle/dist/* $STUDY_UUID
-chown -Rv www-data:www-data $STUDY_UUID
-echo "*** DONE ***"
+# copy git checkout into local env for faster development
+# could probably be removed in production
+cp -r $CHECKOUT_DIR /checkout-dir/
+
+sed -i "s/prepend: ''/$REPLACEMENT/g" /checkout-dir/ember-cli-build.js
+sed -i "s/VideoRecorder.swf/$RECORDER_REPLACEMENT/g" /checkout-dir/lib/exp-player/addon/services/video-recorder.js
```
kimberscott: we don't need (or have) VideoRecorder.swf anymore, can safely remove this line

ember_build/build.sh
```diff
@@ -1,6 +1,32 @@
...
+
+# Copy in required files
+cp /environment /checkout-dir/.env
+cp /VideoRecorder.swf /checkout-dir/public/
```
kimberscott: can safely remove `cp /VideoRecorder.swf /checkout-dir/public/`

Issue #208, opened 2019-01-03
UI improvements for study edit views


kimberscott:
A) View - Edit study

~1) Add a preview button that just links to the study preview at bottom of "Study editor" section. Button should be disabled if preview dependencies aren't built yet, w/ help text saying to build them in order to preview study.~

~2) Add some help text to the "Change study type & dependencies" section to explain what this is & how it works. (Kim can provide text)~

~3) Display preview dependency build status under "Change study type & dependencies"~

~4) Format minimum & maximum age in days/months/years as side-by-side fields and restore help text explaining to researchers how these are actually used to calculate an age range in days~

~5) A warning shows up when saving changes to a study that has already been approved - "Are you sure you want to modify the JSON [...]?" Since this is triggered by *any* changes, not just JSON, change warning text to "Are you sure you want to modify the study [...]?"~

~6) Move "Manage Researchers" to the Study Detail view~

~7) Use modal instead of banner message for invalid JSON error upon saving as this is easy to miss.~

~8) Check that commit SHAs are valid (or blank) upon saving dependencies & prevent save w/ warning if not~

~9) If commit SHA is blank and gets resolved to the latest master commit, store that commit sha in the study so researchers can tell which version they're using! See problem introduced here: https://github.com/lookit/lookit-api/commit/1c6399c9b3fe560e323c2c3d4e71c6f6b19c9830#commitcomment-31840951 Ideally we would fully fold in #120~

B) View - Study detail

1) "Previewed:" -> "Preview dependencies:" to be directly analogous to "Dependencies" status

2) Move "Build study" button next to "Dependencies: built/not built" for clarity

3) Move "Edit study," "View responses," "Email participants," & "Clone study" all in one area where "View responses" and "Clone study" are now, instead of putting edit/email up top

4) Display study participation URL (i.e. lookit.mit.edu/studies/[STUDYID]/) in addition to just the study ID so researchers can find it w/o going back through email notifications

5) Hide comments section unless selected action is "reject" (?)

6) Possibly: use modal instead of banner message for build action ("You will be notified..") and for "activate" error (if dependencies are not built) b/c it's easy to miss these at the top of the page

~7) Log should say something like "Dependencies built" instead of "deploy"~

8) Proposed changes to study state change names:
   "Activate" -> "Start data collection"
   "Pause" -> "Pause data collection"
   "Submit" -> "Submit for review"
[...]

PR #230, opened 2019-01-17
208 ui improvements study edit kim review

kimberscott:
Upon review of #228 : one typo fix, several aesthetic changes for study detail view while I was looking at it anyway, consolidation of study forms mostly for reuse of help text.

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -33,19 +33,36 @@ class BaseStudyForm(ModelForm):
         return cleaned_data
 
 
-# Base form for creating a new study or editing an existing study
-class StudyEditForm(BaseStudyForm):
+# Form for creating a new study or editing an existing study
+class StudyForm(BaseStudyForm):
+
+    structure = forms.CharField(label='Build Study - Add JSON',
+                                widget=AceOverlayWidget(mode='json', wordwrap=True, theme='textmate', width='100%',
+                                                        height='100%', showprintmargin=False), required=False,
+                                help_text='Add the frames of your study as well as the sequence of those frames.  This can be added later.')
+
+    def clean_structure(self):
+        structure = self.cleaned_data['structure']
+        try:
+            json_data = json.loads(structure)  # loads string as json
+        except:
+            raise forms.ValidationError("Save failed due to invalid JSON! Please use valid JSON and save again. If you reload this page, all changes will be lost.")
+        return json_data
 
     class Meta:
         model = Study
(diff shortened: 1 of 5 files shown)


studies/forms.py
```diff
@@ -62,53 +79,24 @@ class Meta:
...
-        fields = StudyEditForm.Meta.fields + ['structure']
-
-        labels = StudyEditForm.Meta.labels.copy()
-
-        help_texts = StudyEditForm.Meta.help_texts.copy()
+class StudyEditForm(StudyForm):
+    
+    class Meta(StudyForm.Meta):
+        help_texts = StudyForm.Meta.help_texts.copy()
+        help_texts['study_type'] += '''<p class="help-block"> Once you've filled in the required metadata, you'll have to build the dependencies for your
```
Datamance: Do we need to have HTML in here? Let's keep HTML out of class members

studies/forms.py
```diff
@@ -62,53 +79,24 @@ class Meta:
...
             'long_description': 'Explain the purpose of your study here.',
             'contact_info': 'This should give the name of the PI for your study, and an email address where the PI or study staff can be reached with questions. Format: PIs Name (contact: youremail@lab.edu)',
-            'criteria': 'Text shown to families - this is not used to actually verify eligibility.'
+            'criteria': 'Text shown to families - this is not used to actually verify eligibility.',
+            'study_type': '''<p>After selecting a study type above, you'll be asked 
+            to fill out some study type metadata as well. This metadata is unique to the 
+            study type, and provides important configurations for building your study.</p>
+            <p>If you're not sure what to enter here, just leave the defaults (you can change this later).</p>
+            <p>For more information on study types and their metadata, please
+            <a href="https://lookit.readthedocs.io/en/develop/experimenter.html#editing-study-type">see the documentation.</a></p>'''
```
Datamance: Let's take the inline HTML out of class members - either hardcode it in the view or create a partial that knows how to deal with (iterates over) these segments of help text

Datamance: Actually forget what I said - this is small potatoes and apparently there's precedent for doing it this way https://docs.djangoproject.com/en/2.1/ref/models/fields/#help-text

studies/forms.py
Datamance: See below and https://docs.djangoproject.com/en/2.1/ref/models/fields/#help-text

Datamance:
fixes #208

PR #233, opened 2019-01-28
Feature/128 session record consent backend

### LINKED ISSUE #128: Only create new session record upon completing consent frame
This would both make the collected data much easier for researchers to interpret (instead of having a separate "record" every time someone refreshed the setup page) and work more in line with how a participant might expect. It would also mean we'd only have conditions assigned in cases where the user proceeded through consent, meaning that relative counts ("we have enough of condition A but need more of condition B") would be more accurate. Plus participants who'd refreshed the setup page a lot wouldn't run into increasingly long load times as the system worked to look up every record and send in groups of 10. In principle, this could be done by either not saving the session data at all until consent, or by only granting access to an admin unless consent was completed. This would likely need to be handled at the level of generating the experiment from the JSON spec - wait for a frame that has an isConsent designation, and every frame thereafter sends data, OR a consent frame sets some property of the session data that's then accessed to check whether to send data each transition.

Datamance:
Fixes #128


api/views.py
```diff
@@ -254,6 +260,7 @@ def get_queryset(self):
             Response.objects.filter(
                 Q(study__id__in=study_ids) | Q(child__id__in=children_ids)
             )
+            .filter(completed_consent_frame=not updating_session)
```
kimberscott: Hmm - just reading code, haven't tested this. Does this mean that if we're not updating session, we require completed_consent_frame=True, and if we ARE updating session, we require completed_consent_frame=False? The latter is NOT something we want, because people need to be able to continue issuing PATCH requests as the study continues after consent. Basically the filter should just only be applied in case we're not updating the session.

Datamance: Resolving this per discussion, but would like one more pass-over for review!

Datamance:
OK, I believe I have the code functionally how we want it now. I've left the stuff we talked about and had previously commented so that I can easily uncomment it and re-write the appropriate tests to actually test what we want should you disagree with the path I've taken.

Slack chat here for reference:
_________________ **SLACK CHAT BELOW** ____________________

OK! We’re down to the last little bit of nitpickery here. Right now, `testGetResponsesListByOwnChildren` is failing because the GET-specific response queryset of the *non-nested* (i.e. non-study-specific) route is just

```
response_queryset = Response.objects.filter(
    study__id__in=ids_of_viewable_studies, completed_consent_frame=True
 )
```

as opposed to the PATCH-specific queryset, which lumps those two together and also says “both those conditions together OR just any response with children belonging to the user.”

Right now, we are _technically_ ok with this, because the parent can acquire the response ID of their child through the *nested* API *if they know the study ID too.*

But now what that means is that the GET functionality of the *un-nested* `/responses` facet of the api and the *nested* `/study/{STUDY_ID}/responses/` API _behave differently_: the former WON’T give you consented+unconsented responses of your own kids, whereas the latter WILL give you all your kids’ responses.

If you think it’s a good idea (I am now coming around the corner to this and think it’s a good idea FWIW), we can reconcile this difference by getting rid of the method-based queryset dispatch in the non-nested block and just have both GET and PATCH use the same queryset filter that you came up with:

```
response_queryset = Response.objects.filter(
                    Q(child__id__in=children_belonging_to_user)  # Case #1
                    | (
                        Q(study__id__in=ids_of_viewable_studies)  # Case #2
                        & Q(completed_consent_frame=True)
                    )
                )
```

This will reconcile the behavior of the two API facets, which will enable the test to pass again. The difference is that now, on the *un-nested* GET calls (i.e. parents/researchers consuming the Responses API directly - no study ID required), they will be able to see responses of their own children.

PR #238, opened 2019-02-28
Feature/136 consent coding

### LINKED ISSUE #136: GUI for consent coding on Experimenter platform + central storage of consent info
*Pain point*: The first step in coding data from a Lookit session is to check that the associated consent video shows a parent making a statement of informed consent. It is possible for data to be collected in the absence of an informed consent statement – for instance, a parent might not read the statement because he or she does not understand written English. Currently, it is up to each lab to come up with a consent coding workflow and avoid viewing any session video until consent has been confirmed. Researchers can download individual videos, all consent videos, or all videos. This manual task is potentially error-prone and represents a lot of duplication of work across labs. A simple GUI for coding consent and central storage of consent information would enforce a clear consent coding process, reduce potential for dangerous human error, and reduce the burden on individual labs. 

*Acceptance criteria*: Researchers can use a view on the experimenter interface to see consent videos, which they can filter by new, marked as non-consent, and marked as consent. Each video is displayed on the page along with some basic information about the associated session, including existing consent information if available. Next to the video, the researcher selects whether the consent video is valid or not (possibly from a short list of reasons why not, to distinguish tech issues from non-reading), and can enter a brief string with any notes. The consent value, note, and identity of the researcher who provided them are stored with the session record. Only videos from sessions where consent has been confirmed may be downloaded. 

*Implementation notes/Suggestions*: Consent judgments should be possible to overwrite, e.g. in case an RA thinks something isn't valid but it is or vice versa. Storing history of such judgments would be nice to have but not critical.

Datamance:
Merge to staging. Fixes #136 . God save the queen.


api/views.py
```diff
@@ -85,30 +85,64 @@ def get_queryset(self):
...
+        if self.request.user.has_perm("accounts.can_read_all_user_data"):
+            return demographics_for_active_users
+
+        studies = get_objects_for_user(
+            self.request.user, "studies.can_view_study_responses"
+        )
+
+        consented_responses = get_consented_responses_qs().filter(study__in=studies)
+
+        demographics_ids = consented_responses.values_list("demographic_snapshot", flat=True)
```
kimberscott: Just a note for the future - we could legally/ethically give access to demographic snapshots associated with any users associated with consented_responses, i.e. researchers are allowed to see updates to demographic data that the family makes after participation. This behavior is more straightforward and in line with what researchers get through the experimenter UI though.

studies/migrations/0043_add_video_constructs.py
```diff
@@ -0,0 +1,205 @@
...
+                            response=response,
+                            is_consent_footage=frame_id in VALID_CONSENT_FRAMES,
+                        )
+                    )
+                    seen_ids.add(video_id)
+
+    return Video.objects.bulk_create(video_objects)
+
+
+class Migration(migrations.Migration):
```
kimberscott: Is the idea that once this is deployed to production we’ll just have researchers redo consent coding via the UI?  Pros and cons of creating ‘accepted’ rulings for existing records with ‘comments’ field set to something like ‘automatically accepted and may not be verified; recorded before consent manager deployed’?

Datamance: We absolutely could create ConsentRulings in bulk with that comment, we'd just need to acquire/create some sort of fake user/bot that would serve as the arbiter for each

Datamance: alternatively, we could use your account

studies/migrations/0043_add_video_constructs.py
```diff
@@ -24,8 +24,31 @@ def generate_videos_from_responses(apps, schema_editor):
...
         generate_videos_from_events(response, VideoModel)
+        apply_initial_consent_ruling(response, ConsentRulingModel, consent_admin)
+
+
+def apply_initial_consent_ruling(response, consent_ruling_model, consent_admin):
+    """
+
+    :param response: a Response object
+    :param consent_ruling_model: the ConsentRuling model
+    :param kim: Kim
```
kimberscott: -> :param consent_admin

Issue #241, opened 2019-03-13
BCC the sender of an email through the participant email view


kimberscott:
**TL;DR**: BCC the sender of an email through the participant email view.

**Narrative**
As an Experimenter, I would like to be able to keep track of emails I send through the participant email view. 

**Acceptance Criteria**
An email sent via exp/studies/[studyid]/email/ is sent to the participant as currently, and the email address in the "sender" field is BCCed.

**Implementation Notes**
This is a stopgap measure until #130 - we just want a few researchers currently using Lookit to be able to use the experimenter platform rather than also needing manually-generated spreadsheets of email addresses.

PR #242, opened 2019-03-19
bcc the sender address when emailing participants on Lookit; stopgap …

kimberscott:
…measure to allow researchers some record of what has already been sent, as per https://github.com/lookit/lookit-api/issues/241. If we can deploy this along with the merge-elf-and-exp-addons changes that would be great.


Datamance:
LGTM

PR #259, opened 2019-03-26
Delete withdrawn data

### LINKED ISSUE #129: Delete or make unavailable withdrawn data
*Pain point*: At the conclusion of any study, parents select a privacy level for the video (in-lab use only, scientific/educational sharing allowed, or publicity use allowed) and whether data may be shared on Databrary. Parents also have the option to withdraw video data entirely. This option is rarely used, but is important because we are filming in families’ homes and sensitive information could be accidentally disclosed during recording. When a parent selects that they would like to withdraw video, this information is included in information given to researchers, but the video is not automatically deleted - that is left to the researcher. This introduces unnecessary potential for human error.

*Acceptance criteria*: Video from any session that a parent withdraws video from is deleted from all Lookit storage automatically, and is not available to the researcher running the study. Other data from the session is still available.

*Implementation notes/Suggestions*: Ideally we might introduce a slight delay (e.g. 1-7 days) before deleting video from Lookit servers, in case the parent withdrew accidentally and asks us to restore the video; however, this is likely much more trouble than it's worth.

Datamance:
Fixes #129 

Starting the pull request now just to have some code up for review


studies/models.py
```diff
@@ -776,6 +777,26 @@ def generate_videos_from_events(self):
...
+    if not response.sequence:
+        return
+
+    current_frame = response.sequence[-1]
+
+    if any(postfix in current_frame for postfix in VALID_EXIT_FRAME_POSTFIXES):
+        # check for withdrawal
+        withdrawal = response.exp_data[current_frame].get("withdrawal", False)
+        if withdrawal:
+            response.videos.delete()  # Should trigger a pre-delete hook.
```
kimberscott: We should actually not include consent footage in the videos to be deleted (for now - leaving this possibility open in the future would be sensible)

studies/models.py
```diff
@@ -940,6 +961,24 @@ def download_url(self):
...
+
+      " Note that the delete() method for an object is not necessarily called when deleting objects in bulk using a
+        QuerySet or as a result of a cascading delete. To ensure customized delete logic gets executed, you can use
+        pre_delete and/or post_delete signals.
+      "
+    """
+    video = instance  # Again, instance is hooked as a kwarg.
+    delete_video_from_cloud.apply_async(
+        video.full_name, countdown=60 * 60 * 24 * 7
+    )  # Delete after 1 week.
```
kimberscott: Deleting after 1 week is very sensible, will also want to act directly on researcher permissions for videos so they're not accessible in the interim

studies/models.py
Datamance: done

studies/models.py
Datamance: Since the video models are deleted, there's no real way for the views to know about what's in S3 (i.e. we aren't using `attachment_helpers` anymore)

studies/helpers.py
```diff
@@ -36,3 +41,56 @@ def send_mail(
...
+                return
+        else:
+            return
+
+        try:
+            return getattr(self, frame_type.lower())(
+                response, frame_data, *args, **kwargs
+            )
+        except AttributeError as e:
+            logger.warning(f"{e.args} is not registered for frame-specific action.")
```
kimberscott: given that the default frame type is 'DEFAULT', want to only warn if it's not registered for frame-specific action and isn't DEFAULT? (Just thinking of readability of these logs.)

studies/helpers.py
```diff
@@ -36,3 +41,56 @@ def send_mail(
...
+        pass
+
+    def exit(self, response, frame_data: dict, *args, **kwargs):
+        """Exit frame hook.
+
+        Args:
+            response: Response Model.
+            frame_data: dictionary of frame data.
+        """
+        withdrawal = frame_data.get("withdrawal", False)
```
kimberscott: should probably at least log warning rather than silently defaulting to False if frame_data has no "withdrawal" property

studies/models.py
```diff
@@ -776,6 +779,20 @@ def generate_videos_from_events(self):
         return Video.objects.bulk_create(video_objects)
 
 
+@receiver(post_save, sender=Response)
+def take_action_on_exp_data(sender, instance, created, **kwargs):
+    """Performs post-save actions based on the current frame.
```
kimberscott: Just a warning about relying on this hook for marking consent footage - if saving the response after a single frame fails for whatever reason, the participant is allowed to continue and the frameplayer will just try again next frame (which will mean adding multiple frames at once). So we might want to consider the edge case where save fails immediately after the consent frame, but then succeeds later, so that it's never the most recent frame. This is less relevant for EXIT type frames, which should come at the end of the study.

studies/models.py
kimberscott: Are there definitely no other cases where we'd want to delete a video model without deleting the actual footage?

studies/tasks.py
```diff
@@ -410,3 +412,12 @@ def build_zipfile_of_videos(
...
     )
+
+
+@app.task(bind=True)
+def delete_video_from_cloud(task, s3_video_name):
+    """Delete videos in S3.
+
+    Meant to have a delay of about 7 days.
+    """
+    S3_RESOURCE.Object(settings.BUCKET_NAME, s3_video_name).delete()
```
kimberscott: What happens if the video isn't found / was already deleted? Is that error a problem for anything?

kimberscott: Also, what (if anything) is the process for cancelling this task? (We should write some description of how this works for lookit-docs, both for researchers to understand in what sense the data is gone / in what edge cases they might briefly have access to it, and for us to reference in case anyone ever says "I didn't mean to withdraw my data!")

studies/helpers.py
Datamance: good catch

studies/tasks.py
Datamance: it'll just bork the celery task, which is nbd. I can add logging if you want but honestly the error raised by the boto3 client will be just as informative as anything I can provide

kimberscott: Actually looks like boto3 might not even raise an error? But either way, should be fine.

studies/models.py
Datamance: Accounting for the "not very cautious contributor" case

studies/models.py
Datamance: Fair point. It's enough to keep it as is.

studies/helpers.py
```diff
@@ -85,11 +85,14 @@ def exit(self, response, frame_data: dict, *args, **kwargs):
...
         """
-        withdrawal = frame_data.get("withdrawal", False)
+        withdrawal = frame_data.get("withdrawal", None)
         if withdrawal:
-            response.videos.filter(
-                is_consent_footage=False
-            ).delete()  # Should trigger a pre-delete hook.
+            for video in response.videos.filter(is_consent_footage=False):
+                video.delete(delete_in_s3=True)
+        else:
```
kimberscott: this warning will also be triggered whenever withdrawal=False - we want to check explicitly for None (or just for that key not existing), and otherwise trigger the withdrawal actions iff appropriate

Datamance: got it

kimberscott:
Confirmed withdrawn video records being deleted as expected on staging!

PR #266, opened 2019-04-04
Quick query as well

Datamance:


web/views.py
```diff
@@ -312,10 +312,15 @@ def get_queryset(self):
...
-        return Study.objects.filter(id__in=study_ids)
+        responses = Response.objects.filter(child__id__in=children_ids)
+
+        study_ids = responses.values_list("study_id", flat=True)
+
+        return Study.objects.filter(id__in=study_ids).prefetch_related(
+            Prefetch("responses", queryset=responses),
+            "responses__child",
+            "responses__feedback",
+        )
```
kimberscott: Will this be all responses from this child or limited to completed_consent_frame = True? (we do want the latter b/c that aligns with participant expectations of "a separate time doing a study" whereas just going to config page doesn't - also aligns with whether the researchers could potentially have access)

Datamance: Ahhhh right completing consent frame != consent. Fixing!

PR #271, opened 2019-04-10
fix bug where past responses provided to frameplayer via /studies/[ST…

kimberscott:
…UDYID]/responses/?{Query string with pagination and child id} included responses for any children associated with this account doing any study (plus, for researchers w permission to view study responses, any consented responses for this study). Now always restricts to this study in nested path, and if query string with child is provided to nested path, only provides responses from that child (w/ same behavior regardless of whether user is researcher).

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -276,14 +276,14 @@ class ResponseViewSet(FilterByUrlKwargsMixin, views.ModelViewSet):
             study_uuid = self.kwargs["study_uuid"]
             study = get_object_or_404(Study, uuid=study_uuid)
 
-            nestedResponses = study.responses
+            nested_responses = study.responses
 
             # CASE 1: Participant session, using query string with child ID.
             # Want same functionality regardless of whether user is a researcher.
-            childID = self.request.query_params.get("child", None)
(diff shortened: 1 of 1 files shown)


api/views.py
```diff
@@ -275,31 +275,41 @@ def get_queryset(self):
...
             study = get_object_or_404(Study, uuid=study_uuid)
-            consented_responses = study.consented_responses
-            if self.request.user.has_perm(
-                "studies.can_view_study_responses",
-                get_object_or_404(Study, uuid=study_uuid),
-            ):
-                return (
-                    Response.objects.filter(
+
+            nestedResponses = study.responses
```
Datamance: nitpick - snake_case please

api/views.py
```diff
@@ -275,31 +275,41 @@ def get_queryset(self):
...
+
+            # CASE 1: Participant session, using query string with child ID.
+            # Want same functionality regardless of whether user is a researcher.
+            childID = self.request.query_params.get("child", None)
```
Datamance: snake_case for variables

PR #311, opened 2019-08-07
Study History contains downloadable session videos

### LINKED ISSUE #123: Allow families to see their videos, & add some additional details to past studies
*Pain Point*: Once parents participate in a study, there is no immediate confirmation that it "worked" and Lookit has their data, or any later confirmation that it has been used to do cool science. They also do not have automatic access to their own data, although they often want to see their videos; instead, a researcher has to provide it if desired, which is labor-intensive and introduces unnecessary possibilities for human error (e.g. sending the wrong child's video). 

*Acceptance Criteria*:  
- Parents can download any video clips from a session from the "Past Studies" view when logged in. (Nice but not required if they can actually just see the clips embedded in the browser.)
- Parents can see some basic information about past study sessions, including which child participated, status of consent coding (yes/no/pending/no video(?)), research group PI/contact.
- Parents should *not* see sessions with completedConsentFrame=false displayed
- Parents can see the "status" of a study that they participated in, as provided by the researchers - e.g., data collection, analysis, writing up results, here's a link to the paper/media article/etc.

*Implementation notes/Suggestions*: 
- We've discussed adding a freeform string field to the study model, editable by researchers in the study edit view, representing the current state of the study.
- Eventually (not as part of this ticket) parents might also be able to (going forward) view any study debriefing text from "Past Studies," since this may contain individual results or condition info. Having set up an "exit survey" frame type to support withdrawal of video will make this easier, in that we'd be able to see if there's a frame of type exit-survey in the expData for this response, and if so, fetch its debriefing text. 
- Eventually (not as part of this ticket) parents might also be able to view/download the consent form text. This is now stored as part of the consent form in expData.

Datamance:
Fixes #123

### DIFF
--- a/attachment_helpers.py
+++ b/attachment_helpers.py
@@ -41,7 +41,7 @@ def get_download_url(video_key):
     return S3_CLIENT.generate_presigned_url(
         "get_object",
         Params={"Bucket": settings.BUCKET_NAME, "Key": video_key},
-        ExpiresIn=60,
+        ExpiresIn=600,
     )
 
 
--- a/web/templates/web/studies-history.html
+++ b/web/templates/web/studies-history.html
@@ -21,13 +21,6 @@
     {% endif %}
 {% endblock %}
 {% block content %}
-    <script>
-        function imgError(image, index) {
-            image.style.display = 'none';
-            var placeholder = document.getElementById('image-placeholder-' + index);
-            placeholder.style.display = 'block';
-        }
-    </script>
     <div>
         <div class="lookit-row lookit-page-title">
             <div class="container">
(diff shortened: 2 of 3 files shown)


attachment_helpers.py
```diff
@@ -41,7 +41,7 @@ def get_download_url(video_key):
     return S3_CLIENT.generate_presigned_url(
         "get_object",
         Params={"Bucket": settings.BUCKET_NAME, "Key": video_key},
-        ExpiresIn=60,
+        ExpiresIn=600,
```
kimberscott: Will this also affect consent coding? (That'd be great!)

Datamance: yah it will! We should still fix the UI signaling though (for that, this is just kicking the can down the road lol)

kimberscott:
Upon trying on staging - could the videos shown be in order, so parents see first the consent video, then each study video in turn?

Datamance:
What order is it showing in now - reverse or off by n?

kimberscott:
Reverse-ish, but I'm not confident yet that it's actually reversed and not random.

kimberscott:
Yep, random I think (or at least not ordered, reverse-ordered, or off-by-n.) I'll send you an example

PR #312, opened 2019-08-07
Query DSL replacing EligibilityModel workflow

### LINKED ISSUE #141: Add checkboxes for various conditions/circumstances for children & families, in child & demographic surveys.
*Pain point*: Many researchers are interested in working with specific populations, but families do not have a standardized way to indicate whether they fall into various groups in a way that would support use of this information for recruitment/eligibility.

*Acceptance criteria*: 
* In the child and demographic forms, parents can indicate things that apply to the child/family (yes/no or checkboxes), from a standard list + "other." E.g., autism, Down syndrome, Williams syndrome, perinatal stroke, vision/hearing impairments, twin, multilingual. [Kim to provide an initial list.] 
* Researchers can use these fields in their study eligibility criteria - e.g., this study is for 4- to 6-month olds who were born before 37 weeks and are being raised in multilingual households. 
* Researchers can use presence/absence of a *sibling* with various criteria in study eligibility. E.g., this study is for 1- to 3-year olds with a sibling who is at least 6 years old, or for children with at least 3 siblings total, or for children without any younger siblings, or for 8- to 11-month-olds with a sibling with autism. (The last form is in fact the primary concrete motivation for this capacity, but being able to use presence of a sibling with an arbitrary condition and/or presence/absence of siblings will be useful to other researchers in the future.) 

*Implementation notes/Suggestions*: MIT OGC advises us that including medically-relevant options  is fine from a legal standpoint (MIT is not a covered entity under HIPAA).

Datamance:
WIP, fixes #141


kimberscott:
I see we're checking the criteria string to check that the form is as expected (so that I get e.g. `Invalid criteria expression: No terminal defined for 'X' at line 1 col 28 deaf AND has_too_many_cats XOR is_silly ^ Expecting: {'OR', 'GT', 'AND', 'LT', 'LTE', 'NE', 'GTE', 'EQ'} `). How complicated would it be to also check the tokens are valid, so that typos don't lead to unpredicted behavior? 

For instance I can currently set eligibility to `deaf AND has_too_many_cats` without error -  upon trying to participate everyone's rejected (although `deaf OR has_too_many_cats` lets deaf kids participate)

accounts/queries.py
```diff
@@ -0,0 +1,277 @@
...
+    }
+    expanded_child.update(expanded_languages)
+
+    ga_enum = expanded_child.pop("gestational_age_at_birth")
+
+    gestational_age_in_weeks = _gestational_age_enum_value_to_weeks(ga_enum)
+
+    expanded_child["gestational_age_in_weeks"] = gestational_age_in_weeks
+
+    return expanded_child
```
kimberscott: Can we add n_languages here?

accounts/queries.py
```diff
@@ -0,0 +1,277 @@
...
+        Returns:
+            The string form of the intended python expression.
+
+        Raises:
+            GrammarError if a gender test does not obey the contract.
+        """
+        if (
+            name == "gender"
+            and comparator not in ("=", "!=")
+            and other not in ("f", "m", "o", "na")
```
kimberscott: I think this should be `name == "gender" and ( comparator not in ("=", "!=") or other not in ("f", "m", "o", "na"))` - not all of these anded together. Right now `gender = female` and `gender > o` are both fine.

accounts/queries.py
```diff
@@ -0,0 +1,277 @@
...
+        ):
+            raise GrammarError(
+                'Gender criteria must fit format "gender (=|!=) (m|f|o|na)"'
```
kimberscott: Just a question about intended behavior - is this error propagated up to display to the researcher? (I can't get it to show up even if I do `gender<purple`)

accounts/tests.py
```diff
@@ -1,3 +1,101 @@
...
+        self.assertTrue(
+            get_child_eligibility(
+                self.older_deaf_child_with_dyslexia, self.compound_and_condition
+            )
+        )
 
-# Create your tests here.
+        self.assertFalse(
+            get_child_eligibility(self.french_twin, self.compound_and_condition)
+        )
```
kimberscott: :) thanks for all this!

studies/forms.py
```diff
@@ -188,10 +212,11 @@ class EligibleParticipantQueryModelForm(ModelForm):
 
     gender_specification = forms.IntegerField(required=False)
 
-    # TODO: validation function
+    # # TODO: validation function
```
kimberscott: actually still todo?

accounts/forms.py
```diff
@@ -211,11 +211,13 @@ class Meta:
...
             "gestational_age_at_birth": "Gestational Age at Birth",
             "additional_information": "Any additional information you'd like us to know",
+            "existing_conditions": "Characteristics and conditions",
+            "languages_spoken": "Languages this child is exposed to at home, school, or with another caregiver.",
         }
 
         help_texts = {
             "given_name": "This lets you select the correct child to participate in a particular study. A nickname or initials are fine! We may include your child's name in email to you (for instance, \"There's a new study available for Molly!\") but will never publish names or use them in our research.",
-            "additional_information": "for instance, diagnosed developmental disorders or vision or hearing problems",
+            "additional_information": "For instance, diagnosed developmental disorders or vision or hearing problems",
```
kimberscott: can we add a help text for gestational age at birth field, "Please round down to the nearest full week of pregnancy completed"

accounts/queries.py
Datamance: gah - totally right. Let's fix that.

studies/forms.py
Datamance: Nah but I'll be deleting this form anyway - just waiting to have everything tested in dev for the query DSL. I can get rid of it now if it bothers you though :)

accounts/queries.py
Datamance: I'm going to fix this by making the grammar more strict

studies/forms.py
kimberscott: Nope it's fine! Just didn't realize this was that form.

kimberscott:
Docs link for criteria expression - https://lookit.readthedocs.io/en/develop/researchers-set-study-fields.html#criteria-expression

accounts/queries.py
Datamance: That will be slightly more tricky but I think I can do it

kimberscott: wouldn't it just be something like `expanded_child["n_languages"] = sum([bool for (lang, bool) in expanded_languages.items()])`?

Datamance: Got it done.

Datamance: did something similar

kimberscott: Ohhh haha I see it was not as simple as adding another field at all. Thanks!

Datamance: Your way would have worked too - we'd just need to still add the grammar rule in the SDT

Issue #310, opened 2019-08-06
Invoke tasks for local development


robo360:
**TL;DR**: Once sentence description of what you want.
Invoke tasks to set up all dependencies to easily get started on contributing to the project
**Narrative**
How will this help developers do their job better?
The invoke tasks will help developers getting started, as all dependencies will be installed and configured automatically. The getting started process that usually takes some days will no longer be an issue. 

**Acceptance Criteria**
This literally just needs Rico sign off.

- [x] Rico sign-off

Datamance:
Moved this to block 5 since it is not critical path.

We can close this out once @Mayowa99 and @robo360 do their presentations at the Boston Python meetup.

Datamance:
Closing this out as the script was technically completed, though we are fixing issues with it now.

PR #315, opened 2019-08-08
Feature/310 add invoke tasks for development

robo360:
In response to issue #310


Pipfile
```diff
@@ -57,6 +56,8 @@ psycopg2 = "==2.7.5"
 google-cloud-storage = "==1.14.0"
 docker = "==3.7.2"
 django-bitfield = "==1.9.3"
+django-sslserver = "*"
```
Datamance: Shouldn't these be under the dev dependencies heading?

project/settings/defaults.py
```diff
@@ -81,6 +80,7 @@
     "allauth",
     "allauth.account",
     "allauth.socialaccount",
+    "sslserver",
```
Datamance: Shouldn't this only be listed as a middleware when DEBUG = True? 

alongside this block on line 98
```python
if DEBUG:
    INSTALLED_APPS += ["debug_toolbar", "shells"]
    MIDDLEWARE_CLASSES += [
        "debug_toolbar.middleware.DebugToolbarMiddleware",
        "pyinstrument.middleware.ProfilerMiddleware",
    ]
```

tasks.py
```diff
@@ -0,0 +1,426 @@
+from invoke import task, run
```
Datamance: Why isn't invoke listed under the dev dependencies header of the Pipenv file? Strange that this was even able to run!

tasks.py
```diff
@@ -0,0 +1,426 @@
+from invoke import task, run 
+import platform 
+import os 
+
+PLATFORM=platform.system()
```
Datamance: Please run `black` on this file - it will autoformat for you so you don't have to worry about manually following PEP8

tasks.py
```diff
@@ -0,0 +1,426 @@
...
+SERVE_HTTPS_SERVER="python manage.py runsslserver --certificate certs/local_lookit.mit.edu.pem --key certs/local_lookit.mit.edu-key.pem"
+SERVE_HTTP_SERVER="python manage.py runserver"
+SERVE_NGROK="ngrok http https://127.0.0.1:8000"
+BASE_DIR=os.getcwd()
+PATH_TO_CERTS=os.path.join(BASE_DIR, 'certs')
+
+
+@task
+def system_setup(c,verbose=False):
+    """
```
Datamance: Running `black` should fix your doc formatting issues, but in case it doesn't: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Your docstrings should look like that

tasks.py
```diff
@@ -0,0 +1,426 @@
...
+    packages=('',"django-sslserver")
+    for package in packages:
+
+        if (run("pipenv install {}".format(package), hide=not verbose, warn=True).ok):
+            run("echo \"===>{} {}\"".format(package, MESSAGE_OK))
+        else:
+            run("echo \"===>{} {}\"".format(package, MESSAGE_FAILED))
+
+@task 
+def pygraphviz(c, verbose=False):
```
Datamance: Are we using PyGraphViz, or not? It looked like you were removing it from the dependency list.

Pipfile
robo360: You are right!

project/settings/defaults.py
robo360: Yeah! I had ignored that there is more to the api than just development.

tasks.py
robo360: what is **black**? First time to hear that!

tasks.py
Datamance: ^^^ make sure to do that everywhere, actually

tasks.py
robo360: I had already installed it through pip. That's why I had it. Anyways, it should be added when I run these commands in the env from pipenv. 

This brings the question of how these tasks will be used. I was thinking of this order. 
1. Pipenv shell 
2. Pipenv install invoke 
3. invoke setup 

.... after looking at this, I am wondering why I am installing pipenv as part of one of the tasks. I think I should go ahead and remove pipenv from packages to be installed.

Datamance:
I merged quite a few branches into develop, so it looks like you'll need to resolve some conflicts in the dependency files.

Thankfully, this should be pretty easy :) it's not like you need to mess around with actual code.

project/settings/defaults.py
robo360: I think it should be listed as an INSTALLED_APPS. I looked around there is nothing in the readme that shows it running as middleware, so I do not know how I can list there. All that said, I have been looking through the documentation, as I realized there is a big chunk of django that I have no idea how it works.

tasks.py
robo360: Gotcha!

tasks.py
robo360: It should be there now.

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+        run(
+            "sudo apt-get install build-essential curl git", hide=not verbose, warn=True
+        )
+        if run("command -v brew", hide=not verbose, warn=True).ok:
+            run('echo "===>brew {}"'.format(MESSAGE_OK))
+        else:
+            run('echo "===>brew {}"'.format(MESSAGE_FAILED))
+
+    if PLATFORM == "Darwin":
+        if run("command -v cask ", hide=not verbose, warn=True).ok:
```
Datamance: Won't this always fail? `cask` isn't the executable, `brew` is

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+                run('echo "===>Cask {}"'.format(MESSAGE_FAILED))
+
+    # creating a local setting file
+    if run('cd project/settings && touch local.py && echo "DEBUG=True" > local.py').ok:
+        run('echo "===>Successfully configured settings for local development"')
+
+
+@task
+def install_dependencies(c, verbose=False):
+    """
```
Datamance: Can you fix your docstrings to follow google docstring format

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+
+        if run("pipenv install {}".format(package), hide=not verbose, warn=True).ok:
+            run('echo "===>{} {}"'.format(package, MESSAGE_OK))
+        else:
+            run('echo "===>{} {}"'.format(package, MESSAGE_FAILED))
+
+
+@task
+def pygraphviz(c, verbose=False):
+    """
```
Datamance: see above comment on docstrings (do this for all your docstrings)

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+
+    if PLATFORM == "Linux":
+        # Installing graphviz
+        if run(
+            "sudo apt-get update && sudo apt-get install graphviz",
+            hide=not verbose,
+            warn=True,
+        ).ok:
+            run('echo "Graphyviz {}"'.format(MESSAGE_OK))
```
Datamance: s/Graphyviz/Graphviz/g

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+        else:
+            run('echo "Graphyviz {}"'.format(MESSAGE_FAILED))
+        # Installing pygraphviz
+        if run(
+            "sudo pip3 install pygraphviz --install-option='--include-path=/usr/include/graphviz' --install-option='--library-path=/usr/lib/graphviz/'",
```
Datamance: See below comment about graphviz - do we really have the same behavior on linux as with OS X? Might make sense to test this inside a docker container (without the specified include paths - I vaguely remember that you do actually need those for OS X)

tasks.py
```diff
@@ -0,0 +1,646 @@
...
+            hide=not verbose,
+            warn=True,
+        ).ok:
+            run('echo "===>pygraphyviz {}"'.format(MESSAGE_OK))
+        else:
+            run('echo "===>pygraphyviz {}"'.format(MESSAGE_FAILED))
+
+    elif PLATFORM == "Darwin":
```
Datamance: leave a comment detailing why we're targeting different library and include paths. I'm guessing this has to do with how mac OS X deals with the underlying Graphviz dependency (the C lib), but it might help future readers to be explicit as to why we're doing things this way.

tasks.py
robo360: I assume you meant I should write correct "Graphyviz" with "graphviz."

Datamance: https://vim.fandom.com/wiki/Search_and_replace

s/foo/bar/g --> "replace all instances of foo with bar"

robo360: Got it!

tasks.py
robo360: Cask is actually already in brew, so I do not need to check whether I have it. All I need to do is to make sure that I have "brew."

robo360: Just to make sure it is there, I will "brew tap," which is a way of adding new repositories for brew to install or keep track off.

tasks.py
robo360: Yeah I tested it inside a linux-based docker container, and I can confirm it does not need graphviz.

tasks.py
robo360: Can you explain to me the reason as welll? cz I really do not know why we must provide the path to Graphviz in the installation process. I know that pygraphviz allows you to manipulate stuff made by Graphviz in python. But I am not sure why exactly we must provide the path.

Datamance: http://pygraphviz.github.io/documentation/pygraphviz-1.5/install.html#providing-path-to-graphviz

```We tried our best to discover graphviz location automatically, but if you would like specify specific location for graphviz you may provide additrional parameters to specify graphviz location```

File: tasks.py [hunk 13]
```diff
@@ -0,0 +1,857 @@
...
+        For debugging purposes, set verbose (bool) to True to print the stdout responses for the run process.
+
+    Usage: 
+        invoke pygraphviz or invoke pygraphviz --verbose
+
+    """
+
+    if PLATFORM == "Linux":
+        # Installing graphviz
+        """
```
Datamance: Go ahead and delete this stuff if it's not being used

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+        if run("sudo pip install pygraphviz", hide=not verbose, warn=True).ok:
+            run('echo "===>pygraphviz {}"'.format(MESSAGE_OK))
+        else:
+            run('echo "===>pygraphviz {}"'.format(MESSAGE_FAILED))
+
+    elif PLATFORM == "Darwin":
+        """
+
+        Pygrapviz on Darwin usually fails to locate graphviz to solve the issue,
+        the install-option falg is used to include the path.
```
Datamance: s/falg/flag/g

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+        run("sudo rabbitmqadmin list queues", warn=True, hide=not verbose)
+    else:
+        run("echo {}".format(MESSAGE_WRONG_PLATFORM))
+
+    return None
+
+
+@task
+def postgresql(c, verbose=False):
+    """ postgresql invoke task.
```
Datamance: Same comment as before. It's nitpicky but it keeps things looking clean and professional

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+            run('echo "=====>certificates {}"'.format(MESSAGE_FAILED))
+    else:
+        run("echo {}".format(MESSAGE_WRONG_PLATFORM))
+
+    return None
+
+
+@task
+def ngrok(c, verbose=False):
+    """ngork invoke task.
```
Datamance: s/ngork/ngrok/g

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+    Returns:
+        None.
+
+    usage: 
+        invoke celery 
+
+    """
+    run(SERVE_CELERY)
+
+    return None
```
Datamance: Don't need to return None; it's implicit in Python

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+        None.
+    
+    Note: 
+        This func does not serve celery and ngrok. 
+
+    usage: 
+        invoke setup
+
+    """
+    return None
```
Datamance: instead of `return None` do `pass`

Datamance: Or you could print something nice like "All finished!"

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+        None.
+
+    Usage: 
+        invoke ngrok_service
+
+    """
+
+    run(SERVE_NGROK)
+
+    return None
```
Datamance: return None is implicit, don't need it

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+            if run("brew cask install docker", hide=not verbose, warn=True).ok:
+                run('echo "===>docker {}"'.format(MESSAGE_OK))
+            else:
+                run('echo "===>docker {}"'.format(MESSAGE_FAILED))
+        run("open /Applications/Docker.app")
+
+    else:
+        run("echo {}".format(MESSAGE_WRONG_PLATFORM))
+
+    return None
```
Datamance: return None is implicit, don't need it

wherever you have return None at the end of a function you can get rid of it

Pipfile
```diff
@@ -60,7 +60,10 @@ psycopg2 = "==2.7.5"
...
+django-debug-toolbar = "*"
 lark-parser = "==0.7.2"
+lark = "*"
```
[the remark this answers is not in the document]
robo360: Gone!

Pipfile
```diff
@@ -60,7 +60,10 @@ psycopg2 = "==2.7.5"
 google-cloud-storage = "==1.14.0"
 docker = "==3.7.2"
 django-bitfield = "==1.9.3"
+django-sslserver = "*"
```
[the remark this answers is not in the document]
robo360: I think I will move it from the requirements/default to requirements/dev as well.

tasks.py
```diff
@@ -211,6 +211,12 @@ def pygraphviz(c, verbose=False):
             run('echo "===>pygraphviz {}"'.format(MESSAGE_FAILED))
 
     elif PLATFORM == "Darwin":
+        """
+
+        Pygrapviz on Darwin usually fails to locate graphviz to solve the issue,
```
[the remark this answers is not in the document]
robo360: Ok

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+    Note:
+        For debugging purposes, set verbose (bool) to True to print the stdout responses for the run process.
+
+    usage: 
+        invoke install-dependencies or invoke install-dependencies --verbose
+
+    """
+
+    run("echo '***INSTALLING ALL DEPENDENCIES***'")
+    packages = ("", "django-sslserver")
```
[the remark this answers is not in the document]
robo360: good point!

tasks.py
```diff
@@ -0,0 +1,857 @@
...
+SERVE_HTTPS_SERVER = "python manage.py runsslserver --certificate certs/local_lookit.mit.edu.pem --key certs/local_lookit.mit.edu-key.pem"
+SERVE_HTTP_SERVER = "python manage.py runserver"
+SERVE_NGROK = "ngrok http https://127.0.0.1:8000"
+BASE_DIR = os.getcwd()
+PATH_TO_CERTS = os.path.join(BASE_DIR, "certs")
+
+
+@task
+def system_setup(c, verbose=False):
+    """system-setup invoke task.
```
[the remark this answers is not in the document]
robo360: done on every function

tasks.py
Datamance: It's ok for this to be commented out for now but let's delete this if it's not being used

tasks.py
```diff
@@ -0,0 +1,833 @@
...
+        # Installing pygraphviz
+        if run("sudo pip install pygraphviz", hide=not verbose, warn=True).ok:
+            run('echo "===>pygraphviz {}"'.format(MESSAGE_OK))
+        else:
+            run('echo "===>pygraphviz {}"'.format(MESSAGE_FAILED))
+
+    elif PLATFORM == "Darwin":
+
+        # Pygrapviz on Darwin usually fails to locate graphviz to solve the issue,
+        # the install-option falg is used to include the path.
```
Datamance: the flag misspelling is here too

tasks.py
```diff
@@ -0,0 +1,833 @@
...
+    This func installs postgresql, create the database of the API and create required tables. 
+
+    Args:
+        c (obj): Context-aware API wrapper & state-passing object.
+        verbose (bool): states whether stdout should be printed.
+
+    Returns:
+        None.
+
+        However, this func echoes MESSAGE_FAILED, MESSSAGE_OK, or MESSAGE_ALREADY_INSTALLED
```
Datamance: too many S's in MESSAGE_OK

tasks.py
```diff
@@ -0,0 +1,833 @@
...
+            + SERVE_HTTP_SERVER,
+            hide=False,
+        )
+
+
+@task
+def ngrok_service(c):
+    """
+
+    This func serves ngrok.
```
Datamance: Move this up to where the summary sentence goes

Issue #323, opened 2019-08-19
Most recently viewed accepted response's details displayed persistently in Consent Manager


kimberscott:
**Description**
The Consent Manager is not supposed to show full session data for pending or rejected responses. However, if the researcher views an accepted response (A) and then a rejected or pending response (B), the full session data for response A is displayed when viewing B, which is misleading.

**How to reproduce**
View an accepted response, then a rejected response.

**Expected behavior**
Full session data should be blank for rejected and pending responses, even after viewing an accepted response

**Screenshots**

**Environment (OS, browser, branch, etc.):**
prod

**Additional context**

PR #324, opened 2019-08-19
Preclude meaningless getter call by forcing non-null/undefined value

Datamance:
Ah, the joys of javascript/jQuery.

Per the API documentation for the .html() method (https://api.jquery.com/html/), an undefined first argument will "dispatch" the getter behavior rather than the setter behavior. We are forcing a value to make sure that the setter is always dispatched.


Datamance:
derp, fixes #323

PR #328, opened 2019-08-20
Merge Develop into Master

### LINKED ISSUE #123: Allow families to see their videos, & add some additional details to past studies

### LINKED ISSUE #138: Tool to upload feedback and other manually researcher-generated session data
*Pain point*: Currently researchers need to use the API to leave feedback to parents about particular sessions. Feedback is one of the better ways we have to build engagement (let parents know a human watches & appreciates their video!) and we want it to be easy for researchers to leave. 

*Acceptance criteria*: 
* Without using the API directly, researchers can provide feedback per session that is associated with session data and (as it is now) accessible to the family. (This should indeed be stored as part of the response, *not* emailed.)  
- Ideally this *can* be done via file upload, even if there is also a nice interface for it, so as not to stand in the way of whatever analysis processes individual labs develop.
* Ideally, researchers can upload arbitrary text data that is stored with/linked to the session data, e.g. a manual "score" for the first part of a longitudinal study that is used by the second session to generate appropriate stimuli. 

*Implementation notes/Suggestions*: 

- Feedback is currently implemented sort of like consent rulings, as a separate model; there can be multiple feedbacks for the same response, perhaps left by different researchers. This is fine but if it complicates things, condensing down to a single string feedback field on the response or making it so there is at most one feedback associated with a response would also work fine. (Migrations would be a little annoying but could just join together multiple feedbacks if needed without messing up any existing .)

- MIT is the only group currently using the feedback API, but we do use it and will need to update code to accommodate changes (see https://github.com/kimberscott/lookit-data-processing/blob/coding-workflow-multilab/scripts/experimenter.py) 

- Feedback would ideally be possible to add/edit via the consent manager, "individual responses" view, and by uploading a copy of the study CSV they downloaded with a "feedback" column updated. It should be possible to add feedback on sessions where the consent ruling is 'rejected' or 'pending.' 

- The arbitrary data should be cleaned to protect the database, and multiple attributes can be stored in a single standard JSON field if that's helpful. This could also be somehow joined with the "feedback" model if easier, e.g. have a "researcher-generated session data" model that's associated with a session and includes feedback and also this other data.

### LINKED ISSUE #141: Add checkboxes for various conditions/circumstances for children & families, in child & demographic surveys.

Datamance:
This is a big Kahuna!

Fixes #138 
Fixes #123 
Fixes #141 

-- each of the above referenced issues should be closed out after this merge.


kimberscott:
Could we link to the docs (https://lookit.readthedocs.io/en/develop/researchers-set-study-fields.html#criteria-expression) in the study edit form here?
Screen Shot 2019-08-20 at 2 44 10 PM

kimberscott:
Could we also indicate in the save study warning that changes to the eligibility criteria do not auto-reject the study (like toggling discoverability)?
Screen Shot 2019-08-20 at 2 45 54 PM

kimberscott:
On staging I'm currently only ever able to see a single video clip in the "past studies" view, even when there are multiple accessible by the experimenter, do you see the same behavior?

kimberscott:
And could we put a little indicator in the "-> Next video" button (or elsewhere nearby) on the "Past studies" view that says which video it is, like "[Video 2 of 5] -> Next video"

Datamance:
```On staging I'm currently only ever able to see a single video clip in the "past studies" view``` --> probably an artifact of the video being small and not being able to see the index of the given video. Addressing this in the changelist.

Datamance:
> Could we link to the docs (https://lookit.readthedocs.io/en/develop/researchers-set-study-fields.html#criteria-expression) in the study edit form here?
> Screen Shot 2019-08-20 at 2 44 10 PM

👍

kimberscott:
On staging I'm no longer seeing the age-based warnings when I try to participate with a child outside the age range for a study, although I do see the warning based on the criteria expression.

kimberscott:
Also, it's no longer possible to select "Not sure or prefer not to answer" on gestational age at birth in the child form - can we add a true "blank" in place of that like for the gender dropdown?

PR #338, opened 2019-10-04
Feature/134 participant analytics

Datamance:
Please feel free to pull down the branch and do any last check-ups before we push this to staging!


exp/views/study.py
```diff
@@ -1326,6 +1318,73 @@ def dispatch(self, request, path, *args, **kwargs):
...
+        child_age_in_days = (datetime.date.today() - resp.child.birthday).days
+        languages_spoken = popcnt_bitfield(
+            int(resp.child.languages_spoken), "languages"
+        )
+        response_data.append(
+            {
+                "Response (unique identifier)": resp.uuid,
+                "Child (unique identifier)": resp.child.uuid,
+                "Child Age in Days": child_age_in_days,
+                "Child Age in Months": round(child_age_in_days / 30, 2),
```
kimberscott: rounding to two decimal places gives approx. same precision as days - can just do int(child_age_in_days / 30) to get rounded down to nearest integer. (I think that'll be more straightforward to interpret than actual rounding, esp. for years.)

Datamance:
So now that we have more data, I'm seeing two separate issues regarding participation plots:

1) Axes aren't lining up in the two participation plots. This is almost certainly due to the cumulative plot *not* getting day-rounded and counted per day, which precludes us from doing day-precision point interpolation that matches up exactly with the binned plot. (This is actually happening with the registration plots as well but it's not so important there)

2) What appears to be improper caching, replicated as such:
- pick a study-specific timeseries
- unselect it
- pick an "all studies" timeseries

The observed effect is that the cumulative participation plot does not update to the correct new all-studies TS data - instead, it seems to plot the old study-specific TS data. This does *not* happen if we go in the other direction - e.g., activating an "all-studies" line prior to activating a study-specific one.

PR #339, opened 2019-10-07
add a researcher login link at the bottom for people who are willing …

kimberscott:
Because people remain confused about this no matter how many times I say that they need to go to [prod/staging]/exp/ to log in. I don't necessarily want to make it more prominent or include info in the participant login, because with substantially more participants than researchers we will end up confusing more people total, but some link on the page may be a reasonable compromise.


Datamance:
👍

PR #376, opened 2019-10-23
framework for Recruitment Manager perms

Datamance:
Adds two important granular permissions, one for allowing access to the Analytics View link and view itself, and the other allowing for data from all responses/studies from within that view. There are also a few other permissions I've added in case we want to be restrictive in granular ways later with researchers.


accounts/migrations/0044_new-analytics-permissions.py
```diff
@@ -0,0 +1,65 @@
...
+
+    dependencies = [("accounts", "0043_gestational-age-can-be-blank")]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="child",
+            options={
+                "ordering": ["-birthday"],
+                "permissions": [
+                    ("can_view_child_analytics_data", "Can View Child Analytics Data")
```
kimberscott: Can the more human-readable name be something like Can View All Children from Analytics for clarity (since everyone can view child analytics data, the question is which children)

accounts/migrations/0044_new-analytics-permissions.py
```diff
@@ -0,0 +1,65 @@
...
+                "permissions": (
+                    ("can_create_users", "Can Create User"),
+                    ("can_view_users", "Can View User"),
+                    ("can_edit_users", "Can Edit User"),
+                    ("can_remove_users", "Can Remove User"),
+                    ("can_view_user_permissions", "Can View User Permissions"),
+                    ("can_edit_user_permissions", "Can Edit User Permissions"),
+                    ("can_read_all_user_data", "Can Read All User Data"),
+                    ("can_read_usernames", "Can Read User Usernames"),
+                    ("can_view_user_analytics_data", "Can View User Analytics Data"),
```
kimberscott: we don't use can_view_user_analytics_data, correct? It seems like it might be complicated anyway because the pivot table uses demographic, child, & response info - not actually anything directly stored on the user model anyway, and the demo/child/response info aren't especially separable

accounts/migrations/0044_new-analytics-permissions.py
```diff
@@ -0,0 +1,65 @@
...
+                ],
+            },
+        ),
+        migrations.AlterModelOptions(
+            name="demographicdata",
+            options={
+                "ordering": ["-created_at"],
+                "permissions": [
+                    (
+                        "can_view_demographics_analytics_data",
```
kimberscott: we don't use this, correct? (see comment below about why this might be complicated anyway)

accounts/models.py
```diff
@@ -267,6 +268,7 @@ class Meta:
             ("can_edit_user_permissions", _("Can Edit User Permissions")),
             ("can_read_all_user_data", _("Can Read All User Data")),
             ("can_read_usernames", _("Can Read User Usernames")),
+            ("can_view_user_analytics_data", "Can View User Analytics Data"),
```
kimberscott: (again, unused right?)

accounts/models.py
```diff
@@ -523,6 +528,12 @@ class DemographicData(models.Model):
     extra = DateTimeAwareJSONField(null=True)
 
     class Meta:
+        permissions = [
+            (
+                "can_view_demographics_analytics_data",
+                "Can View Demographics Analytics Data",
```
kimberscott: (again, unused right?)

studies/migrations/0054_new-analytics-permissions.py
```diff
@@ -0,0 +1,24 @@
...
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="response",
+            options={
+                "base_manager_name": "related_manager",
+                "ordering": ["-demographic_snapshot__created_at"],
+                "permissions": (
+                    ("view_response", "View Response"),
+                    ("view_response_analytics_data", "View Response Analytics Data"),
```
kimberscott: Can the human-readable name of this one also be clearer, e.g. View All Response Data in Analytics or View Responses from All Studies in Analytics

accounts/migrations/0044_new-analytics-permissions.py
Datamance: We don't; see initial PR description/comment - I just put these here in case we wanted to leverage them later. We could theoretically use them (this one in particular, for, say, precluding certain users from seeing registration data) but you're probably right in that it's not likely. I'll nix them.

accounts/migrations/0044_new-analytics-permissions.py
Datamance: See above

accounts/models.py
Datamance: See above

accounts/migrations/0044_new-analytics-permissions.py
Datamance: That's a much better name - will do

kimberscott:
Whoops, I entirely missed the PR comment about the additional permissions, sorry. (But yeah it does seem like we should just add those later based on the actual use case if we need them.)

PR #381, opened 2019-11-01
Reduce query load by using .values()

Datamance:
The tricky part here was in dealing with the related videos - we still
have to do a bit of testing to see if there's no way to keep the
prefetch as it existed rather than doing the manual join in our app
code.

We've also lost some slight distinction in Consent Ruling comments
between "no previous ruling" and "no comment in previous ruling" due to
the nature of the optimized query.

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -35,8 +35,12 @@ from studies.fields import (
 )
 from studies.forms import StudyEditForm, StudyForm
 from studies.helpers import send_mail
-from studies.models import Feedback, Response, Study, StudyLog, StudyType
-from studies.queries import get_annotated_responses_qs, get_study_list_qs
+from studies.models import ACCEPTED, Feedback, Response, Study, StudyLog, StudyType
+from studies.queries import (
+    get_annotated_responses_qs,
+    get_responses_with_current_rulings_and_videos,
+    get_study_list_qs,
+)
 from studies.tasks import build_zipfile_of_videos, ember_build_and_gcp_deploy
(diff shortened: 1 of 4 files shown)


exp/views/study.py
```diff
@@ -910,53 +914,60 @@ def get_context_data(self, **kwargs):
...
+            total_stats["children"].add(response["child_id"])
 
-            response_data = response_key_value_store[str(response.uuid)] = {}
+            response_json = response_key_value_store[str(response["uuid"])] = {}
 
-            response_data["videos"] = [
+            response_json["videos"] = [
                 {"aws_url": video.download_url, "filename": video.filename}
-                for video in response.consent_videos
+                for video in response["videos"]
```
kimberscott: it looks like response["videos"] has only the consent videos due to how get_responses_with_current_rulings_and_videos works - confirm?

studies/queries.py
```diff
@@ -17,14 +19,90 @@ def get_annotated_responses_qs():
...
+        )
+        annotated_query = annotated_query.annotate(
+            ruling_comments=Coalesce(comment_subquery, models.Value("N/A"))
+        )
+
+    return annotated_query
+
+
+def get_responses_with_current_rulings_and_videos(study_id):
+    """Gets all the responses for a given study, including the current ruling and videos.
```
kimberscott: consent videos only - add to the docstring?

exp/views/study.py
Datamance: confirmed

Issue #337, opened 2019-10-02
Long latency on requests for individual responses pages


kimberscott:
**Description**
It takes 10-20s to load each tab of the "Individual studies" pages
(/exp/studies/[N]/responses/attachments/, /exp/studies/[N]/responses/demographics/, /exp/studies/[N]/responses/all/, /exp/studies/[N]/responses/?sort=-date_modified, /exp/studies/[N]/responses/consent_videos/) 

**How to reproduce**
Go to any of the paths listed above - e.g. from the study detail view, click "View Responses."

**Expected behavior**
These should take a few seconds at most.

**Screenshots**
<img width="1366" alt="Screen Shot 2019-10-02 at 1 45 00 PM" src="https://user-images.githubusercontent.com/9554627/66074164-956d5880-e526-11e9-931e-7379e3c444c6.png">

**Environment (OS, browser, branch, etc.):**
Staging, prod

**Additional context**
@Datamance says: 
> it’s almost certainly all the in-python joins that have to be done with prefetching
> the ORM is doing a lot of work to rectify all those relationships
> The optimization will require some pretty ugly query construction, I think

Just listing as a "bug" to make it easier for researchers concerned about this to find it, but it's not a proper bug.

PR #385, opened 2019-11-04
Override bad queryset from StudyResponsesMixin

Datamance:
This should have a reasonably salient effect in prod.


Datamance:
This should fix #337

Issue #321, opened 2019-08-16
Reduce potential for accidental disclosure of personal information


kimberscott:
**TL;DR**: Make it easy to download data with potentially identifying information obscured, and scary to download things you should be careful with.

**Narrative**
As an Experimenter, I want to establish a reproducible data analysis pipeline and share the results of my research with others. However, some of the data I download may contain personal information about participants that needs to be removed before I can share it, and some files may not be appropriate to share at all. I want to get as little personal information as possible in the "default" files available for me to download. I also want to be reminded to be careful when downloading files that require special care.

**Preliminary decisions to make**
We could alternately provide age in days at time of participation but (by default) not include any timestamps. This would support even infancy studies and prevent researchers from having to combine child data with session data to get exact ages. But we would need to remove timestamps of sessions, demographic data, maybe consent rulings from the file downloads, and we'd need to make sure timestamps were removed from all event data in ember-lookit-frameplayer (e.g. by measuring from the start of the experiment). 

**Implementation Notes**
It will be important to be very clear to researchers about how age rounding works, and we probably want the rounding results to be independent of when in the year the child was born (i.e., we do not want to go by calendar months). To the nearest 10 or 30 days would be reasonable (or 10 until 360 and thereafter 30).
@kimberscott can write the warning text.

**TODO/Acceptance**

- [x] Add checkboxes on all responses page for including/not including identifying info in all responses (CSV only). Two columns: checkboxes for how to get age (birthdate, age in days, rounded age) and for what other data to include (either other child data, grouped, and other account data; or break down into gender, gest age, conditions, languages, addl info, name). 
- [x] Implement rounded age field/property/fn
- [x] Implement use of checkboxes (including for file naming) for CSV
- [x] Extend to JSON all response data
- [x] Set up individual response data to create JSON, overview CSV, or frame CSV
- [x] Add checkboxes for identifying info on individual response page & implement (CSV, JSON; disabled for frame CSV)
- [x] Add demographic/account data download button & implement. Remove list of birthdays and replace with list of approximate ages of children in home.
- [x] Add warning text to demographic data
- [x] Add warning/explanatory text for all response view
- [x] Participant view: remove/process list of birthdays as in account data download (separate, just for readability)
- [x] Add demographic data dictionary download option
- [x] Add child data download button & implement
- [x] Add child data dictionary download button & implement
[...]

Issue #148, opened 2018-10-03
Improve CSV data provided to researchers


kimberscott:
A CSV file generated automatically shows the session data collected during the study, but the data is highly nested and is not flattened before turning into a CSV. Simple flattening would go a long way to making this data more usable; it would be even better if researchers could select which fields they wanted to include. A (partial) data dictionary could be generated and available for download alongside the CSV. The only problem with doing this after launch is that researchers relying on this data may have scripts that rely on the old version. @kimberscott  may be able to work on it along with frames/documentation. 

- [x] Scoping and plan of which files to create
- [x] Generate flattened session summary CSV info/headers for download
- [x] Generate session summary data dictionary for download
- [x] Hook in data dictionary download to UI (add to dropdown, use radio buttons, or just have separate download buttons)
- [x] Create longer-format session data CSV and hook in to download
- [x] Create longer-format session data data dictionary and hook in to download
- [x] Format download buttons nicely
- [x] Write documentation about use of these files

kimberscott:
Plan, with many thanks to @jchu10: 

Provide download of 4 separate files:
* Session summary CSV: has headers 

```
response_id
response_uuid
response_sequence.[N]
response_conditions.[N].[frameName, each key flattened]
response_completed
response_withdrawn
response_consent_ruling
response_consent_arbiter
response_consent_time
response_consent_comment
study_id
study_uuid
participant_id
participant_uuid
participant_nickname
child_id
child_uuid
child_name
child_birthday
child_gender
child_age_at_birth
child_additional_information
```

* Session summary CSV data dictionary, filled in, since above headers are standard

* Session data CSV, longer format: breaks down response_global_event_timings and response_exp_data. Has headers
```
response_uuid
frame_id (e.g. 5-mood-survey OR global for events in response_global_event_timings)
key (e.g. formData.happy OR event.0.name, event.0.timestamp, etc.)
value (e.g. 6)
```

* Session data CSV data dictionary, partially prepopulated. Show all unique frameID - key pairs and provide space for researcher to give explanation of that value's meaning (e.g., 5-mood-survey / formData.happy: this is how happy the participant is on a scale from 1 to 7, blank means they skipped the question)

kimberscott:
Here's what the download data page looks like now:
![Screen Shot 2019-12-12 at 1 41 00 PM](https://user-images.githubusercontent.com/9554627/70739547-18f4b400-1ce5-11ea-8747-2980009222cc.png)

PR #399, opened 2019-12-11
148: Improve CSV data

kimberscott:
Downloadable CSV data was previously largely unusable for analysis (i.e. contained large chunks of JSON so you needed to process the JSON regardless). Has been split into an overview with one line per response + a detailed file with data in a fairly standard 'long' format, both with data dictionaries / dictionary templates! Addresses #148 and forms a foundation for #321 coming next.

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -334,6 +334,22 @@ class Child(models.Model):
     def age_at_birth(self):
         return GESTATIONAL_AGE_CHOICES[self.gestational_age_at_birth]
 
+    @property
+    def language_list(self):
+        return " ".join(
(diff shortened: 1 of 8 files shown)


kimberscott:
Generating the frame data files causes 502 errors on studies with a lot of responses:

staging example - https://staging-lookit.cos.io/exp/studies/94/responses/all/
prod example - https://lookit.mit.edu/exp/studies/2/responses/all/

Offending code is likely

`exp/mixins/study_responses_mixin.py: StudyResponsesMixin.build_framedata_csv`

and/or how it's used in

`exp/views/study.py: StudyResponsesFrameDataCSV, StudyResponsesFrameDataIndividualCSV, StudyResponsesFrameDataDictCSV`

PR #403, opened 2020-01-06
Fix account query parse with unspecified GA

### LINKED ISSUE #402: Server error when evaluating eligiblity without gestational age listed
**Description**
If any child associated with an account has "don't know or prefer not to answer" listed for gestational age, any gestational-age-based comparisons cause a server error when evaluating eligibility on the study detail page.

**How to reproduce**

1. Register at least one child on your account and choose "don't know or prefer not to answer" for gestational age.
2. Create a study that includes a gestational age comparison in the criteria expression, e.g. `gestational_age_in_weeks <= 37`. Build dependencies, approve, start, and navigate to the study detail page. Can use https://staging-lookit.cos.io/studies/59d84b9d-c9f6-4225-ac9d-be45ac0f313d/ as an example too.
3. Observe the 500 error, associated with this Sentry error https://sentry.io/share/issue/bb94f5a53e25464e8c486859356c8b8a/ (`'<=' not supported between instances of 'NoneType' and 'int'`)
4. Verify that this is the problem: go back to the child, change gestational age to any number, and reload the study detail page. It should load as expected.

**Expected behavior**

The study detail page should load correctly regardless of whether there are any children on the account with unknown gestational age at birth. The "na" value should cause all gestational_age_in_weeks comparisons to evaluate to false, except `gestational_age_in_weeks = na`. 

Staging, production.

Datamance:
Also adds .isort.cfg to play nice with black, fixed some formatting
issues.

Fixes #402


accounts/tests.py
```diff
@@ -173,3 +183,18 @@ def test_num_languages_spoken(self):
...
+                self.unspecified_gestational_age_range_condition,
+            )
+        )
+
+        self.assertFalse(
+            get_child_eligibility(
+                self.child_with_unspecified_gestational_age,
+                self.gestational_age_range_condition,
+            )
+        )
```
kimberscott: these look good to me!

Issue #408, opened 2020-01-22
Too much memory used generating framedata csv downloads


kimberscott:
Generating the frame data files causes 502 errors on studies with a lot of responses:

staging example - https://staging-lookit.cos.io/exp/studies/94/responses/all/
prod example - https://lookit.mit.edu/exp/studies/2/responses/all/

Offending code is likely

exp/mixins/study_responses_mixin.py: StudyResponsesMixin.build_framedata_csv, build_framedata_dict_csv

and/or how they're used in

exp/views/study.py: StudyResponsesFrameDataCSV, StudyResponsesFrameDataIndividualCSV, StudyResponsesFrameDataDictCSV

@kimberscott can likely fix by building csvs incrementally instead of building big objects and then writing all at once.

kimberscott:
@Datamance actually I think I need help on this one; the simple improvements in https://github.com/lookit/lookit-api/pull/409 don't fix it. 

I suspect the main issue is having the CSV file in memory. The response overview CSV does download successfully, suggesting that having `responses` in memory and passing that to `build_summary_csv` is ok. (I'm not sure what makes the framedata dict not work in that case though, since that file should be pretty small.)

Ideas:

* Is this something we can (reasonably) fix (/kick down the road) by allocating more resources on the server? For reference the file generated in this test case (1000 responses) is probably somewhere around 40Mb and 300k lines.

* Can we write to an actual file (e.g. on GCP), the user downloads from there, and then we delete it, instead of having the CSV output in memory like this?

```
        output, writer = self.csv_dict_output_and_writer(headers) # output is io.StringIO()

        for resp in responses:
            this_resp_data = self.get_frame_data(resp)
            writer.writerows(this_resp_data["data"])

        return output.getvalue()
```

kimberscott:
Also weirdly I can download the JSON (83MB) just fine.

kimberscott:
finally fixed in https://github.com/lookit/lookit-api/pull/424 - needed to paginate and use values() to avoid OOM errors and speed things up, then also make the frame-data dict download asynchronous to avoid timing out when processing the ~100MB of frame data.

PR #409, opened 2020-01-22
first shot: just build csvs incrementally instead of making giant lis…

kimberscott:
…t of dicts and then putting into file. Meant to address #408


kimberscott:
(tests do pass despite travis failure)

PR #412, opened 2020-02-03
initial cleanup of StudyResponsesMixin (move small general-purpose fn…

kimberscott:
…s to utils, move functions only used in one view, remove unused code); avoid using default argument values of []; try using .values() to reduce memory demands for frame data download function only

still trying to address #408

### DIFF
--- a/exp/utils.py
+++ b/exp/utils.py
@@ -1,0 +1,62 @@
+import csv
+import io
+
+
+def flatten_dict(d):
+	"""Flatten a dictionary where values may be other dictionaries
+
+	The dictionary returned will have keys created by joining higher- to lower-level keys with dots. e.g. if the original dict d is
+	{'a': {'x':3, 'y':4}, 'b':{'z':5}, 'c':{} }
+	then the dict returned will be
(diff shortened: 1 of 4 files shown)


exp/utils.py
```diff
@@ -0,0 +1,62 @@
...
+            ]
+        else:
+            return [(key, value)]
+
+    items = [item for k, v in d.items() for item in expand(k, v)]
+
+    return dict(items)
+
+
+def csv_dict_output_and_writer(headerList):
```
Datamance: does black not change `headerList` to `header_list`? We may need to bring pyflakes as a dependency to make sure we adhere to PEP8

kimberscott: sorry about that - no, black does not change the variable names, I do run it!

Datamance: argh yeah ok after this PR I'll plug in a dependency and add an `invoke` script to run all the housekeeping goodies. I've been too reliant on my own PyCharm setup

kimberscott: ooh thanks! my apologies, camelcasing variables got drilled into me very early and has been hard to kick

kimberscott:
will do!

PR #424, opened 2020-02-05
minor cleanup mostly to make a PR for review

kimberscott:
Fixes to make the file downloads not run out of memory or time out are finally working & ready for review! 

QUESTION: do we need to be cleaning up video archives and frame data dict files that get put in blobs on GCP, like we do with the docker containers & checkouts?

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -1866,7 +1866,7 @@ class StudyResponsesFrameDataIndividualCSV(StudyResponsesAll):
 
 class StudyResponsesFrameDataDictCSV(StudyResponsesAll):
     """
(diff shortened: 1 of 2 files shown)


Datamance:
> Fixes to make the file downloads not run out of memory or time out are finally working & ready for review!
> 
> QUESTION: do we need to be cleaning up video archives and frame data dict files that get put in blobs on GCP, like we do with the docker containers & checkouts?

I think we addressed this via slack but yeah let's do a timed expiration like we do with video deletion

kimberscott:
Re: clearing old files - decided to go with GCP bucket object lifecycle rules instead - easy to set up and customize per instance if we want to in the future, and one less thing for celery to deal with.

PR #435, opened 2020-02-11
detect consent footage in pipe hook, when creating video object, rath…

kimberscott:
…er than immediately upon consent frame being saved (will deprecate that helper) - at that point the video object hasn't been created yet.


kimberscott:
This is misnamed as a hotfix, it is a bug fix (off develop).

PR #438, opened 2020-02-11
Remaining cleanup from https://github.com/lookit/lookit-api/pull/431:…

kimberscott:
… pull out the actual fields needed instead of using **recipient.__dict__, only use participant_slug


exp/views/study.py
```diff
@@ -598,11 +591,8 @@ def get_context_data(self, **kwargs):
                 },
                 "subject": message.subject,
                 "recipients": [
-                    {
-                        **recipient.__dict__,  # This is terrible. I know :(
-                        "slug": self.participant_slug_from_model(recipient),
-                    }
-                    for recipient in message.recipients.all()
+                    {**recipient, "slug": self.participant_slug(recipient)}
+                    for recipient in message.recipients.all().values("uuid", "nickname")
```
Datamance: is `values()` here not kicking off another query? That's what we want to avoid - the per-message re-issue of the recipient query

kimberscott: Ohh, I'm sorry, I misinterpreted the aim here - will work on this!

kimberscott: actually my reading of https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related is that this is avoided already?

Issue #444, opened 2020-02-18
Travis builds consistently fail


Datamance:
**TL;DR**: Travis builds should just work, for all environments

**Narrative**
Travis is currently dying on feature and bugfix branch builds because of the docker-containerized version of postgres that we're using to create test databases. Prior to January of this year, the problem seemed transient, however it seems that a recent upgrade to the `postgres:9.6` image may have been related to more consistent failure. We may have seen these effects even in our latest failures to deploy to master.

Two bits of information lead me to this conclusion, both from the Matts (plural) over at COS:

Matt Clark: `Took a bit for me to determine that the issue was a minor update to postgres was breaking the test block. I've pinned it to 9.6.16 and develop is now passing.`

w.r.t. the production deployment pipeline, which runs through Jenkins. There's also this bit from Matt Frazier:

`But it's not pulling that from your quay. Appears that it's just pulling the one from docker-library https://github.com/docker-library/postgres`

w.r.t. the non-production branch runs.

Regardless, we should be able to eliminate this as a problem altogether by having Travis manage the instance of postgres on the VM level and just have the lookit docker container do port forwarding with the localhost.

**Acceptance Criteria**

- [ ] Travis build fixed for this branch

**Implementation Notes**
See: https://docs.travis-ci.com/user/database-setup/#postgresql

Datamance:
This is fixed

PR #445, opened 2020-02-18
Fixing travis builds for non-production pipeline branch commits

Datamance:
Hopefully we can just have Travis manage PG!


Datamance:
Fixes #444

PR #450, opened 2020-03-02
Feature/62 pattern preview after participate

kimberscott:
Addresses https://github.com/lookit/ember-lookit-frameplayer/issues/62 from ember-lookit-frameplayer because I originally figured the solution would be just on the frameplayer side.


exp/static/exp/img/preview.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
Datamance: What is this file?

studies/migrations/0060_auto_20200219_1036.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
Datamance: In the future - to avoid having too many migrations per PR, one thing I do is roll back my changes by migrating to the previous migration and then re-create the files so that it "squashes" the migration actions into one file. You can think of it like a `git reset --soft HEAD~${NUM_OF_COMMITS_BACKWARD}`

exp/static/exp/img/preview.svg
kimberscott: preview.svg, or the w3 file? preview.svg is a background image (like the superuser mode image) for the preview items on the individual responses page (like this: https://github.com/lookit/ember-lookit-frameplayer/issues/62#issuecomment-588403569) This may be silly though?

studies/migrations/0060_auto_20200219_1036.py
kimberscott: Oh, good point, thanks!

studies/tests.py
```diff
@@ -1,9 +1,200 @@
...
+        self.study.study_admin_group.user_set.add(self.study_admin)
+        self.study.study_read_group.user_set.add(self.study_reader)
 
-from accounts.models import Child, User
-from studies.models import Response, Study, Video
+        self.children_for_participants = []
+        self.demo_snapshots_for_participants = []
+        self.responses = []
+        self.preview_responses = []
+        for (iPart, part) in enumerate(self.participants):
```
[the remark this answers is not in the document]
kimberscott: Yep! Actually it's not really used anyway anymore so just removed it.

kimberscott:
SQL queries from Django toolbar,

* when deleting 3 preview responses (maybe there is a more sensible way to do this):
Screen Shot 2020-03-07 at 10 37 52 AM

* on preview detail view:
Screen Shot 2020-03-07 at 10 33 35 AM

kimberscott:
@Datamance does this look ok to merge in to dev? I'll need to update the frameplayer code used by each study at the same time, and would like to make sure that process goes smoothly before making any changes on prod.

Datamance:
@kimberscott this is fine for now - if the perf during deletions becomes a problem we can deal with it then. Let's go ahead and merge it

Issue #451, opened 2020-03-03
PLACEHOLDER: CI/CD Pipeline


Datamance:
Placeholder for CI/CD Pipeline work. Add finished tasks as necessary

- [x] cloudbuild.yaml
- [x] staging DNS flip
- [x] Staging Postgres flip
- [x] Prod Postgres flip
- [x] Sunset OSF CI/CD

Datamance:
Done.

PR #452, opened 2020-03-03
Feature/451 cloudbuild files

Datamance:


Datamance:
#451 - cloudbuild pushes to GCR

PR #460, opened 2020-03-24
fix 'check for updates' on study edit view only to remove current sha…

kimberscott:
… from list of updates, not whatever the least-recent update is, so that we don't remove an extra update if the current sha isn't on the master branch

### DIFF
--- a/studies/templates/studies/_study_type.html
+++ b/studies/templates/studies/_study_type.html
@@ -43,9 +43,6 @@
                     url: commit_options_api_url,
                     dataType: 'json'
                 }).done(function(data) {
-                    console.log(data);
-                    console.log(commit_options_api_url);
-                    
                     if (data.length > 0) {
                         // Commits returned are inclusive of the reference commit, 
                         // which will be last (most recent first) if it's in master - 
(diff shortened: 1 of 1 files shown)


studies/templates/studies/_study_type.html
```diff
@@ -31,30 +31,34 @@
...
             var commit_options_api_url = current_repo.replace('https://github.com/', 'https://api.github.com/repos/') + `/commits?since=${current_commit_date}`;
             
             $.ajax({
                     url: commit_options_api_url,
                     dataType: 'json'
                 }).done(function(data) {
-                    // Note commits returned are inclusive of the reference commit, which will be last (most recent first) - don't list or count this one
-                    var commits_since = $.map(data.slice(0,-1), function(c) {return `<tr><td>${c.commit.author.date.split('T')[0]}</td><td>${c.commit.message}</td><td>${c.sha}</td></tr>`}).join('');
-                    if (data.length > 1) {
+                    console.log(data);
```
Datamance: Lets remove these console logs

kimberscott: whoops, thanks!

PR #478, opened 2020-04-03
explicitly choose appropriate on_delete behavior for accounts, studies models

kimberscott:
I skipped the Organization-related ones since we plan to change/remove that model anyway.

Thoughts welcome on how this *should* work. My main motivations are:

- I'd like to be able to manually (via Admin) delete old test studies and responses, especially on staging, to clean things out periodically. Right now that involves manually going through study logs, responses, Videos, consent rulings, etc. and making sure I've gotten everything (in a particular order!), plus going to S3 if I'm feeling ambitious.
- I'd like to make sure we're in a good position to (still manually) delete data if requested by a participant. (We can refine this behavior later in the automated-GDPR-tools ticket.)

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -123,7 +123,7 @@ class User(AbstractBaseUser, PermissionsMixin, GuardianUserMixin):
     nickname = models.CharField(max_length=255, blank=True)
     organization = models.ForeignKey(
         Organization,
-        on_delete=models.CASCADE,
+        on_delete=models.PROTECT,
         related_name="users",
         related_query_name="user",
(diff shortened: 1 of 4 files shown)


Datamance:
As long as we pretty much only delete things through admin, this sounds reasonably safe to me.

kimberscott:
Yeah, the only other times we delete things are:
* when a response is withdrawn the associated non-consent video gets deleted
* there's a way to delete all preview responses from the all responses view

It's not possible to delete anything through the API, as Dawn helpfully documented very clearly :)

PR #490, opened 2020-04-17
Develop

kimberscott:


Datamance:
We don't need to request review for merge to master - I'm noodling on a new workflow for all of this, so that we can document it for new contributors

kimberscott:
Sorry about that - it seems like there's a standard question for dev->master PRs which isn't about code review but more "hey is this an ok time to merge / are these changes sufficiently tested on staging / want to go ahead when it's convenient." Maybe that can be included in the new workflow?

Datamance:
I see what you're saying - in the master branch, using the review mechanism as a sort of "hey, please make sure we're cool with these commits going into prod" rather than another review.

PR #493, opened 2020-04-29
WIP: Upgrade django

kimberscott:


studies/models.py
```diff
@@ -127,7 +131,7 @@ class Study(models.Model):
     shared_preview = models.BooleanField(default=False)
     creator = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
 
-    metadata = DateTimeAwareJSONField(default={})
+    metadata = DateTimeAwareJSONField(default=dict)
```
Datamance, line 134: Oh wow surprised that wasn't a bug earlier - never use mutable objects as defaults on keyword args, otherwise the reference stays the same because the dict literal is evaluated at function parse time. That's why they make you use factory funcs

studies/migrations/0064_auto_20200429_1433.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 2.1 on 2020-04-29 18:33
```
Datamance, line 1: What is this migration generated from? Was it manually invoked?

If so, could you migrate back one and re-run the makemigration command with `--name=whatever` so we know what it's about?

studies/migrations/0063_auto_20200428_1253.py
```diff
@@ -0,0 +1,90 @@
+# Generated by Django 2.0 on 2020-04-28 16:53
```
Datamance, line 1: See comment below about naming your migrations. Also, try to "squash" your migrations - if you find more changes are being made,`python manage.py migrate ${number of previous}`, delete the old migration file, and then re-run `python manage.py makemigrations --name=${whatever the name is}` and then finally run `python manage.py migrate` again

project/urls.py
```diff
@@ -31,7 +31,7 @@
 urlpatterns = [
     url(r"^favicon\.ico$", favicon_view),
     url(r"^admin/", admin.site.urls),
-    url(r"^api/", include(api_urls)),
+    url(r"^api/", include((api_urls, 'api'))),
```
Datamance, line 34: s/'/"/g

You should configure your editor to run `black` every time you save. That way you don't have to remember :)

Pipfile
```diff
@@ -26,10 +26,10 @@ bcrypt = "==3.0.0"
 boto3 = "==1.4.5"
 celery = "==4.3.0"
 ciso8601 = "==1.0.3"
-Django = "==1.11.28"
+Django = "==2.2"
 django-ace-overlay = "==0.8.0"
 django-allauth = "==0.32.0"
-django-bitfield = "==1.9.3"
+django-bitfield = "==2.0.1"
```
Datamance, line 32: make sure you change this django-bitfield requirement in `requirements/default.txt` as well

accounts/migrations/0046_auto_20200428_1253.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 2.0 on 2020-04-28 16:53
```
Datamance, line 1: See below comments about naming and squashing migrations

project/urls.py
Datamance, line 34: Are you not able to define `app_name="api"` in the URLConf like you were for `web` and `exp`?

studies/models.py
kimberscott: Yep, that's a surprise. At least we'd caught the other instances of this.

project/urls.py
kimberscott: I just set up pre-commit to do this, isort, & check for merge conflicts. If you don't want that as part of a forced workflow I can keep those running only locally :)

Pipfile
kimberscott: There were some others that weren't aligned, I think now everything's lined up between Pipfile & requirements

project/urls.py
kimberscott: correct, this had to do with the way the urlpatterns were generated differently in the api vs. web/exp.

studies/migrations/0064_auto_20200429_1433.py
kimberscott: Done, sort of. This one was generated when switching to Django 2.0 but there are no accompanying model changes, so I just named it based on that upgrade. Similarly for the studies migrations (there were originally one set each for 1.11 -> 2.0, 2.0 -> 2.1). If these names should be more descriptive about the relevant changes that are being imposed I can change them, the Django version seemed more practically relevant for us knowing when these need to be applied.

project/settings/defaults.py
```diff
@@ -30,7 +30,7 @@
 )
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = bool(os.environ.get("DEBUG", False))
+DEBUG = True  # bool(os.environ.get("DEBUG", False))
```
Datamance, line 33: Usually you should just be able to edit this in `settings/local.py` and it'll override the default.

We should keep this as `bool(os.environ.get("DEBUG", False))` so that any env vars we pass are respected. Also, we should generally not have DEBUG on when deploying :)

studies/0063_auto_20200428_1253.py
```diff
@@ -0,0 +1,146 @@
+# Generated by Django 2.0 on 2020-04-28 16:53
```
Datamance, line 1: OK, looking at these, I don't think any of the migrations have to do with the upgrade to django itself - these are probably just field alterations that are happening now because we didn't migrate the last time we edited model fields (I think that might have been your last PR, which was modifying `on_delete` behavior).

As such, when I merge, I'm probably going to leave out these migrations (unless you can think of a good reason not to)

project/settings/defaults.py
kimberscott: ACK I'm sorry about that, that was not supposed to be committed. It is a hack for my local setup where I couldn't get it respecting the darn environment variable for a while.

studies/0063_auto_20200428_1253.py
kimberscott: Oh, that makes sense. (And I guess I don't understand migrations well enough, because now I'm surprised deploying those changes actually modified the behavior of deletion.)

If they ought to have happened along with the previous PR, though, why wait on applying them?

Some of the operations are actually related to the Django upgrade - changing default values so they're not {}, and something that got created about changing a manager:

```
  operations = [
        migrations.AlterModelOptions(
            name="response",
            options={
                "base_manager_name": "related_manager",
                "ordering": ["-demographic_snapshot__created_at"],
                "permissions": (
                    (
                        "view_all_response_data_in_analytics",
                        "View all response data in analytics",
                    ),
                ),
            },
        ),
        migrations.AlterField(
            model_name="study",
            name="metadata",
            field=project.fields.datetime_aware_jsonfield.DateTimeAwareJSONField(
                default=dict
            ),
        ),
        migrations.AlterField(
            model_name="study",
            name="structure",
            field=project.fields.datetime_aware_jsonfield.DateTimeAwareJSONField(
                default=studies.models.default_study_structure
            ),
        ),
        migrations.AlterField(
            model_name="studytype",
            name="configuration",
            field=project.fields.datetime_aware_jsonfield.DateTimeAwareJSONField(
                default=studies.models.default_configuration
            ),
        ),
    ]
```

kimberscott:
General summary of what had to be done:
- Update dependencies that aren't compatible with Django 3.0 - mostly due to use of python 2 shims (django-bitfield, django-debug-toolbar , django-allauth, django-bootstrap3, django-cors-headers, django-countries, django-dynamic-fixture, django-guardian, django-localflavor, django-multiselectfield, django-revproxy, djangorestframework, djangorestframework-jsonapi, google-cloud-storage, transitions
- Clarify namespace in defining URLs - e.g. `view_name="organization-detail"` -> `view_name="api:organization-detail"` in accounts, studies, & api serializers (and tests); add app_name for web & exp urls
- `MIDDLEWARE_CLASSES` -> `MIDDLEWARE`
- provide constructors for default model values that are objects, rather than object instances (e.g. {})
- minor change as shown in `exp_extras.py` due to maybe change in / more stringent testing
- switch to transitions.Machine instead of GraphMachine
- set up tests to actually provide Organization & StudyType for studies, birthdays for Childs where needed, no study field on ConsentRuling, demographic_snapshot for responses as needed
- remove `context` param from `from_db_value`
- replace `render_to_response` with `render`
- skip the test `testAdminsCannotAutomaticallyViewEmails` because it incorrectly relies on a non-existent field `is_org_admin`
- make image on web:study-detail view optional OR mock up an image field for Study for testing

PR #500, opened 2020-05-11
fix bug about NoneType child birthdates in demographic form

### LINKED ISSUE #498: Demographic snapshot won't download
**Description**
I've tried to download the demographic data for Flurps and Zazzes in both JSON and CSV format - each one gives back a Server Error (500) message.

**How to reproduce**
On the Demographic Snapshots page (https://lookit.mit.edu/exp/studies/10/responses/demographics/), click the buttons (JSON or CSV) to download the data.

**Expected behavior**
A file should download automatically.

Mac OS 10.14.6 (Mojave), on Chrome

kimberscott:
Fixes #498 

Previously all elements of the child birthdate list were assumed to be dates, so anything else (specifically None) broke processing into csv/json files. This was likely due to migration of accounts from JamDB system a few years ago (see import_jamdb_participants.py), since format of new data is enforced.

Also upon consideration of the `child_rounded_age` field in the demographic data, make the rounded ages be at the time of the response rather than the time at which the demographic data was saved. (This makes it easier to match to participants and count older/younger siblings, ask if they have an x-year-old sibling, etc.; harder to interpret demographic variables in light of age of children at time of filling out survey - figuring former will be more common.)

### DIFF
--- a/exp/utils.py
+++ b/exp/utils.py
@@ -1,5 +1,6 @@
 import csv
 import io
+import datetime
 
 RESPONSE_PAGE_SIZE = 500  # for pagination of responses when processing for download
 
@@ -60,7 +61,7 @@ def round_age(age_in_days):
 def round_ages_from_birthdays(child_birthdays, date_created):
     return [
         round_age((date_created.date() - birthdate).days)
-        if birthdate and type(birthdate) == type(date_created.date())
+        if birthdate and type(birthdate) is datetime.date
         else None
         for birthdate in child_birthdays
     ]
--- a/exp/utils.py
+++ b/exp/utils.py
@@ -60,5 +60,7 @@ def round_age(age_in_days):
 def round_ages_from_birthdays(child_birthdays, date_created):
     return [
         round_age((date_created.date() - birthdate).days)
+        if birthdate and type(birthdate) == type(date_created.date())
+        else None
         for birthdate in child_birthdays
     ]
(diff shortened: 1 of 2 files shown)


exp/utils.py
```diff
@@ -60,5 +60,7 @@ def round_age(age_in_days):
 def round_ages_from_birthdays(child_birthdays, date_created):
     return [
         round_age((date_created.date() - birthdate).days)
+        if birthdate and type(birthdate) == type(date_created.date())
```
Datamance, line 63: `type(birthdate) is datetime.date`

exp/views/study.py
```diff
@@ -1955,6 +1955,7 @@ def get_response_values_for_demographics(self, study):
             .select_related("child", "child__user", "study", "demographic_snapshot")
```
Datamance, line 1955: While we're here, can we fix the misspellings and mixed casing in the ~40 lines above? stuff like `allHeaders` -> `all_headers`, `StudyParticiapnts` -> `StudyParticipants`

exp/views/study.py
```diff
@@ -2097,12 +2098,11 @@ def get_csv_demographic_row_and_headers(self, resp=None):
             (
                 "demographic_child_rounded_ages",
                 round_ages_from_birthdays(
```
Datamance, line 2100: So, instead of using `DemographicSnapshot.created_at` we're using `Response.date_created`? Can you clarify how this solves the problem, seeing as both are `models.DateTimeField(auto_now_add=True)` (in other words, auto-populated)?

exp/utils.py
kimberscott: yep, was foolishly trying to avoid importing datetime just for that :)

exp/views/study.py
kimberscott: done, thanks for noticing those

PR #505, opened 2020-06-01
WIP: Feature/140 lab permissions

Datamance:
Split into 3 major commits. The first, https://github.com/lookit/lookit-api/commit/ec3bdb7903c33ad93cca1aeedc888ad721456ab1, is a brand new file that would be the skeleton for the rest of everything else. If you want to start from a different branch (say - the django 3.0 upgrade), or pursue a different approach, it would make sense to cherry pick that commit to start (guaranteed no merge conflicts), and then cherry-pick and resolve merge conflicts for the actual model changes and relevant migrations (https://github.com/lookit/lookit-api/commit/5828935e05e2102c1478ee1b5b47e8f367d243be). 

Then, depending on if you want to use the bulk of my work or try things differently for the redesigning and refactoring stage, you could maybe pick and choose files from (https://github.com/lookit/lookit-api/commit/bfdd72625abdf01b416be8905f78c6121127afd9).


studies/models.py
```diff
@@ -133,6 +181,35 @@ class Study(models.Model):
...
+    )
+    researcher_group = models.OneToOneField(
+        Group, related_name="study_for_research", on_delete=models.SET_NULL, null=True
+    )
+    manager_group = models.OneToOneField(
+        Group, related_name="study_to_manage", on_delete=models.SET_NULL, null=True
+    )
+    admin_group = models.OneToOneField(
+        Group, related_name="study_to_administer", on_delete=models.SET_NULL, null=True
+    )
```
kimberscott, line 211: Is there a reason that some of the `related_name`s are `study_to_...` and some are `study_for_...`?

exp/views/study.py
```diff
@@ -2782,7 +2888,12 @@ def get_permitted_triggers(view_instance, triggers):
             # 3) trigger is found in user's study permissions
         if not user.is_superuser:
             if trigger in admin_triggers:
-                if not (user.organization == study.organization and user.is_org_admin):
+                if not (
+                    user.lab == study.lab
+                    and user.has_perm(
+                        LabPermission.CHANGE_STUDY_STATUS.codename, user.lab
+                    )
+                ):
```
kimberscott, line 2896: I'm a little confused about what this section is doing and why the admin_triggers (approve, reject) are being broken out separately for non-superusers. This seems to say that you can't cause a study to be approved/rejected unless you have the *lab* level permission specifically, rather than either lab or study level (as for e.g. can_edit_study). Also why include 'approve' which is never granted to non-superusers?

exp/views/study.py
```diff
@@ -445,43 +468,59 @@ def manage_researcher_permissions(self):
...
+                    user_to_update.groups.add(study_admin_group)
+                    user_to_update.groups.remove(study_researcher_group)
+                    self.send_study_email(user_to_update, "admin")
+                if name_of_role_to_enable == "study_read":
+                    # if read, removes user from study admin and adds to study read
+                    # Must have more than one admin to make this change.
+                    if study_admin_group.user_set.count() > 1:
+                        user_to_update.groups.add(study_researcher_group)
+                        user_to_update.groups.remove(study_admin_group)
+                        self.send_study_email(user_to_update, "read")
```
kimberscott, line 504: does this need to change now to encompass all the new study roles?

exp/views/study.py
Datamance: I think for this particular section I had just changed some of the organization stuff and replaced it with lab stuff, just to get the view serving.

If something looks really out of place, it's probably that sort of thing.

exp/views/study.py
Datamance: Yep - we've got new roles and new permissions. Wherever we're using old roles and perms, we need to change.

Datamance:
Merged into #513

PR #523, opened 2020-06-25
Bugfix/cleanup permissions changes

kimberscott:
Small fixes - will add tests! But can be merged in if looks ok.

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -190,6 +190,16 @@ class User(AbstractBaseUser, PermissionsMixin, GuardianUserMixin):
             ]
         )
 
+    def labs_user_can_create_study_in(self):
+        lab_ids = [
+            lab.id
+            for lab in self.labs.only("id")
+            if self.has_perm(
+                LabPermission.CREATE_LAB_ASSOCIATED_STUDY.codename, obj=lab
+            )
+        ]
+        return self.labs.filter(id__in=lab_ids)
+
     def has_any_perms(self, perm_list, obj=None):
         """
         Returns True if the user has ANY of the specified permissions. If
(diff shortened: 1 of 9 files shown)


accounts/models.py
```diff
@@ -186,7 +186,7 @@ def can_create_study(self):
                 self.has_perm(
                     LabPermission.CREATE_LAB_ASSOCIATED_STUDY.codename, obj=lab
                 )
-                for lab in self.labs.values("id")
+                for lab in self.labs.all()
```
Datamance, line 189: ugh, my bad. Change to `only("id")` - keeps the query fast; we only ever need ID for permission checks

accounts/models.py
```diff
@@ -186,10 +186,20 @@ def can_create_study(self):
                 self.has_perm(
                     LabPermission.CREATE_LAB_ASSOCIATED_STUDY.codename, obj=lab
                 )
-                for lab in self.labs.values("id")
+                for lab in self.labs.only("id")
             ]
         )
 
+    def labs_user_can_create_study_in(self):
+        lab_ids = [
```
Datamance, line 194: `get_objects_for_user` will keep it all in one query rather than evaluating each lab with `self.has_perm`

You probably want something like this instead:

```python
self.labs.filter(id__in=get_objects_for_user(self, LabPermission.CREATE_LAB_ASSOCIATED_STUDY.prefixed_codename).only("id"))
```

As an interesting sidenote... doing your first query (the list comp) in the PyCharm console actually breaks with a `SynchronousOnlyOperation` error. I'm going to pull down your code in a second to test it and see if this is just a weird issue with the PC console, or if there's something else going on.

Datamance: ok, looks like it's just a weird thing with the PyCharm console (the REPL debugger probably uses an async loop, which Django doesn't like). Regardless, I just tested it and confirmed that this will prevent duplicate queries (you can see this with just 2 labs but it won't start having an impact on performance until you have more than 2 labs)

kimberscott: yes, I saw that too in the PyCharm console and was puzzled! Thanks, and sorry, that should have been an obvious place to use get_objects_for_user.

PR #526, opened 2020-07-17
Feature/511 multi factor authentication

Datamance:
OK!

I would request a thorough review of this code (including local testing), as well as some suggestions for new tests. 

There are two things that this PR is missing, both having to do with cleanup/deprecation:
1. A migration file to drop the `socialaccount_*` and allauth-based `accounts_*` tables.
2. Deletion of the entire `osf_oauth2_adapter` app.

I will feel more comfortable adding these things once we have *really* kicked the tires on this code.


kimberscott:
Issues & questions from local testing:

* Login as researcher: Researcher login links from homepage and from participant login point to /exp/studies and returns a 403 error because AnonymousUser is not a researcher account if you try to access it to log in. Similarly, when I log out as a researcher I'm directed to /exp/studies and get a 403 error - can redirect to homepage or to researcher login.
* Is it actually all one login view for participants & researchers now? If so we should just rename the "Participant Login" header. 
* OTP setup/management as researcher: 
   * needs a link to the OTP setup page; upon logging in I see that "In order to access experimenter pages, you must have two-factor authentication enabled, and have logged in with your OTP key." but can't see how I would do that. 
   * Once the researcher has set up OTP, is there a way for them to disable it and re-enable, or only by contacting us? 
   * Can there be some indicator (e.g. in /account/) that OTP is enabled for this account, and that any changes require contacting us? (The warning about not being able to come back later is helpful, but no one will remember exactly what it said or who they were supposed to email later.) 
   * Why is is a security risk to display the QR code generator again *if the account hasn't had 2FA set up yet*? This wouldn't let you regain access to Experimenter if you didn't have the OTP, it'd just let you not do the setup immediately if you didn't want to. That way if you don't have your phone on you or whatever, you can come back later.
* Password management: The change-password form that used to be in the participant account form (/account/) has disappeared.
* Registration and user fields as researcher: 
   * How would a researcher know where to go to register? (I see it's /registration/ from urls.py, just no links to it.) In particular we may need to *stop* researchers from trying to sign up as participants. We could put a link in the footer where researcher login currently is, plus a link from participant signup (e.g. "This form is for families interested in participating with their children. Looking for [researcher registration instead]?")
   * "First name" -> "Given name", "Last name" -> "Family name" (just for consistency with how these have been labeled). We could actually auto-populate the "nickname" field with the researcher's given name, and they can change it later if they want; it'll make the form less confusing. (Right now researchers generally have blank nicknames, but that's admittedly confusing when collecting preview data!)
   * A researcher needs to have read and accepted both Terms of Use and Privacy Policy, not just Privacy Policy.
   * Still need to remove all usages of timezone (e.g. I get pytz.exceptions.UnknownTimeZoneError: '' on accessing a study detail page with a newly-created researcher account)
   * Do we need to remove usage of identicon, or will that be ok when the field is blank? (Or does it get magically populated?)
* Login as participant: Participant login asks for username, password, AND OTP at once. Will all participants have to set up 2FA? I thought it would be optional for them, only required for researchers. If it's optional, can it be at a second step, only for accounts where it's required? (I see how asking for all three simultaneously is more secure, but I don't think it's worth the amount of confusion it will cause participants.)
* OTP setup/management as participant: 
    * How would a participant turn on (and set up) or turn off OTP? 
    * If this isn't an option yet: it *really* shouldn't be part of the participant login form :)
    * If it is an option, we'll likely want slightly more thorough setup instructions - happy to draft these if so
* Formatting etc. 
   * Can we have a link to /account/ under the "Researcher Name" menu at top right of Experimenter? (Right now it's  just a log out link)
   * Spell out OTP -> "One-time password"
   * This is dumb but can the QR code on /2fa/ be smaller (e.g. just give a max size in pixels)?

accounts/forms.py
```diff
@@ -14,6 +20,122 @@ class Meta:
...
+        super().__init__(*args, **kwargs)
+        # Don't autofill passwords, in the interest of security.
+        self.fields["password1"].widget.attrs["autocomplete"] = "new-password"
+        self.fields["password2"].widget.attrs["autocomplete"] = "new-password"
+
+    def save(self, commit=True):
+        """Just flip the active and researcher flags."""
+        user = super().save(commit=False)
+        user.is_active = True
+        user.is_researcher = True
```
kimberscott, line 124: should probably move the post-save signal `add_researcher_to_labs` from studies/models.py here

Datamance:
a few questions answered in no particular order:

- identicon is automagically created
- The change password form only presents for participant accounts (`{% if not request.user.is_researcher %}`)
- Once a researcher sets OTP up, the idea was that we don't let them deactivate it. If we did, then all an attacker would need is a username and password to log in, after which they could deactivate (really, erase) the old OTP and promptly set up a new one. One thing I just thought of, though, is providing a deactivation view that takes all your credentials (username, password, OTP) as input - that way, destroying the old OTP secret can only be done if you have the current one. But that seems a little silly - I can't really see a situation where a researcher with ready access to their OTP code would want to get rid of the one that already works

Datamance:
More questions answered:
- The login view, as it stands, is currently just a single view. Since the user hasn't logged in yet, we can't tell what kind of account it'll be, so we have to present all of the credentials fields. Unless you want to make (and link to) two distinct login pages, there's no real good way to infer which fields to present to the user. I'm all ears if you have an idea on how to approach this!
- The way we have it set up right now, visiting the `/2fa/` page automatically triggers the creation of an OTP secret associated with the user in the backend - without that, no QR code can be generated/scanned by the user. Once that OTP secret is created, we're in the situation I described before where allowing normally authenticated users to directly reset OTP secrets is as good as having no 2FA at all.
- Where exactly do you want "One-time password" spelled out - which template, that is?

kimberscott:
> The change password form only presents for participant accounts ({% if not request.user.is_researcher %})

How do researchers change their passwords though? Don't we want them to do that periodically?

> Once a researcher sets OTP up, the idea was that we don't let them deactivate it. If we did, then all an attacker would need is a username and password to log in, after which they could deactivate (really, erase) the old OTP and promptly set up a new one. One thing I just thought of, though, is providing a deactivation view that takes all your credentials (username, password, OTP) as input - that way, destroying the old OTP secret can only be done if you have the current one. But that seems a little silly - I can't really see a situation where a researcher with ready access to their OTP code would want to get rid of the one that already works

If they want to switch phones? Possibly rare enough on the scale of things that we don't need to worry about it, though.

Datamance:
> How do researchers change their passwords though? Don't we want them to do that periodically?

Ah good point! We're using the same login system now so we should allow researchers to change their passwords the same way that participants do.

kimberscott:
> The login view, as it stands, is currently just a single view. Since the user hasn't logged in yet, we can't tell what kind of account it'll be, so we have to present all of the credentials fields. Unless you want to make (and link to) two distinct login pages, there's no real good way to infer which fields to present to the user. I'm all ears if you have an idea on how to approach this!

So for instance when you log in on OSF, you first enter your username/password, and then if those match, based on whether MFA is enabled you either have to enter the OTP or you're fully logged in. The idea would be a two-step process essentially. (This does make it easier to brute-force a password, which is what I meant about it being less secure.) I'm not sure if this is easier or harder than two login pages though.

> The way we have it set up right now, visiting the /2fa/ page automatically triggers the creation of an OTP secret associated with the user in the backend - without that, no QR code can be generated/scanned by the user. Once that OTP secret is created, we're in the situation I described before where allowing normally authenticated users to directly reset OTP secrets is as good as having no 2FA at all.

Couldn't we store whether OTP has ever successfully been set up (i.e. the test OTP works and the user has their first 2fa-validated login) and allow seeing that page until then?

Datamance:
> Couldn't we store whether OTP has ever successfully been set up (i.e. the test OTP works and the user has their first 2fa-validated login) and allow seeing that page until then?

Sure, we can do that.

> So for instance when you log in on OSF, you first enter your username/password, and then if those match, based on whether MFA is enabled you either have to enter the OTP or you're fully logged in. The idea would be a two-step process essentially. (This does make it easier to brute-force a password, which is what I meant about it being less secure.) I'm not sure if this is easier or harder than two login pages though.

Yeah, I'm a little hesitant to do a two-step login for that reason. Two different login pages also doesn't make too much sense because theoretically, participants should have the _option_ to secure with 2FA as well.

What if we just put help text underneath that says something like "`You only need to fill this out if you've set up two-factor authentication (required for experimenters)`"?

We could even hide/display the field with a checkbox to signal to users, "this is a special field that probably doesn't pertain to you!"

kimberscott:
> Yeah, I'm a little hesitant to do a two-step login for that reason. Two different login pages also doesn't make too much sense because theoretically, participants should have the option to secure with 2FA as well.
> 
> What if we just put help text underneath that says something like "You only need to fill this out if you've set up two-factor authentication (required for experimenters)"?
> 
> We could even hide/display the field with a checkbox to signal to users, "this is a special field that probably doesn't pertain to you!"

Despite the confirmation that would be provided in a brute-force attack (side note - do we have any rate limiting? is that easy to add in django? because that would likely do substantially more good here), I think the tradeoff here in usability vs. security still favors a multi-step process. In addition to avoiding confusion for families trying to log on (although I really wouldn't underestimate that): 

* Multiple steps seems to be the clear standard for 2fa, to the point where it's even included in many descriptions of how 2fa *works*. I think we're going to be fighting against an expected pattern here if we do it differently.

* If the user is slightly forgetful, and isn't sure which of two passwords they've used, it's going to be annoying to enter both the password and the 2fa code every attempt. 

* A multi-step process is more compatible with potential future options for mfa which rely on the username and/or timing of the request (e.g., piece of knowledge, push notification). 

* There are also arguments to be made that phishing is slightly more complicated if you have to fake multiple pages for login; that autofill on browsers may mess with password + OTP fields; and that a two-step process may make it easier to detect (more suspicious!) cases of correct password + incorrect OTP code; but I'm not sure I'd be convinced about those.

Example discussion which includes some of these points - https://security.stackexchange.com/questions/55318/why-not-ask-for-a-password-and-2-factor-token-at-the-same-time

The advantage of two login pages, if you're very concerned about the brute-force-password-then-OTP-sequentially approach, is that you could do:
- participant login with two-step process. If the account is a participant account without 2fa, they're logged in right away. Participant with 2fa, they're prompted for the 2fa code. Incorrect credentials or correct credentials but researcher account - they're told that these are not valid credentials for a participant account. 
- experimenter login with one-step process. Now there's no way to sequentially brute-force passwords & OTP for researcher accounts, because you wouldn't get any feedback on whether the username/password combo was correct from the participant account.

kimberscott:
> Where exactly do you want "One-time password" spelled out - which template, that is?

It looks like all the uses are in accounts/forms.py - added proposals to my PR.

kimberscott:
One more thing - we'll need to eliminate the references to OSF profiles / login / etc. throughout. Searching for 'osf' does a pretty good job identifying cases to look at, e.g.: `osf_profile_url` on the User model and use in `exp/templates/exp`; use of `osf_url` in Message.send_as_email and helpers.send_mail (and any associated use in templates); the `local_user_exists.html` template; the `resend_confirmation` and `reset_password` email templates; references in `LabMembersView`

kimberscott:
Note: the password reset form errors upon invalid form input (e.g. mismatched or too short new password) but that's fixed by https://github.com/lookit/lookit-api/pull/529/commits/489267379f8c3c06e6a6ab89fbe38354b66ba69d

2FA shouldn't be "enabled" immediately upon clicking the "set up 2FA" link, only upon confirming a valid test code. Otherwise, the moment someone clicks on that either accidentally or out of curiosity (expecting to get more information about what this is and how to set it up), they have to set up 2FA now or never. (Fortunately they're not actually locked out of the account, but even that's not obvious ahead of time.) I feel like I must be missing something here, I know we've discussed the "you can't leave!!" messaging - I understand the danger in allowing someone to reset OTP *after* setting it up the first time, but I'm not seeing the problem if it hasn't been actually enabled yet anyway.

Could we put the 2FA indicator (Enable / enabled / logged in with 2FA) only inside the account form (/account/), where the password change form is? Otherwise it's really salient *all the time* for both participants and experimenters, for something that they should only interact with very rarely.

Also can we allow *participant* accounts to disable and enable 2FA? I think that's actually important; we'll have people who try it and then decide it's more hassle than they need, and we'll often have multiple parents sharing an account (but not a phone). It'll be slightly less secure for the reasons we've discussed, but it's very standard to be able to do this - I think it's a reasonable tradeoff to allow it for participants and not for researchers. (And people can't change their own researcher status in any event.)

Datamance:
> 2FA shouldn't be "enabled" immediately upon clicking the "set up 2FA" link, only upon confirming a valid test code.

Ah, my bad - I had a problem with the template; wasn't checking the new `activated` flag :[ Should be fixed now.

> Could we put the 2FA indicator (Enable / enabled / logged in with 2FA) only inside the account form (/account/), where the password change form is?

> Also can we allow _participant_ accounts to disable and enable 2FA?

Sure, I've now put everything in the Account Update (formerly ParticipantUpdate) view.

kimberscott:
I spent about half an hour reviewing again, but it doesn't look like this is ready. Can you make sure you've looked through comments above and tested locally? (going through each expected workflow - e.g. creating an account as a participant, enabling/disabling 2FA, creating an account as a researcher... - checking that the content you see makes sense, and you're able/unable to do the expected things) Then I'm happy to look through more thoroughly. Here are notes from as far as I got - 

- still a researcher login link from home page, & goes to exp/studies so there's a 403
- link to account management under account name from exp interface?
- **if you log in on a 2FA-enabled account, before entering the OTP, you can go to studies & see past videos, "account management" & see/edit children, demographics, email, etc.**
- "forgot password" link on OTP entry may be confusing (they clearly haven't, they've already entered the password, so they may think this is a way to reset OTP secret) - may want to instead note how to request reset of 2FA if needed here
- currently allowing disabling 2FA for experimenters - I actually think this makes sense (they can only do this when logged in with 2FA, and would have to re-set up to see exp views again; this allows them to e.g. use a new phone) but double checking you're ok with it since it seems counter to previous decisions

minor review suggestions that don't wouldn't fall under testing :)
- "account management" -> "my account"?

kimberscott:
In terms of behavior this is looking terrific - very smooth and seamless. (And I love how guided the "hey you're going to need to set up 2FA" is for researchers!)

Small requests, last round:
- would it make sense to redirect from /exp/ views to login if someone isn't logged in at all? (for researchers who emailed a link to something, bookmarked a link, etc.) If that's a headache with all the redirect issues with tests, nevermind.
- could we require both family name & given name on the researcher signup?
- after staring at it some more I feel like maybe the balance of "make sure researchers don't accidentally sign up as participants" vs. "make sure participants don't accidentally sign up as researchers / get confused about what's going on" could shift towards the latter since there will (hopefully!) be many more participants. I made a proposal in my review branch, does that look reasonable?

I still owe you actual code review in the sense of reading the changes to the code!

accounts/admin.py
```diff
@@ -55,7 +61,16 @@ class MessageAdmin(GuardedModelAdmin):
     search_fields = ["subject", "body"]
 
 
+class TOTPAdmin(GuardedModelAdmin):
+    list_display = ("user",)
+    search_fields = ("user__username", "user__nickname")
```
kimberscott, line 66: would use username & family_name here as existing researchers usually have blank nickname fields

accounts/backends.py
```diff
@@ -0,0 +1,46 @@
...
+
+from accounts.models import User
+
+TWO_FACTOR_AUTH_SESSION_KEY = "using_2FA"
+
+
+class TwoFactorAuthenticationBackend(ModelBackend):
+    """Grabs a regular user, but checks OTP as well as password.
+
+    DEPRECATED
```
kimberscott, line 14: why is this marked deprecated?

accounts/forms.py
kimberscott: just bumping this up - does it make sense to do this now or let it be?

accounts/models.py
```diff
@@ -73,6 +79,47 @@ class Meta:
...
 
 
+class GoogleAuthenticatorTOTP(models.Model):
+    """OTP Secret model for time-based OTP.
+
+    Currently, we only support Google authenticator format. Please see:
+    https://github.com/google/google-authenticator/wiki/Key-Uri-Format
+    """
+
+    issuer = "Lookit"
```
kimberscott, line 89: Is this the label that shows up in the Authenticator app? Do we want it to say e.g. Lookit vs. Lookit-staging vs. Lookit-dev?

accounts/models.py
```diff
@@ -536,7 +580,6 @@ class Message(models.Model):
     def send_as_email(self):
         context = {
             "base_url": settings.BASE_URL,
-            "osf_url": settings.OSF_URL,
             "custom_message": mark_safe(self.body),
```
kimberscott, line 583: (I'm not saying much because everything is looking really nice and clearly written, but I am at least double-checking e.g. to make sure there aren't any uses of `osf_url` as I go along)

accounts/scripts/get_jamdb_users.py
```diff
@@ -1,47 +0,0 @@
-import json
```
kimberscott, line 1: hooray!

accounts/signals.py
```diff
@@ -1,40 +1,7 @@
...
-#    for app_config in apps.get_app_configs():
-#        create_permissions(app_config, apps=apps, verbosity=0)
-#
-#    Organization = sender.get_model("Organization")
-#    org, created = Organization.objects.get_or_create(
-#        name="MIT", url="https://lookit.mit.edu"
-#    )
-
-
-def post_migrate_create_social_app(sender, **kwargs):
```
kimberscott, line 16: Will this mess with the ability to set up the db & run through migrations from scratch, or are we ok there? Will we be able to do backwards migrations / does this code belong somewhere in a specific migration?

accounts/templates/emails/resend_confirmation.txt
```diff
@@ -1,8 +0,0 @@
-Dear {{ researcher_name }},
```
kimberscott, line 1: so many little bits of nonsense gone!

accounts/tests.py
```diff
@@ -1,15 +1,229 @@
...
+
+        self.assertEqual(
+            response.redirect_chain, [(reverse("accounts:2fa-setup"), 302)]
+        )
+        self.assertEqual(response.status_code, 200)
+
+        # Mock correctly entered OTP
+        otp = response.context["user"].otp
+        response = self.client.post(
+            reverse("accounts:2fa-setup"), {"otp_code": otp.provider.now()}, follow=True
```
kimberscott, line 82: I know I'm likely impressed by the wrong things but it's very cool that we're able to test this out.

accounts/tests.py
```diff
@@ -1,15 +1,229 @@
...
+        self.assertNotIn(TWO_FACTOR_AUTH_SESSION_KEY, response.wsgi_request.session)
+        self.assertEqual(old_qr_svg, new_qr_svg)
+
+    def test_researcher_login_redirect_to_2FA_verification(self):
+        response = self.client.post(
+            reverse("login"),
+            {"username": self.researcher_email, "password": self.test_password},
+            follow=True,
+        )
+        self.assertTrue(response.status_code == 200)
```
kimberscott, line 123: is there a reason not to use assertEqual?

accounts/tests.py
```diff
@@ -1,15 +1,229 @@
...
+        response = self.client.post(
+            two_factor_auth_url, {"otp_code": str(int(self.otp.provider.now()) + 1)},
+        )
+
+        # We just reloaded the page, so we should get a 200
+        self.assertEqual(response.status_code, 200)
+        # There are form errors...
+        self.assertNotEqual(len(response.context["form"].errors), 0)
+        # And the session key isn't set.
+        self.assertIsNot(
```
kimberscott, line 180: assertFalse?

accounts/views.py
```diff
@@ -1,3 +1,291 @@
...
+
+    def get_success_url(self) -> str:
+        user: User = self.request.user
+        otp: GoogleAuthenticatorTOTP = getattr(user, "otp")
+        if otp and otp.activated:
+            return reverse("accounts:2fa-login")
+        elif user.is_researcher or user.is_staff:
+            messages.warning(
+                self.request,
+                "If you're a researcher or Lookit staff, you'll want to set up "
```
kimberscott, line 33: this is so friendly!

accounts/views.py
```diff
@@ -1,3 +1,291 @@
...
+
+        Leverage the fact that `ModelFormMixin.form_valid` sets `self.object = form.save()`
+        prior to calling the super() method (`FormMixin.form_valid`), which actually does
+        the work of constructing the HttpRedirectResponse for us.
+        """
+        user: User = getattr(self, "object")
+        view_target = (
+            "accounts:2fa-setup"
+            if user.is_researcher
+            else "web:demographic-data-update"
```
kimberscott, line 105: ooh actually we shouldn't be sending participants to demographic data update unless they have no existing demographic data. Otherwise it'll be annoying to get sent to the demographic form every time they log in.

exp/templates/exp/_navigation.html
```diff
@@ -57,3 +55,8 @@
     </div><!-- /.navbar-collapse -->
   </div><!-- /.container-->
 </nav>
+<script type="application/javascript">
+    $(function () {
+        $('[data-toggle="tooltip"]').tooltip()
+    });
+</script>
```
kimberscott, line 62: are we using tooltips on this page / does this end up covering other pages magically?

exp/views/mixins.py
```diff
@@ -1,23 +1,55 @@
...
+                    request,
+                    "Please sign in with your researcher account to see Experimenter app.",
+                )
+                return HttpResponseRedirect(reverse("login"))
 
 
 class SingleObjectParsimoniousQueryMixin(SingleObjectMixin):
 
-    object: Union[Study, Lab]
+    object: Model
```
kimberscott, line 52: ah-hah, THAT is what we wanted! very neat.

project/settings/defaults.py
```diff
@@ -227,46 +226,16 @@
...
 BASE_URL = os.environ.get(
     "BASE_URL", "https://localhost:8000"
 )  # default to ember base url
-OSF_URL = os.environ.get(
-    "OSF_URL", "https://staging.osf.io/"
-)  # default osf url used for oauth
 
-LOGIN_REDIRECT_URL = os.environ.get("LOGIN_REDIRECT_URL", "https://localhost:8000/exp/")
+LOGIN_REDIRECT_URL = "web:home"
 ACCOUNT_LOGOUT_REDIRECT_URL = os.environ.get("ACCOUNT_LOGOUT_REDIRECT_URL", "/api/")
```
kimberscott, line 236: completely unrelated but this seems like an odd default - I guess it doesn't matter because we have the env variable set

web/views.py
```diff
@@ -106,12 +104,14 @@ class ParticipantUpdateView(LoginRequiredMixin, generic.UpdateView):
     """
     Allows a participant to update their name and password -
     extra code in this view because there are multiple forms on this page.
+
+    DEPRECATED: Now using
```
kimberscott, line 108: are we actually using this?

accounts/backends.py
Datamance: oh dear that's not right. Initially it was, but then I started using it again for Admin views

accounts/forms.py
Datamance: IMO the lab member-adding logic is fine where it is - is there a particular reason you'd want to bring it into the save method?

accounts/models.py
Datamance: Excellent point!

accounts/tests.py
Datamance: nah, not quite sure why I did that - maybe I was switching from testing one thing to another and just switched the tested element without thinking about the semantics? Regardless, it's changed now

exp/templates/exp/_navigation.html
Datamance: It'll cover pages where this partial is injected - and it's just to create the bootstrap-skinned tooltips. This was just to provide a nice bootstrappy tooltip over the original 2FA badge, which we dropped, so I'll go ahead and get rid of this.

project/settings/defaults.py
Datamance: Yeah I saw that! Barrett Harber added it in 2017 and it hasn't been touched since ಠ_ಠ

web/views.py
Datamance: Nah, I'm nuking it now

accounts/signals.py
Datamance: We should be fine - the only code that touches SocialApp model is in the osf adapter app, and that's now disconnected from the main app and just needs to be cleaned up. The fact that Travis seems to be running without dying at test DB initialization seems promising

accounts/forms.py
kimberscott: I was under the impression you didn't like putting logic into the signals if we didn't have to, but it works fine there - I'm happy with it if you are!

accounts/views.py
Datamance: Well actually this is `ResearcherRegistrationView`, not `ParticipantSignupView`, so I'm not really sure why I had that check in the first place... gonna delete it

Datamance: Also if they're registering and logging in as their user account for the first time, they'll never have demographic data

kimberscott:
Ack, I'm sorry - I just realized we neglected to deal with the changes this implies for the API. I think nothing has changed for experiment sessions, but right now, researchers access the API by authenticating with a Lookit token + OSF token. We'll need to figure out how that works now.

Issue #181, opened 2018-11-15
SCOPING: Support studies in languages other than English, and with multiple language options


kimberscott:
*Pain point*: Researchers would like to be able to test families in multiple languages, using translated materials. Right now that would require making multiple versions of the same study, and there is no way for parents to simply see the appropriate version.

Notes from an email to a researcher interested in this functionality, regarding scoping needed:
The reason that more general multiple-language support, while planned, has not been planned immediately is that pretty quickly it affects a lot of UI decisions for both participants and researchers. 
* Should researchers specify for each study which languages it is available in, from a list of potential languages? How should we treat slightly-different dialects that might matter for language studies but not others (e.g., providing British-accented and American-accented versions of voiceover stimuli)? Do we need to know whether the parent needs to understand a spoken vs. written language to participate? How should researchers specify which languages the parent needs to speak vs. which languages the child needs to be learning? This is probably also the time to integrate information about whether the parent and/or child to be able to hear and/or see to participate, and to provide alternate versions of stimuli where possible for blind and/or DHH kids and parents. 
* How should different-language versions of study stimuli actually be specified by the researcher - e.g., as described above, versus a more general standard way across all frames of specifying that stimuli should come from different locations based on the language selected? (The latter is very doable and probably fits in with some ways we've been specifying parameters when randomizing, but I'd want a bit more time to plan and test such a setup. It would be nice to integrate some "automatic" translation in addition for common elements researchers aren't currently required to provide text for, like "next" buttons or troubleshooting text.)
* How should families indicate what languages they & their children speak and to what degree? Do we want a list of languages in preference order or just a list of languages? Should this directly affect what studies children are "eligible" for? (And should it apply separately to parents' / children's capabilities? - e.g., the parent needs to speak English / Spanish / Mandarin in order to do the consent form and/or understand instructions for study A, but it doesn't matter that the child is being raised hearing almost exclusively Cantonese because the stimuli are just pictures and music - but for study B, the child also needs to be speaking one of the listed languages?) 
[...]

PR #527, opened 2020-07-20
Many babies at home - internationalization

rhodricusack:
Hi,
I've been working on support for internationalization within LookIt.

I modified
* templates, using the {% trans "..." %}
* form and other code, using django.utils.translation import gettext_lazy 

To test things out, we've started working on a Japanese locale file (locale/ja/LC_MESSAGES/django.po) currently just automatically generated. Do you agree that it would be easiest to have a single locale file per language (achieved by setting LOCALE_PATHS in defaults.py)? 

I modified project/settings/local-dist.py to include the hint:
# SESSION_COOKIE_SECURE= False # For local debugging
as this took me a while to discover.

I added two locale-enabling lines to projects/settings/defaults.py 

I modified project/urls.py to support language specific URLs like /ja/studies for Japanese. 

Our changes are far from complete, but I felt it would be best to get your feedback early on the changes, and as so many files are changing, to make sure that we don't end up with a nasty merge problem later.

I look forward to your feedback, thank you very much for the elegant codebase.
Best, Rhodri

### DIFF
--- a/accounts/templates/account/local_user_exists.html
+++ b/accounts/templates/account/local_user_exists.html
@@ -10,17 +10,16 @@
 <div class="container">
   <div class="jumbo" style="padding-top: 25px">
       <h2 style="font-size: 48px; text-align: center; font-weight: 300;">
-          Login Unsuccessful
+          {% trans "Login Unsuccessful" %}
       </h2>
       <h3 style="text-align: center; font-weight: 300;">
-        A Lookit study participant exists with the email address for your OSF user account.
-        Create an OSF user with a different email address and try logging into Experimenter with that.
+        {% trans "A Lookit study participant exists with the email address for your OSF user account. Create an OSF user with a different email address and try logging into Experimenter with that. " %}
       </h3>
       <h3 style="text-align: center; font-weight: 300;">
-          If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io</a>.
+          {% trans "If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io" %}</a>.
       </h3>
       <h4 style="text-align: center; font-weight: 300;">
-          <a href="/accounts/login/">Return to Experimenter</a>
+          <a href="/accounts/login/">{% trans "Return to Experimenter" %}</a>
       </h4>
   </div>
 </div>
--- a/accounts/templates/account/login_errored_cancelled.html
+++ b/accounts/templates/account/login_errored_cancelled.html
@@ -10,13 +10,13 @@
 <div class="container">
   <div class="jumbo" style="padding-top: 25px">
       <h1 style="font-size: 48px; text-align: center; font-weight: 300;">
-          Login Unsuccessful
+          {% trans "Login Unsuccessful" %}
       </h1>
       <h4 style="text-align: center; font-weight: 300;">
-          If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io</a>.
+          {% trans 'If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io' %}</a>.
       </h4>
       <h3 style="text-align: center; font-weight: 300;">
-          <a href="/exp/">Return to Experimenter</a>
+          <a href="/exp/">{% trans "Return to Experimenter" %}</a>
       </h3>
   </div>
 </div>
--- a/accounts/templates/accounts/participant_detail.html
+++ b/accounts/templates/accounts/participant_detail.html
@@ -1,6 +1,7 @@
 {% extends 'exp/base.html' %}
 {% load bootstrap3 %}
 {% load exp_extras %}
+{% load i18n %}
 
 {% block title %}{{ user.nickname |default:"Participant Detail" }}{% endblock %}
 {% block flash %}
@@ -11,7 +12,7 @@
       <div class='row'>
           <div class="col-xs-12">
               <ol class="breadcrumb">
-                <li><a href="{% url 'exp:participant-list' %}"> All Participants </a></li>
+                <li><a href="{% url 'exp:participant-list' %}"> {% trans "All Participants" %} </a></li>
                 <li class="active"> {{ user.nickname |default:"Participant Detail" }} </li>
               </ol>
           </div>
@@ -19,22 +20,22 @@
 
       <div class="row">
           <div class="col-xs-12">
-              <h1> {{ user.identicon_html }} {% if user.nickname %} {{ user.nickname }}{% else %} Participant ID {{ user.id }} {% endif %}</h1>
-              <span><label class='pr-xs'> Last Active: </label> {{ user.last_login|date:"n/d/Y"|default:"N/A" }} </span>
-              <div><label class='pr-xs'> Participant global ID: </label> {{ user.uuid }} </div>
+              <h1> {{ user.identicon_html }} {% if user.nickname %} {{ user.nickname }}{% else %} {% trans "Participant ID" %} {{ user.id }} {% endif %}</h1>
+              <span><label class='pr-xs'> {% trans "Last Active:" %} </label> {{ user.last_login|date:"n/d/Y"|default:"N/A" }} </span>
+              <div><label class='pr-xs'> {% trans "Participant global ID:" %} </label> {{ user.uuid }} </div>
           </div>
       </div>
 
     <div class='row'>
         <div class='col-md-6 col-sm-5'>
-            <h3>Children</h3>
+            <h3>{% trans "Children" %}</h3>
             {% for child in children %}
             <div class="pb-md">
                 <h4 class="child-name"><em>{{child.given_name}}</em></h4>
                 <div class="pl-md">
                     <div class="row">
                         <div class="col-xs-5">
-                            <b> Birthday: </b>
+                            <b> {% trans "Birthday" %}: </b>
                         </div>
                         <div class="col-xs-7">
                             {{child.birthday|date:"n/d/Y"|default:"N/A"}}
@@ -42,7 +43,7 @@
                     </div>
                     <div class="row">
                         <div class="col-xs-5">
-                            <b> Age: </b>
+                            <b> {% trans "Age" %}: </b>
                         </div>
                         <div class="col-xs-7">
                              {{child.birthday | timesince}}
@@ -50,7 +51,7 @@
                     </div>
                     <div class="row">
                         <div class="col-xs-5">
-                            <b> Gender: </b>
+                            <b> {% trans "Gender" %}: </b>
                         </div>
                         <div class="col-xs-7">
                              {{child.get_gender_display}}
@@ -58,7 +59,7 @@
                     </div>
                     <div class="row">
                         <div class="col-xs-5">
-                            <b> Gestational Age at Birth: </b>
+                            <b> {% trans "Gestational Age at Birth" %}: </b>
                         </div>
                         <div class="col-xs-7">
                              {{child.age_at_birth|default:"No response"}}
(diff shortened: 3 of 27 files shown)


accounts/templates/account/local_user_exists.html
```diff
@@ -10,17 +10,16 @@
...
+          {% trans "Login Unsuccessful" %}
       </h2>
       <h3 style="text-align: center; font-weight: 300;">
-        A Lookit study participant exists with the email address for your OSF user account.
-        Create an OSF user with a different email address and try logging into Experimenter with that.
+        {% trans "A Lookit study participant exists with the email address for your OSF user account. Create an OSF user with a different email address and try logging into Experimenter with that. " %}
       </h3>
       <h3 style="text-align: center; font-weight: 300;">
-          If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io</a>.
+          {% trans "If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io" %}</a>.
```
kimberscott, line 19: should this include the closing tag or not include the link at all?

accounts/templates/account/login_errored_cancelled.html
```diff
@@ -10,13 +10,13 @@
 <div class="container">
   <div class="jumbo" style="padding-top: 25px">
       <h1 style="font-size: 48px; text-align: center; font-weight: 300;">
-          Login Unsuccessful
+          {% trans "Login Unsuccessful" %}
       </h1>
       <h4 style="text-align: center; font-weight: 300;">
-          If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io</a>.
+          {% trans 'If this should not have occurred and the issue persists, please report it to <a href="mailto:support@osf.io">support@osf.io' %}</a>.
```
kimberscott, line 16: same comment here - I don't know if this causes any issues but should probably either include the closing tag or skip the link entirely

accounts/templates/accounts/participant_detail.html
```diff
@@ -11,62 +12,62 @@
       <div class='row'>
           <div class="col-xs-12">
               <ol class="breadcrumb">
-                <li><a href="{% url 'exp:participant-list' %}"> All Participants </a></li>
+                <li><a href="{% url 'exp:participant-list' %}"> {% trans "All Participants" %} </a></li>
```
kimberscott, line 15: are the quotes needed here for something?

accounts/templates/accounts/participant_detail.html
```diff
@@ -75,78 +76,78 @@ <h4 class="child-name"><em>{{child.given_name}}</em></h4>
...
+                    <b> {% trans "Explanation for Guardians:" %} </b>
                 </div>
                 <div class="col-xs-7">
                     {{ demographics.number_of_guardians_explanation|default:"No Response" }}
                 </div>
             </div>
             <div class="row">
                 <div class="col-xs-5">
-                    <b> Race: </b>
+                    <b>{% trans "Race" %}:</b>
```
kimberscott, line 150: will this sort of thing need a contextual marker or be fine?

project/settings/local-dist.py
```diff
@@ -18,3 +18,5 @@
 EMAIL_BACKEND = (
     f"django.core.mail.backends.{'console' if DEBUG else 'smtp'}.EmailBackend"
 )
+
+# SESSION_COOKIE_SECURE= False # For local debugging
```
kimberscott, line 22: This shouldn't *in general* be necessary - I'd hold off on incorporating it as part of internationalization, but we should look at it when we update setup instructions

accounts/templates/account/local_user_exists.html
rhodricusack: I've put the closing tag inside the {% trans } and made the inner quotes single.

`
{% trans "If this should not have occurred and the issue persists, please report it to <a href='mailto:support@osf.io'>support@osf.io</a>" %}.
`

project/settings/local-dist.py
rhodricusack: My ubuntu workstation won't offer https:/// connections on localhost without me configuring a certificate. It took me quite a long time that it was the lack of a stateful session that was causing all manner of odd problems with the authentication... it seemed to authenticate, but then acted like it didn't know me on the new page. I think I'd changed a few hundred lines by the time I discovered it was just this one. So, I agree - not at all related to internationalization, and would be well placed in the instructions.  If you'd like Ubuntu 18.04 setup instructions I've been through a few iterations and now have a pretty tidy list I think.

accounts/templates/accounts/participant_detail.html
rhodricusack: It does seem like a candidate, I'll highlight this general issue (with this example) to the translation group, and perhaps we will probably need to bring in contextual markers where needed.

rhodricusack:
On the flat pages, there doesn't seem to be any internationalization support within django.contrib.flatpages. From a brief browse these two alternatives seem suggested:
[django-flatpages-i18n](https://github.com/PragmaticMates/django-flatpages-i18n). Seems to have been around for a while and to be regularly maintained. Looks simple, if you have no need for fancier features here, perhaps worth evaluating first.   
[django CMS](http://docs.django-cms.org/en/latest/)  Full featured option! 
Do you have any opinion here? If not I can evaluate these. I don't know if there will be encoding issues with storing unicode in the postgresql?

A similar issue will arise for the internationalization of the studies themselves. I was thinking here that a study in a different languages should effectively be a different studie, by adding a language indicating field (e.g., "ja" for Japanese) to studies_study that would allow for filtering on the "Current Studies" page. Would there be any downside to them effectively being different studies? Like some maintenance aspect (permissions?) or summary reports?

Datamance:
> I modified project/settings/local-dist.py to include the hint:
> 
> # SESSION_COOKIE_SECURE= False # For local debugging

Hi Rhodri,

I recommend following the instructions [here](https://github.com/lookit/lookit-api/pull/529#pullrequestreview-452829111) (assuming you are using a mac) to set up a local root CA and ssl cert so that you can serve HTTPS locally. It'll become useful if you ever want to test the webhook functionality for video storage, and it fits with the notion of aligning your dev environment to production as much as possible.

project/settings/local-dist.py
Datamance: @rhodricusack is there a reason why you can't configure a local certificate? `mkcert` might be able to do most of the legwork for you, and then you'd just have to configure `runserver_plus` with the `--cert-file` flag.

kimberscott:
@Datamance did you mean to link to our docs or something else?

Datamance:
> @Datamance did you mean to link to our docs or something else?

Good catch, I meant to post [this link](https://blog.arrogantrabbit.com/ssl/Root-CA-macOS/)

rhodricusack:
Hi
Thanks for the directions! The main thing that took me time was to discover that it was the HTTP connection that was the issue. Django didn't throw any errors, it just didn't become stateful! Very confusing. This would work in the docs, I've taken out the line

# SESSION_COOKIE_SECURE= False # For local debugging

in the pull request.

I did try to follow some instructions for a local cert (I'm using Ubuntu 18.04) but it didn't work first time, and in any case at the moment I'm just using the django-admin runserver, which I don't think supports https. I can switch to apache2 and mod-wsgi and make a self-signed cert.

Best, Rhodri


________________________________

rhodricusack:
Hi @kimberscott @Datamance,
I'm not an experienced pull-requester and have no idea regarding etiquette. I'm just checking in what the next stage for this is. Do you agree that merging as we go along would be better than waiting until the end, to give us fewer complications in the merge? Is there something I'm supposed to do now? Or should I just be patient? 
Best, Rhodri

kimberscott:
Yes - merging as we go makes sense overall (not waiting until all translation work is done), although with some batching to avoid lots of deployments to staging/production. If there's work you're continuing now you're welcome to keep updating this branch. 

We currently have some changes in the works (mostly regarding the login setup, removing OSF authentication and adding 2FA - see https://github.com/lookit/lookit-api/pull/526). I think it will make sense to merge those in first, then the internationalization likely next week. We can help resolve any merge conflicts and find cases where we may have introduced un-translated text in these recent changes, and then going forward, make sure we keep participant-facing text in {% trans %} tags.  but I'm also assuming that the exact timing doesn't affect ManyBabies as you're not directly using Lookit yet - is that accurate?


We also always first merge into the `develop` branch, which deploys to lookit-staging.mit.edu, and then `master` which controls lookit.mit.edu. I think I can just edit this PR to target develop, just a heads-up so it's not surprising!

rhodricusack:
Hi Kim
That sounds great. We're not using it yet - we're just gearing up on many fronts.

In response to your feedback, and to fix a couple of small things, I made some further changes which are new commits. Will they automatically be included in the pull request, or do I need to do something?

All the best
Rhodri
________________________________

kimberscott:
Sounds good - and yes, those commits are automatically included here!

kimberscott:
> On the flat pages, there doesn't seem to be any internationalization support within django.contrib.flatpages. From a brief browse these two alternatives seem suggested:
> [django-flatpages-i18n](https://github.com/PragmaticMates/django-flatpages-i18n). Seems to have been around for a while and to be regularly maintained. Looks simple, if you have no need for fancier features here, perhaps worth evaluating first.
> [django CMS](http://docs.django-cms.org/en/latest/) Full featured option!
> Do you have any opinion here? If not I can evaluate these. I don't know if there will be encoding issues with storing unicode in the postgresql?

My apologies, I think we never responded to this - no opinions here so if you're up for evaluating that'd be great. We aren't making heavy or fancy use of flat pages - we have the home page, "meet the scientists," FAQ, resources, and privacy policy/terms of use. It *is* nice to be able to change these without deploying the whole app but they're pretty stable.

> A similar issue will arise for the internationalization of the studies themselves. I was thinking here that a study in a different languages should effectively be a different studie, by adding a language indicating field (e.g., "ja" for Japanese) to studies_study that would allow for filtering on the "Current Studies" page. Would there be any downside to them effectively being different studies? Like some maintenance aspect (permissions?) or summary reports?

As an initial solution, assuming we're dealing with a smallish number of studies in a smallish number of languages, just making a translation a separate study will work fine (and yes, could likely be integrated into planned filtering on the studies page). For a solid longer-term solution there are a lot of things to think about (see https://github.com/lookit/lookit-api/issues/181) and I think it'll be worth figuring out a way to represent translations as translations rather than as separate studies for maintenance reasons (managing permissions, reviewing consent - although I guess actually you might not have the same people reviewing in all languages!, starting/stopping data collection, letting families know if they've already participated, downloading/combining data, review/approval). 

Because the login changes touched a lot of templates we have some merge conflicts to resolve before proceeding - want me to take a stab at that and you can confirm things still look ok, or would you prefer to handle those?

rhodricusack:
Thanks for following up. I'll take a look at django-flatpages-i18n<https://github.com/PragmaticMates/django-flatpages-i18n>. Could you share the database contents for this table and I can check the pages renders properly and I'll also send them to the translation teams?

On the experiments, it seems like a key issue here is how much we would want different permissions for different languages. There may be cases where people local to a jurisdiction have to review the consents before it is accessible by other teams. Given this maybe it would be easiest to set then up as different studies to start and then we'll know exactly what is needed when implementing and more elegant solution.

I am happy to merge if you like. Would I just pull your latest changes (development branch?) and resolve the conflicts locally? Or I'm also happy for you to do it!

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

kimberscott:
Oh, that's a good point about the permissions/teams potentially being different. Sounds like a plan to start with separate studies & gradually refine what's needed.

If you're up for merging changes that'd be great - thanks! But just let me know if there are cases where it's confusing. Yes, you'd just pull the dev branch and resolve conflicts locally.

PR #532, opened 2020-08-06
Feature/511 multi factor authentication

Datamance:
Two things:
1) Bye bye allauth
2) fix password reset issue


kimberscott:
(Late, but LGTM! Read through the code and ran migration locally.)

Issue #157, opened 2018-10-03
Survey study support


kimberscott:
Does not require substantial development, but need to set up for non-video studies w/ IRB so that we can have a modified consent procedure for studies that do not involve any recording. Might also want to have the option for the data to be available to all Lookit researchers.

PR #534, opened 2020-08-08
Feature/157-protocol generator

kimberscott:
Changes to allow researchers to specify protocol generator rather than static protocol json as described in #157

Accompanies lookit/lookit-docs#95, lookit/ember-lookit-frameplayer#158; addresses lookit/ember-lookit-frameplayer#157


studies/forms.py
```diff
@@ -223,7 +309,7 @@ class StudyEditForm(StudyForm):
     """Form for editing study"""
 
     def __init__(self, *args, **kwargs):
-        user = kwargs.get("user")
+        user = kwargs.pop("user")
```
Datamance, line 312: Why are we `pop`ing here? I've seen this in a couple of places. Is the mutation intentional? If it's not, better to use `get` or `brackets`

studies/forms.py
```diff
@@ -94,13 +159,27 @@ class StudyForm(ModelForm):
...
-    def __init__(self, *args, **kwargs):
-        # Limit lab options to labs this user is a member of
-        user = kwargs.pop("user")
-        super().__init__(*args, **kwargs)
-        self.fields["lab"].queryset = user.labs.filter(
-            id__in=user.labs_user_can_create_study_in()
-        )
+    # Define initial value here rather than providing actual default so that any updates don't
+    # require migrations: this isn't a true "default" value that would ever be used, but rather
+    # a helpful skeleton to guide the user
```
Datamance, line 164: Image

studies/forms.py
```diff
@@ -46,6 +50,59 @@ class Meta:
 
 PROTOCOL_HELP_TEXT_INITIAL = f"{PROTOCOL_HELP_TEXT_EDIT}  You can leave the default for now and come back to this later."
 
+DEFAULT_GENERATOR = """function generateProtocol(child, pastSessions) {
```
Datamance, line 53: Do we want to have tests to see that the default generator always parses (would probably require a dev dependency like [pyjsparser](https://github.com/PiotrDabkowski/pyjsparser)), or is it pretty much not going to change after this? Up to you, just throwing that out there.

exp/views/study.py
```diff
@@ -194,16 +194,22 @@ def get_form_kwargs(self):
     # UserPassesTestMixin.get_test_func()
     test_func = user_can_edit_study
 
+    # Rely on form validation to make sure structure is valid JSON; don't load/dump so we
+    # preserve formatting and ordering.
+    #
```
Datamance, line 199: get rid of empty comment line

exp/views/study.py
```diff
@@ -194,16 +194,22 @@ def get_form_kwargs(self):
...
     def get_initial(self):
         """
```
Datamance, line 201: Fix indentation + adhere to Google style docstrings

studies/templates/studies/_study_fields.html
```diff
@@ -85,4 +186,12 @@
 {% bootstrap_field form.duration %}
 {% bootstrap_field form.contact_info %}
 {% bootstrap_field form.lab %}
-
+<div id="structure-container">
```
Datamance, line 189: Looking at the [django-bootstrap3 docs](https://django-bootstrap3.readthedocs.io/en/latest/templatetags.html#bootstrap3.templatetags.bootstrap3.bootstrap_field), shouldn't the field partial already render with error elements? If you really want to change the default class, I guess you could (see `error_css_class`), but I don't really see the point here

exp/static/exp/css/app.css
```diff
@@ -533,6 +533,10 @@ a:hover.build-study-link {
     max-width: 50em;
 }
 
+p.form-validation-error {
```
Datamance, line 536: See comment below about `bootstrap_field` template helper

studies/forms.py
kimberscott: Yes, intentional because we use `user` here to determine the lab queryset but don't want to include it in the call to `super().init`. This is also done e.g. in the new `TOTPCheckForm` so I thought it was ok?

studies/forms.py
kimberscott: So my feeling on server-side validation here of the javascript overall was that the extra dependencies we'd need wouldn't quite be worth the benefit, and would introduce a tighter coupling between ember-lookit-frameplayer and lookit-api than we really want to deal with maintaining. Only introducing this for tests is an interesting compromise - but yeah, I think the default pretty much shouldn't change after this. (It also shouldn't be used directly, so it's essentially a cosmetic issue if it's ever flagged as invalid, albeit one we should avoid!) I'll keep it in mind in case we have any more cases where JS parsing would be relevant, or upon changing the default. 

Eventually it'd definitely be helpful on a variety of templates to be able to test the JS that's *on the template* in context - e.g. render the page, do some action, check that the right thing is showing up.

exp/views/study.py
kimberscott: sorry about that! edited this comment which was actually a relic of trying a variety of approaches to smoothly move to checking for valid JSON but storing the actual string so that from the user's standpoint it's like a big text box.

studies/forms.py
Datamance: sounds reasonable to me!

exp/views/study.py
kimberscott: Fixed a few docstrings here (not feeling super ambitious about taking on all the docstrings of classes I touched at the moment, but let me know if you meant all of them)

Datamance: no, just the classes/functions you touch is fine

studies/templates/studies/_study_fields.html
kimberscott: Ooh, okay, I think my thinking was I shouldn't mess with the proper bootstrap error elements when adding client-side validation, to avoid interfering with errors added when actually submitting the form. Changing now.

kimberscott: (The reason they're in their own divs is just to make the JS for identifying the appropriate editor elements more natural and less dependent on the details of how the ace-editor gets inserted/what IDs are used.)

studies/forms.py
Datamance: Uy, no 🙈  my bad.

TOTPCheckForm's initializer should be like so:
```
    def __init__(self, request, *args, **kwargs):
        self.request = request
        self.otp = getattr(request.user, "otp")
        super().__init__(*args, **kwargs)
```

Argument parsing shouldn't be done imperatively - let the signature itself decide what's expected at call time.

PR #538, opened 2020-08-18
Announcement Emails

### LINKED ISSUE #132: Expanded email functionality: announcement emails
*Pain point*: Email announcements of new studies for a child are currently handled by a collection of python scripts run on Kim's laptop, writing to / checking from text files of email addresses already notified. This will not scale well and allows no independence for researchers in turning on/off notifications, specifying what to say, etc. Researchers cannot send these emails themselves because they (appropriately) do not have the ability to email Lookit users who have not already participated in their studies. 

*Acceptance criteria*:
* Researchers can specify, for each study, the message for an "announcement" email that is automatically sent daily to any users with at least one child eligible for a study who has not previously participated in or previously received an announcement about this study. (Ideally this would be sent as a daily digest, one email per participant even if multiple studies have become available.)
* Researchers can turn on/off announcements for their study.
* Announcements are off if the study is non-discoverable. 

*Implementation notes/Suggestions*: Centralizing email announcements of new studies allows us to prevent participants from receiving overwhelming numbers of announcements. Sending study announcements will allow us to better leverage the existing participant userbase for recruitment as the platform grows, and to keep families engaged with research.

Datamance:
Includes parsimonious target selection, template generation.

Fixes #132


studies/tasks.py
```diff
@@ -39,6 +47,166 @@
...
+            yield user, valid_message_targets
+
+
+def _segmented_by_study(validated_groups):
+    for user, child_study_pairs in validated_groups:
+        yield user, dict(map_reduce(child_study_pairs, itemgetter(1), itemgetter(0)))
+
+
+@app.task
+def send_announcement_emails():
```
kimberscott, line 190: Out of curiosity what is the process for actually adding the task to do this daily?

studies/tests.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # First level is the same - it's the same user...
+        self.assertEqual(len(targets), 1)
+        user, study_child_mapping = targets[0]
+        self.assertEqual(user, self.participant_two)
+
+        # ... But no study one this time.
+        self.assertDictEqual(
+            study_child_mapping, {self.study_two: [self.child_two, self.child_three]},
+        )
```
kimberscott, line 246: Would propose a few additional tests (or additions to existing tests) here:
* No targets generated for Study that is active but non-discoverable or paused but discoverable
* No targets generated for User with email_new_studies set to false
* No (user, study, child) target if child is older than max age for study or younger than min age for study
* No (user, study, child) target if child does not meet criteria expression (e.g. child has to speak Spanish to participate)

studies/tests.py
```diff
@@ -0,0 +1,246 @@
...
+        self.assertEqual(message_object.related_study, self.study_two)
+        self.assertIsNone(message_object.sender)
+
+    def test_study_excluded_from_targets_after_message(self):
+        Message.send_announcement_email(
+            self.participant_two, self.study_one, [self.child_three]
+        )
+        # New targets should have only study two targets
+        targets = list(
```
kimberscott, line 232: could we have a helper function for `list(_segmented_by_study(_validated(_deserialized(_grouped_by_user(potential_message_targets())))`? It seems like that's how these are really designed to be used, and it might be easier to reason about appropriately updating tests in the event of any changes to the logic.

studies/tasks.py
Datamance: Actually, we need to add a row in `django_celery_beat_periodictasks` - I should take care of that in a migration

studies/templates/emails/study_announcement.html
```diff
@@ -0,0 +1,19 @@
+
+<p>Dear {{ user.nickname }},</p>
+
+<p>We're writing to invite you and your children to participate in the study "{{ study.name }}" that {{ study.lab.name }} is running on Lookit!</p>
+
+<ul>
+    <li>Who: <ul>{% for child in children %}<li>{{ child.given_name }}</li>{% endfor %}</ul></li>
+    <li>What happens: {{ study.short_description }}</li>
+    <li>Why: {{ study.long_description }}</li>
+    <li>Compensation: {{ study.compensation_description|default_if_none:"This is a volunteer-based study." }}</li>
```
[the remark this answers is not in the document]
Datamance: Fixed - I glossed over that one. I'll add a test for it.

studies/tasks.py
```diff
@@ -39,6 +47,166 @@
...
+            )
+
+
+def _validated(deserialized_groups):
+    """Yield only groups with targets that satisfy criteria for their respective studies."""
+    for user, child_study_pairs in deserialized_groups:
+        valid_message_targets = []
+        for pair in child_study_pairs:
+            child, study = pair
+            eligible = get_child_eligibility_for_study(child, study)
```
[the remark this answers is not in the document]
Datamance: What do you want to do in the case that the `birthday` field is null?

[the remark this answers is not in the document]
Datamance: Resolved per slack discussion

studies/tests.py
Datamance: Done. Was thinking that we should explicitly test criteria expression stuff in that suite, but I guess it makes sense to see that the checker was called properly.

studies/templates/emails/study_announcement.html
```diff
@@ -0,0 +1,19 @@
+
+<p>Dear {{ user.nickname }},</p>
+
+<p>We're writing to invite you and your children to participate in the study "{{ study.name }}" that {{ study.lab.name }} is running on Lookit!</p>
```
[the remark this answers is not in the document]
kimberscott: for translatability going forward, should we use [pluralize](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#pluralize) here?

studies/templates/emails/study_announcement.html
```diff
@@ -0,0 +1,19 @@
...
+
+<ul>
+    <li>Who: <ul>{% for child in children %}<li>{{ child.given_name }}</li>{% endfor %}</ul></li>
```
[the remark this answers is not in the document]
kimberscott: Is there some not-too-awkward thing to say after study.criteria to indicate the children listed might not be eligible? "This may include:"? I'm thinking of existing studies where criteria include currently being in the US, for instance (for IRB/funding reasons).

studies/tests.py
kimberscott: 🎈 🍰 am I doing this right

kimberscott:
LGTM aside from very minor email-template comments - if you want to start testing on staging let's go for it

PR #539, opened 2020-08-18
protocol generator: display initial value for generator function in study update form 

kimberscott:
otherwise this is looking good on staging.


exp/views/study.py
```diff
@@ -219,6 +221,8 @@ def get_initial(self):
                 initial["structure"] = structure["exact_text"]
             else:
                 initial["structure"] = json.dumps(structure)
+        if not (type(self.object.generator) is str and self.object.generator.strip()):
```
Datamance, line 224: prefer `issubclass(type(self.object.generator), str)`

Although doesn't the model already restrict you with `TextField`? Why not just `if self.object.generator`?

PR #542, opened 2020-08-22
standardize child age calculation for announcements; update announcement template

kimberscott:
Sorry for not catching this earlier - we define a year as 365 days throughout, so to avoid people getting notifications when they "age in" but then being told they're not eligible when they get to Lookit (which would be the case for every study for children at least 1 year old), use that definition here too.

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -596,11 +596,13 @@ class Message(models.Model):
             Creates a corresponding message object in the database.
         """
         subject = create_subject_for_study_notification(study, children)
+        children_string = create_string_listing_children(children)
         context = {
             "base_url": settings.BASE_URL,
             "user": user,
             "study": study,
             "children": children,
+            "children_string": children_string,
         }
 
         text_content = get_template("emails/study_announcement.txt").render(context)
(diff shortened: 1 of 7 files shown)


accounts/models.py
```diff
@@ -657,16 +659,29 @@ def send_as_email(self):
         self.save()
 
 
-def create_subject_for_study_notification(study, children):
-    latter_half = f' invited to take part in "{study.name}" on Lookit!'
+def create_string_listing_children(children):
     child_names = [child.given_name for child in children]
+    num_children = len(child_names)
```
kimberscott, line 664: apologies for not using `:=` here - precommit was balking at it even as already used?

Datamance: How were you trying to use it?

Datamance: oh I see what you're saying - yeah, that's strange. If our precommit environments are identical that shouldn't happen, unless something actually changed in the code itself that would invalidate the syntax.

Datamance: on that note - make sure your precommit and virtualenv are both using 3.8.x (I think our pipenv env will force 3.8.3 on you)

PR #550, opened 2020-08-28
:bug: Deleted children excluded from target query

### LINKED ISSUE #548: Announcement message query should not include soft-deleted children
**Description**
The announcement message query pulls targets for children that have been soft-deleted. This should definitely not happen.

**How to reproduce**
This happens whenever announcement emails are fired off.

**Expected behavior**
Soft-deleted children should not be considered in the message-target creation process.

Datamance:
Fixes #548


kimberscott:
LGTM :)

PR #551, opened 2020-08-29
Bugfix/546 message participants fix - add tests!

kimberscott:
* Add contact view tests
* Limit recipients to study participants (beyond limiting options in UI)
* Test requirement for 2fa login on ALL exp views
* Use ExperimenterLoginRequiredMixin on ParticipantDetailView for consistency

### DIFF
--- a/exp/tests/test_contact_views.py
+++ b/exp/tests/test_contact_views.py
@@ -1,6 +1,283 @@
-# TODO: StudyParticipantContactView
-# - check can only GET if return user.is_researcher and user.has_study_perms(
-#             StudyPermission.CONTACT_STUDY_PARTICIPANTS, study
-#         )
-# - check context["participants"] has participants for this study but not another one
-# - check can post a single message (postpone checks of appropriate recipients & message type)
+import datetime
+
+from django.contrib.sites.models import Site
+from django.test import Client, TestCase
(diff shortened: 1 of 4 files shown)


exp/tests/test_contact_views.py
```diff
@@ -1,6 +1,285 @@
...
+class Force2FAClient(Client):
+    @property
+    def session(self):
+        _session = super().session
+        _session[TWO_FACTOR_AUTH_SESSION_KEY] = True
+        return _session
+
+
+# Run celery tasks right away, but don't catch errors from them.
+@override_settings(CELERY_TASK_ALWAYS_EAGER=True)
```
Datamance, line 24: Do we actually need these? The other thing we could do is use `mock.patch` decorator on the unit tests that call `some_task.delay` and check that the method was actually called (see [assert_called](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called)). In fact this is probably preferable; we have generally been bad about drawing distinctions between unit and integration tests but better start late than never.

kimberscott: Done - very cool, I didn't realize we could patch `some_task.delay` so easily!

Datamance (approved):
LGTM

PR #552, opened 2020-08-29
Invoke scripts cleanup

kimberscott:
*Minor* changes based on issues that have come up when people try to use the invoke scripts - this is not fully tested, but should only help. Proposing to merge in just so we don't have to maintain & point people to a separate invoke scripts branch.


requirements/defaults.txt
```diff
@@ -40,6 +40,7 @@ pygraphviz==1.5
 pyinstrument==3.1.3
 qrcode==6.1
 requests==2.21.0
+sendgrid-django==4.2.0
```
Datamance, line 43: Why is this being added here? This should pretty much only be in `requirements/release.txt`

kimberscott: Ohh changed back but it looks like both this and the sslserver move were because of how DEBUG is set: based on an environment variable in project/settings/defaults.py, then separately in /local.py. So if you don't have the environment variable set already, you end up with errors when trying to start the server because having DEBUG set to False during the rest of project/settings/defaults.py makes some incorrect choices based on the local value.

It looks like the invoke scripts also just initialize local.py with DEBUG=True rather than copying local-dist if local doesn't exist - should they do the latter?

We actually use `os.environ.get` a lot throughout defaults.py. So to let other people set up for development, we need to do one of
* add "adding the following environment variables" to docs and/or invoke scripts
* go through and check that in each case any values in local-dist.py correctly override the values in defaults (e.g., would need to override both `EMAIL_FROM_ADDRESS` and `DEFAULT_FROM_EMAIL`)
* use dotenv and provide a local-dist .env template
* restructure settings/__init__.py so it loads necessary local settings *first* (in addition to overriding afterwards?)

Datamance: Per slack discussion, recommend us switching over to using `dotenv`

kimberscott: Done! Can you take a look when you have a chance and see if this looks reasonable?

Pipfile
```diff
@@ -61,6 +62,7 @@ pydenticon = "==0.3"
 pyotp = "==2.3.0"
 qrcode = "==6.1"
 requests = "==2.21.0"
+sendgrid-django = "==4.2.0"
```
Datamance, line 65: this still should not be in the Pipfile - the only place where it's needed is in `release.txt` along with our other deployment-only dependencies

project/__init__.py
```diff
@@ -1,7 +1,24 @@
...
+except ImportError:
+    warnings.warn(
+        "Module dotenv not available; not importing local settings from .env",
+        ImportWarning,
+    )
+else:
+    # Load .env variables here, *before* setting remaining values, so that e.g. DEBUG from .env is used to
+    # determine other values. Note this will not overwrite any system-level environment variables (e.g. in .bashrc)
+    env_path = Path(__file__).resolve().parents[1] / ".env"
+    load_dotenv(dotenv_path=env_path, verbose=True)
```
Datamance, line 24: does `find_dotenv` not work?

```python
from dotenv import load_dotenv, find_dotenv
load_dotenv(find_dotenv())
```

Otherwise, why not
```python
Path("..") / ".env"
```
?

project/settings.py
```diff
@@ -14,11 +14,10 @@
 
 from django.contrib.messages import constants as messages
 
-MODE = "prod"  # Overridden by local settings.
 ENVIRONMENT = os.environ.get("ENVIRONMENT", "develop")
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
```
Datamance, line 20: Try
```python
Path(".").parent.resolve()
```

project/settings.py
```diff
@@ -14,11 +14,10 @@
...
 # root path for ember builds
 EMBER_BUILD_ROOT_PATH = os.path.join(BASE_DIR, "../ember_build")
```
Datamance, line 22: Let's just completely lean into [using pathlib](https://adamj.eu/tech/2020/03/16/use-pathlib-in-your-django-project/)

Datamance (changes requested):
mostly just suggestions to use pathlib

requirements/defaults.txt
Datamance: see new comments

project/__init__.py
kimberscott: Using find_dotenv now - thanks! but the reason not to do `Path("..") / ".env"` is that `.` refers to the current working directory, *not* the file, so the path would become dependent on where this file was run. Demo where `project/testfile.py` just prints out `Path(".").resolve()`: - 

```
(lookit-api) bash-3.2$ python project/testfile.py
/Users/kms/lookit-api
(lookit-api) bash-3.2$ cd ~
(lookit-api) bash-3.2$ python lookit-api/project/testfile.py
/Users/kms
(lookit-api) bash-3.2$ 
```

project/settings.py
kimberscott: switched to using Path, see above about why not using `.`

kimberscott: Actually though - upon going through changing the paths to use Path, I'd like to push back on the need to switch right now:

* These settings are expected in a variety of places (including Django code) to be strings rather than Path objects. So we need to either be careful to cast them to strings whenever passing them (e.g., call to `static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)`) or need to store them as strings. The latter is doable enough - use Path to construct the path, cast to a string to store, then construct new paths like `str(Path(BASE_DIR, 'subdirectory'))`. But Path objects by design always strip trailing slashes, so we'd be back to using os.path anyway to construct e.g. `MEDIA_ROOT`. 

* Upon considering the various settings for local assets (e.g. `STATICFILES_LOCATION = "/static"`), we might also need to update those to be filesystem-independent - but that would involve changing usage in templates too...

Overall this just seems like a separate endeavor and, although potentially worth doing, not a priority right now - is that ok?

project/__init__.py
Datamance: Ah, right - in that case, I'd probably pass [`sys.argv[0]`](https://docs.python.org/3/library/sys.html#sys.argv) since it's guaranteed to be the location of the manager script (`__file__` can be weird sometimes, particularly on windows though that doesn't affect us much)

project/settings.py
Datamance: hmmmm... In theory, since we use Python 3.6+, Path objects should be acceptable wherever strings used to be (for example, `os.path.join` accepts `Path` objects as well)...

... Ah, I see - we need to upgrade to 3.1 if we want [Path objects to work for stuff like staticfiles dirs](https://github.com/django/django/commit/77aa74cb70dd85497dbade6bc0f394aa41e88c94). This is a bug that's been fixed.

Up to you - if you decide to defer the `Path` conversion can you make a ticket for it and assign it to me?

project/settings.py
Datamance: See Above

Datamance (approved):
LGTM, just see comments about upgrading django to fix the staticfiles path behavior (unless you want to defer it to another ticket in which case just write the ticket and assign it to me)

project/__init__.py
kimberscott: Hmm that "guarantee" seems like it depends on the details of how this gets called which might lead to hard-to-reason-about deployment weirdness (e.g., this runs fine locally but not on Travis build). Proposing to stick with __file__ for now given this isn't broken / isn't a change from the Django boilerplate.

project/settings.py
kimberscott: Made a ticket for it!

PR #553, opened 2020-08-29
Feature/add tests

kimberscott:
Add some tests for web views, fix an existing study edit view test.

I was initially collecting tests to make a bigger PR at some point but it may make more sense esp. if only tests are affected to merge more frequently and keep everything better synced?


Datamance (approved):
LGTM

PR #555, opened 2020-09-01
Allow case-insensitive login

### LINKED ISSUE #554: Make usernames case-insensitive
**TL;DR**: Usernames on Lookit should be case-insensitive.

**Narrative**
"As a Participant, I get confused if I registered under e.g. First_last@gmail.com, but then can't log in using my email address first_last@gmail.com. If I request a password reset, I receive it successfully and can reset the password - but then can't actually log in because I've actually reset the password for First_last@gmail.com."

"As a Participant, if I can't log into my account I may figure I didn't really have one and try creating a new account, which may by chance have different capitalization. This can cause errors and it's also confusing to have multiple accounts unintentionally."

We had about 20 duplicate-up-to-case usernames as of several months ago, and I manually cleaned them up on prod (merging/deleting accounts as appropriate). I've handled tech support for 3 participants over 4-5 days (out of ~300) who were having trouble logging in after resetting their password which turned out to be due to the capitalization of their username.

(Note: this affects researchers too, there are just fewer of us and we seem more likely to be using password managers and all-lowercase email addresses for registration.)

**Acceptance Criteria**

- [x] Manual cleanup of staging, prod databases to avoid duplicate accounts
- [x] Migration that safely changes email addresses for duplicate accounts (e.g. adding +<UNIQUESTRING>) so we can fix any that are added right before 
- [x] Add tests for account creation and username changes(don't allow duplicates), for what happens if duplicate account does exist when logging in, and for logging in as user@company.com and UsEr@ComPany.COM 
- [x] QA on new feature: double check we can log in to previously-existing account originally under all-lowercase; can log in to previously-existing account with capital letters; can not change username to an existing address with capitalization changes; can not register new account using existing address with capitalization changes; can send email to participant; can trigger password reset email, reset password, and log in successfully

**Implementation Notes**
Could have migration actually try to handle merging accounts (e.g., move all children to acct 1, all messages/logs/etc. reference acct 1) and then match children manually afterwards, but I think that'll be substantially more trouble than it's worth to test carefully enough, versus just handling the relatively few cases manually.

Datamance:
Fixes #554

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -4,12 +4,40 @@ from bitfield.forms import BitFieldCheckboxSelectMultiple
 from django import forms
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.forms import PasswordChangeForm as DjangoPasswordChangeForm
-from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.forms import UserCreationForm, UsernameField
 from django.core.exceptions import ValidationError
+from django.forms import EmailField
 from django.utils.translation import gettext_lazy as _
 
 from accounts.backends import two_factor_auth_backend
 from accounts.models import Child, DemographicData, User
+
+
+class ForceLowercaseEmailField(EmailField):
+    def to_python(self, value):
+        return super().to_python(value).lower()
+
+
+class LowercaseUsernameUserCreationForm(UserCreationForm):
+    def clean(self):
+        """Check DB for case-insensitive matches on username.
+
+        TODO: Consider using a migration to render this code obsolete.
+        """
+        cleaned_data = super().clean()
+        username = cleaned_data.get("username")
+        user_model_class = type(self.instance)
+
+        try:
+            user_model_class.objects.get_by_natural_key(username)
+            raise ValidationError("A user with this username already exists!")
+        except user_model_class.DoesNotExist:
+            pass
+
+        return cleaned_data
+
+    class Meta(UserCreationForm.Meta):
+        field_classes = {"username": ForceLowercaseEmailField}
 
 
 class UserForm(forms.ModelForm):
@@ -124,7 +152,7 @@ class TOTPLoginForm(AuthenticationForm):
         return self.cleaned_data
 
 
-class ResearcherRegistrationForm(UserCreationForm):
+class ResearcherRegistrationForm(LowercaseUsernameUserCreationForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Don't autofill passwords, in the interest of security.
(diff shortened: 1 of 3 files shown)


accounts/tests.py
```diff
@@ -250,6 +266,25 @@ def test_lab_researcher_can_create_study(self):
     def test_unaffiliated_researcher_cannot_create_study(self):
         self.assertFalse(self.unaffiliated_researcher.can_create_study())
 
+    def test_create_user_lowercases_username(self):
+        # TODO: Do we actually use `create_user` anywhere?
```
kimberscott, line 270: only in create_superuser - otherwise users are created via the ParticipantSignupForm & ResearcherRegistrationForm. Ok for me to push addition of clean_username methods to those on this branch?

accounts/models.py
```diff
@@ -42,11 +42,16 @@
...
+        return self.get(**{self.model.USERNAME_FIELD + "__iexact": username})
+
     def create_user(self, username, password=None):
         if not username:
             raise ValueError("Users must have a username")
 
-        user = self.model(username=self.normalize_email(username), is_active=True)
+        user = self.model(
+            username=self.normalize_email(username).lower(), is_active=True
+        )
```
kimberscott, line 54: Like you noted below we don't usually use this - to do the equivalent when people create/update accounts, we'd want to handle that in the ParticipantSignupForm, ResearcherRegistrationForm, and AccountUpdateForm

accounts/tests.py
Datamance: Which branch are those currently on?

accounts/tests.py
```diff
@@ -250,6 +266,25 @@ def test_lab_researcher_can_create_study(self):
...
+        # Test that login is case-insensitive
+        response = self.client.post(
+            reverse("login"),
+            {
+                "username": "mixedcase@gmail.com",
+                "password": self.test_password,
+                "next": "/",
+            },
+            follow=True,
+        )
```
kimberscott, line 284: should also check we can log in by posting a not-all-lowercase email

accounts/forms.py
```diff
@@ -179,6 +207,7 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = User
+        field_classes = {"username": ForceLowercaseEmailField}
```
kimberscott, line 210: how come we need this here but not in ResearcherRegistrationForm?

Datamance: Meta is inheriting from `LowercaseUsernameUserCreationForm.Meta` in the other cases

kimberscott (approved):
LGTM! Thanks for handling this so fast and for adding the tests. Fingers crossed we avoid a bunch of confusion for participants here!

Issue #588, opened 2020-11-13
Bad request error on sending email to multiple participants


kimberscott:
**Description**
Sending a message through the Lookit contact form to more than one participant causes a 400 error 

**How to reproduce**
Go to the contact form, e.g. https://lookit-staging.mit.edu/exp/studies/73/contact/, and send a message to >1 participant. You will not see an error and the message will be saved in the DB, but no email will actually be sent (per SendGrid) & the 400 error will be visible on Sentry.

**Expected behavior**
Email should be BCCed to the selected recipients

**Screenshots**

**Environment (OS, browser, branch, etc.):**
Prod, staging (NOT local b/c this is via SendGrid)

**Additional context**

Came up when researcher emailed multiple participants who'd experienced a technical error to offer compensation & invite to try again.

Example error here: https://sentry.io/organizations/massachusetts-institute-of-technology/issues/2028279115/?project=299965&query=is%3Aunresolved

Sendgrid docs suggest we need something in the to field: https://sentry.io/organizations/massachusetts-institute-of-technology/issues/2028279115/?project=299965&query=is%3Aunresolved

PR #589, opened 2020-11-13
quick initial fix for bulk email

kimberscott:
To address #588, would just like to confirm on staging that this is a fix


Datamance (approved):
👍

PR #592, opened 2020-11-14
fix queryset for labs on study edit/create forms

kimberscott:
* Use ...Lab.object.filter... not user.labs.filter... to choose all labs user has perms for, not just those they're in
* Remove User.labs_user_can_create_study_in which was only used twice in these forms, somewhat circuitously, because it doesn't make sense to import Lab in accounts.models
* Add fairly exhaustive tests


Datamance (approved):
The tests are excellent!

PR #593, opened 2020-11-14
give non-researcher participants 4 days from starting a study before …

### LINKED ISSUE #544: Bump session timeout to 2+ days
**TL;DR**: Increase time until user is automatically logged out slightly to accommodate most cases where families leave a tab open to complete a study later.

**Narrative**
"As a participant, I sometimes find a study that's interesting and get as far as the first few frames, but then realize that my child isn't ready right now and leave the browser tab open until a day or two later. Then we would like to participate, but we have been logged out."

**Acceptance Criteria**
- [ ] `SESSION_COOKIE_AGE` is set to 48+ hours, at least for participants; OR session expiration is set to 48+ hours from now upon beginning participation in a study.

**Implementation Notes**
Will complement changes to address https://github.com/lookit/ember-lookit-frameplayer/issues/166 - this will make it less likely, that will do a better job catching any cases and prompting accordingly.

kimberscott:
Fixes #544 - give participants (but not researchers) extra time before sessions expire starting from any time they begin participating in a study.

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -338,14 +338,16 @@ class ExperimentProxyView(LoginRequiredMixin, UserPassesTestMixin, ProxyView):
 
     test_func = user_can_participate
 
-    def dispatch(self, request, *args, **kwargs):
-
+    def setup(self, request, *args, **kwargs):
         # Give participant 4 days from starting any study before session times out, to allow for cases where the
         # parent does instructions etc. but then waits until a convenient time to do the actual test with the
-        # child.
-        if not request.user.is_researcher:
+        # child. This will measure "inactivity" but since we're not saving the session on each request it will
+        # effectively expire after 4 days regardless of activity (renewed each time this view is accessed)
+        if not request.user.is_anonymous and not request.user.is_researcher:
             request.session.set_expiry(60 * 60 * 24 * 4)
+        return super().setup(request, *args, **kwargs)
 
+    def dispatch(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


web/views.py
```diff
@@ -340,6 +340,12 @@ def user_can_participate(self):
 
     def dispatch(self, request, *args, **kwargs):
 
+        # Give participant 4 days from starting any study before session times out, to allow for cases where the
+        # parent does instructions etc. but then waits until a convenient time to do the actual test with the
+        # child.
+        if not request.user.is_researcher:
+            request.session.set_expiry(60 * 60 * 24 * 4)
```
Datamance, line 347: Looks good! Two things:
1) Perhaps would make sense to put this code in [`self.setup`](https://github.com/django/django/blob/master/django/views/generic/base.py#L82) (make sure to `super()`, since `setup` is not really a hook)?
2) Just to double check, this means that every time a participant visits _any_ experiment view, their login expiration gets reset to 4 days in the future. As long as we're cool with that - looks great to me!

Datamance (approved):
Approve with just a small note and potential change (staying out of dispatch)

web/views.py
kimberscott: 1. Moved to setup - thanks! 
2. That's correct - I think I'm ok with that but it does mean that someone could in principle stay logged in indefinitely by participating every <4 days. I guess the concern is that an attacker got access to a session but doesn't actually have login credentials, but especially as this is a participant-only extension I don't *think* we'd be making anything substantially more secure by having them log out and back in?

Issue #576, opened 2020-09-26
Limit number of announcements sent for a given study per day


kimberscott:
**TL;DR**: Don't send more than X announcements for a given study in a day, so we don't "use them up" the first day the study launches.

**Narrative**
"As a Researcher, I may start my study, then realize there is some minor problem that I want to fix. During the time I'm fixing it, participants see my study is not available, and may not try again; because we don't want to bug them again about the same study, I will have "used up" my notifications in that initial launch.

"As a Researcher, if many many participants are notified about my study when it first launches, I may quickly reach the target sample size too quickly."

"As a Participant, I may see an email about a new study and think that would be fun to try this weekend. But when I go to participate in it, it's no longer available. I would rather hear less often but only about studies that are actually available." 

**Acceptance Criteria**
A precise description of how/when this issue will be considered closed, e.g.:

- [x] There is a maximum number of emails about any given study that will be sent when announcement emails are sent out. (Reasonable value is probably around 50-100 at this point.)
- [x] Test for not sending >X emails about given study

**Implementation Notes**

* Setting a hard-coded cap X of emails/study would be fine for now.
* As a crude approximation (because I think this would be easier?), we could just cap the number of total emails sent (as X * number of active public studies). If we're selecting randomly from possible emails this would work a bit differently for studies with more vs. fewer eligible participants. E.g. if we'll only send 50 emails/study/day, study A has 40 eligible participants total, and study B has 3000 eligible participants, then with a simple cap per study we'd email all 40 for study A the first day. With a combined cap we'd email on average 1.2 participants per day about study A (because we'd choose 100 out of 3040 emails, 40 of which are about study A). I think this would be harder to reason about at least.
* Could also make this cap a field on the study with a reasonable default, even if we don't provide a way for researchers to edit it at this point; I could imagine at some point it being useful for researchers to be able to adjust the pace of contacting eligible participants.
* By "per day" I really mean "per email-sending task" (i.e. if we also switch to only sending emails once a week we just want some cap)

PR #595, opened 2020-12-04
limit the number of announcement emails sent about any given study; m…

kimberscott:
…ake sure the ones sent are randomly selected (i.e. not always to the same people first). Implements #576

### DIFF
--- a/studies/tasks.py
+++ b/studies/tasks.py
@@ -8,6 +8,7 @@ import shutil
 import tempfile
 import time
 import zipfile
+from collections import Counter
 from io import StringIO
 from itertools import starmap
 from operator import attrgetter, itemgetter
@@ -94,6 +95,7 @@ FROM message_targets mt
 WHERE lsnfc.latest_sent_time IS NULL
 ORDER BY mt.user_id, mt.child_id, mt.study_id;
 """
+MAX_EMAILS_PER_STUDY = 50
 
 
 class MessageTarget(NamedTuple):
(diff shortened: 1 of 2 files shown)


studies/tasks.py
```diff
@@ -187,12 +189,51 @@ def _segmented_by_study(validated_groups):
...
+    all_user_study_children_tuples = []
+    for user, study_child_mapping in potential_targets_segmented_by_study:
+        study, child_list = random.choice(list(study_child_mapping.items()))
+        all_user_study_children_tuples.append(
+            (user.id, study.id, [child.id for child in child_list])
+        )
+
+    # Randomly select the first <= N study-user pairs for each study. We don't want to just yield the first N per study
+    # because then we'll always invite some families to participate first, others later
+    random.shuffle(all_user_study_children_tuples)
```
Datamance, line 219: What about just doing `potential_targets_as_list = list(potential_targets_segmented_by_study)` to build the list in memory, then `random.shuffle(potential_targets_as_list)` and iterating over that? No need to throw the objects away and re-fetch them from postgres, especially since we're limited to 100 parents from the upstream generator (`_deserialized`)

kimberscott: Ohh, I must have misunderstood - I thought if we iterated over the entire `_deserialized` output we would end up getting through all the potential targets, not 100 at a time - and that this was part of why we'd want to avoid generating such a list. That's what we want, so that we don't always start with the same 100, then the next 100, etc. - shuffling within those potential targets wouldn't really help.

If this would only make a list of 100 users, how come we were able to send substantially more than 100 emails when iterating over the output in `send_announcement_emails`?

Datamance: Sorry, that was sloppy phrasing: a maximum of 100 grouped family-to-study mappings are present in the upstream `_deserialized` generator at any time - furthermore, these groupings are "hydrated" from the results of a batched query to avoid shooting off a bazillion queries. Since it's in a generator pipeline, you still only effectively hydrate just one group at a time before it gets checked in `_validate`. 

The reason for mentioning that is just to say that in building out the full list, there's an upper bound on how many objects are getting allocated on the heap upstream. You'll still get the full list of hydrated targets when you exhaust the generator with `list(potential_targets_segmented_by_study)`

kimberscott: Ohh I see - ok, so building the list of validated targets doesn't involve ever building the whole big list they come from. I'd still be concerned that, say, someday someone wants to run a study that's for ages 0-18 and the list of validated targets is large, or that there are several new studies sending out email across a variety of ages. I.e., if the big list might choke the µwsgi worker, I'd be nervous making assumptions that the smaller list never will.

Datamance: We do use `more_itertools` - if you wanted to get real fancy pantsy you could use the [`sample` function](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.sample), which uses [optimized reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm) under the hood. The tradeoff there is that you're not guaranteeing an explicit # maximum emails per study - just picking a random sample of families and hoping that you're covering all the studies fairly

kimberscott: Yeah, that was my initial hope but the issue is that you end up getting roughly exponential decay in emails sent per study at some point, instead of actually just emailing everyone eligible on a reasonable timescale. This matters a little more for studies with narrow inclusion criteria - e.g. if you want to email everyone with a newborn, or everyone with a kid with ASD - which are "competing" with studies with broader criteria. Ideally we want to actually exhaust all the eligible kids in a simple-to-understand way. So I think we do actually want to go up to a maximum per study rather than sampling families, unless you're concerned about any of the costs involved here.

PR #596, opened 2020-12-07
Fix that "Preview now" was displaying as just "now" because the trans…

kimberscott:
…late tag didn't include quotes


Datamance (approved):
👍

Issue #598, opened 2020-12-14
Consent videos not always marked as consent footage


kimberscott:
**Description**
Occasionally the Video object for a consent video is not marked as `is_consent_footage`, and so doesn't show up in the consent manager, even both the video and data about the consent frame are available.

**How to reproduce**
Create a study with a consent frame. Complete the consent video and wait ~1 min. THEN proceed to next frame. Look for the video in the consent manager - a record of the session is there, but no video. (Or view Videos from admin interface.) 

Compare to completing the consent video and proceeding quickly to the next frame, which creates a video marked correctly as consent footage.

**Expected behavior**
All videos completed during consent frames should be marked as consent footage.

**Environment (OS, browser, branch, etc.):**
staging, prod

**Additional context**
What's happening is that the webhook that processes the videos upon upload to S3 relies on fetching data from the database about the frame to determine if it's a consent frame. So if the frame hasn't been completed by the time the S3 upload completes, we don't know it's a consent frame and the video record is created but not marked as consent footage. 

Two options:
1) Add info about this being consent footage to the Pipe payload string, and have the webhook check for that flag
2) Save consent videos early (upon frame load) so that info is available in the database by the time we go fetch it. Set up the post-save hook to check for a flag that the frame is not actually complete.

(2) is a slightly more convoluted from the frameplayer standpoint, and still technically leaves us with a race condition - it would be more straightforward to bundle this info with the video so it can't get lost.

Both solutions will involve making changes to lookit-api (how to process webhook (1) or post-save hook (2)), then allowing update to ember-lookit-frameplayer.

PR #599, opened 2020-12-14
allow use of consent flag in pipe payload to determine whether video …

kimberscott:
…is consent footage

Addresses #598, in conjunction with https://github.com/lookit/ember-lookit-frameplayer/pull/202 (this should be deployed to prod before having anyone use the ember update)


Datamance (approved):
👍

PR #637, opened 2021-02-17
Extend cloudbuild timeout

jtrouth:
Sets the Cloud Build timeout to 20 minutes. Builds have been timing out under the default 10-minute time limit.


kimberscott:
ah-hah - thanks very much!

PR #639, opened 2021-02-19
Add container cleanup task

jtrouth:
Prevents stopped containers from accumulating and filling up the docker partition.


kimberscott:
Thanks so much! Just replacing with #640 to also add a migration adding the task to the DB

PR #669, opened 2021-04-20
Update OTP form

okaycj:
I have updated the experimenter log in form.  

Screen Shot 2021-04-20 at 4 55 41 PM
Screen Shot 2021-04-20 at 4 56 01 PM
Screen Shot 2021-04-20 at 4 56 22 PM

Open questions/Known issues:
- Once a user has entered their username and password (experimenter or otherwise) they will marked as "logged in".  This means that an experimenter can be in a limbo state were they're logged in but they having't entered their OTP and are not authorized.  
- Is there a programmatic way to figure if a user still needs to enter an OTP?
- Should the link for the OTP help be in the environment file to avoid hardcoding links?


kimberscott:
> Once a user has entered their username and password (experimenter or otherwise) they will marked as "logged in". This means that an experimenter can be in a limbo state were they're logged in but they having't entered their OTP and are not authorized.

Yes, this is slightly odd but by design. The idea is that you only need to authenticate with 2FA to access the experimenter features - not things a participant can also access, like participation in a study or registering a child. This also lets us put the "manage 2FA" functionality on the "manage account" page.

> Is there a programmatic way to figure if a user still needs to enter an OTP?

Yes, but not directly - essentially if they're a researcher & don't already have a 2FA session key. See https://github.com/lookit/lookit-api/blob/master/exp/views/mixins.py#L42

> Should the link for the OTP help be in the environment file to avoid hardcoding links?

Yes. Let's punt on that for now though (there'll be a few other places to change this) - I'll make a [separate issue](https://github.com/lookit/lookit-api/issues/670).

One note about the template change: if this were participant-facing it'd need to be inside a {% trans %} block. But we're deliberately not internationalizing all of the Experimenter app yet, and this text is only ever supposed to be shown to experimenters - so I agree with leaving it untranslated to avoid bogging down the ManyBabies folks with more translation.

kimberscott (approved):
Looks good to me! See responses to questions in separate comment.

PR #671, opened 2021-04-22
[Bug fix/661] Add reply-to email address

okaycj:
I have added reply-to email addresses on lab and study submission emails.

### DIFF
--- a/exp/views/lab.py
+++ b/exp/views/lab.py
@@ -373,7 +373,7 @@ class LabUpdateView(
 
 
 class LabCreateView(
-    ExperimenterLoginRequiredMixin, UserPassesTestMixin, generic.CreateView,
+    ExperimenterLoginRequiredMixin, UserPassesTestMixin, generic.CreateView
 ):
     """
     LabCreateView allows creating a new lab.
@@ -411,6 +411,7 @@ class LabCreateView(
                     name=SiteAdminGroup.LOOKIT_ADMIN.name
                 ).user_set.values_list("username", flat=True)
             ),
+            reply_to=[self.request.user.username],
             **email_context,
         )
 
--- a/studies/helpers.py
+++ b/studies/helpers.py
@@ -15,7 +15,14 @@ logger = logging.getLogger(__name__)
 
 @app.task
 def send_mail(
-    template_name, subject, to_addresses, cc=None, bcc=None, from_email=None, **context
+    template_name,
+    subject,
+    to_addresses,
+    cc=None,
+    bcc=None,
+    from_email=None,
+    reply_to=None,
+    **context,
 ):
     """
     Helper for sending templated email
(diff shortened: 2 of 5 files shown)


studies/models.py
```diff
@@ -512,6 +512,7 @@ def notify_administrators_of_submission(self, ev):
                     name=SiteAdminGroup.LOOKIT_ADMIN.name
                 ).user_set.values_list("username", flat=True)
             ),
+            reply_to=[self.request.user.username],
```
kimberscott, line 515: self in this context is the study, so it doesn't have a request like a view would - it looks like we have the user making the request as `ev.kwargs.get("user")`

exp/views/lab.py
```diff
@@ -411,6 +411,7 @@ def notify_admins_of_lab_submission(self):
                     name=SiteAdminGroup.LOOKIT_ADMIN.name
                 ).user_set.values_list("username", flat=True)
             ),
+            reply_to=[self.request.user.username],
```
kimberscott, line 414: nice job finding both of these, I hadn't realized they were buried in somewhat different places!

studies/helpers.py
```diff
@@ -16,7 +15,14 @@
...
-    template_name, subject, to_addresses, cc=None, bcc=None, from_email=None, **context
+    template_name,
+    subject,
+    to_addresses,
+    cc=None,
+    bcc=None,
+    from_email=None,
+    reply_to=None,
+    **context,
 ):
```
kimberscott, line 26: I *think* Black shouldn't be changing the formatting on this file if we're both using the same configuration from pre-commit. Is it running when you commit changes or are you having to run it manually? In the latter case it's possible it's a different version or with different preferences on line length etc.

kimberscott (changes requested):
Thank you for a "bug fix" that is really just to make my job easier! I hadn't realized when I made this issue that these two emails are generated in quite different ways, so good job tracking both down. 

The study submission email isn't generated from inside a view so just needs to get the user's email differently. 

Let's just sort out what's going on with Black & confirm you can see emails that are generated when running locally before merging.

And hooray for the tests!!

studies/helpers.py
okaycj: See commit https://github.com/lookit/lookit-api/pull/671/commits/4217b36baba62fe967657d1e51a2dfa9cd917d17 for my solution to this issue.

studies/models.py
okaycj: Addressed in commit https://github.com/lookit/lookit-api/pull/671/commits/854fafc171a64f0fc335e93db276031431e6e49e.

studies/models.py
```diff
@@ -512,7 +512,7 @@ def notify_administrators_of_submission(self, ev):
                     name=SiteAdminGroup.LOOKIT_ADMIN.name
                 ).user_set.values_list("username", flat=True)
             ),
-            reply_to=[self.request.user.username],
+            reply_to=[ev.kwargs.get("user")],
```
kimberscott, line 515: this is actually passing a User object rather than just the username (email address) - I think you should just be able to add `.username` here though!

kimberscott (changes requested):

studies/models.py
okaycj: This should be resolved by commit https://github.com/lookit/lookit-api/pull/671/commits/a60970f34d7c3c746a6d2abf5bfc7aa9cd551d43.

kimberscott (approved):
Looks good! I double checked SendGrid respects the reply-to addresses here too.

PR #676, opened 2021-04-29
Update invoke tasks

okaycj:
### Summary

To update our invoke's aging tasks script, this PR will replace brew commands that no longer work and replace pipenv with [poetry](https://python-poetry.org).

### Installation Instructions

(I will create a PR on the docs repo after we agree this is how we want to move forward.)

1. Install [Docker](https://docs.docker.com/docker-for-mac/install/) 
2. Create Postgres db using the following command: 
    ```
    docker run --name lookit-postgres -d -e POSTGRES_HOST_AUTH_METHOD="trust" -e POSTGRES_DB="lookit" -p 5432:5432 postgres:9.6
    ```
3. Install [Brew](https://brew.sh)
4. install [asdf](https://asdf-vm.com/#/core-manage-asdf?id=install)
5. Add Python plugin using the following command.  Here's [documentation](https://github.com/danhper/asdf-python) to help with errors:
    ```
   asdf plugin-add python
   ```
6. Install appropriate version of python using the following command while in the root of the project.  The [asdf Python plugin docs](https://github.com/danhper/asdf-python) can help install older versions of python
   ```
   asdf install
   ```
7. Install [Poetry](https://python-poetry.org/docs/#installation)
8. Before we install Python libs we'll need to install Graphviz.
   ```
   brew install graphviz
   ```
9. Install Python libs.
   ```
   poetry install
   ```
10. Use invoke to run setup.
    ``` 
    poetry run invoke setup
    ```

### Run servers

These two commands will be ran in two terminal windows and celery assumes rabbitmq is running.

```
poetry run invoke server
poetry run invoke celery-service
```


kimberscott (commented):
I realize this is outside the scope of updating the invoke tasks, but while we're switching to poetry, is it possible to update the [Dockerfile](https://github.com/lookit/lookit-api/blob/develop/Dockerfile) so that we actually use poetry during deployment?

kimberscott (changes requested):
This is lovely! It's refreshing to have working setup scripts/instructions :)

Minor requests & questions:
- Can you make sure the postgresql & system_setup task docstrings are still accurate?
- for postgresql - do we want the task to install postgres on Linux? We'd talked about not touching the Linux instructions but now that I see that it feels like we probably want the invoke tasks to nominally do the same thing on different operating systems, to make it at least approachable to update the docs & tasks as it comes up.
- do we still want the docker invoke task, or for that to move over to docs?
- just confirming - the poetry.lock file has dependencies ported over from the Pipfile.lock?

As per separate comment this also seems like the time to make sure we're *using* the poetry.lock file during deployments; otherwise all this careful alignment goes to waste.

Thank you for all your work on this and for taking on this cleanup right away!

okaycj:
> I realize this is outside the scope of updating the invoke tasks, but while we're switching to poetry, is it possible to update the [Dockerfile](https://github.com/lookit/lookit-api/blob/develop/Dockerfile) so that we actually use poetry during deployment?

I made an up date to the Dockerfile in https://github.com/lookit/lookit-api/pull/676/commits/3de986d9fca3e131dedea6c28133aa4a06352c3d.

Dockerfile
```diff
@@ -1,45 +1,28 @@
...
-        libxml2-dev \
-        libxslt1-dev \
-        libffi-dev \
-        python-dev \
-        libpq-dev \
-        graphviz \
-        libgraphviz-dev \
-        pkg-config \
-        curl \
-        gosu \
```
kimberscott, line 20: Why don't we need these anymore?

okaycj: TBH, I don't know.  I spent Friday building and rebuilding this docker image to see if there's anything there that I am missing, but it doesn't appear that we need these libraries.  Once this is deployed to staging we'll see if we need these.

okaycj:
> * just confirming - the poetry.lock file has dependencies ported over from the Pipfile.lock?

Yes, I wrote a script that compared the `python -m pip freeze` from the virtualenv created by the pipenv on develop branch to the poetry virtualenv on this branch.  The output I got was:

```
Package version that don't match across lock files:

        identify 2.2.3 != 2.2.4
        invoke 1.4.1 != 1.5.0
        virtualenv 20.4.3 != 20.4.4

Package version that is NOT in the pipenv.lock file

        appnope 0.1.2

Package version that is NOT in the poetry.lock file

        None
```

Note:  The first version number is from pipenv and the second is from poetry.

okaycj:
> * for postgresql - do we want the task to install postgres on Linux? We'd talked about not touching the Linux instructions but now that I see that it feels like we probably want the invoke tasks to nominally do the same thing on different operating systems, to make it at least approachable to update the docs & tasks as it comes up.
> * do we still want the docker invoke task, or for that to move over to docs?

I removed Postgres install instructions.  I did the same for Ngrok, Rabbitmq, and Docker.

okaycj:
> * Can you make sure the postgresql & system_setup task docstrings are still accurate?

Updates to docstring are handled in the https://github.com/lookit/lookit-api/pull/676/commits/4b36294eef1b258de720f56f1957fb4d392dc907 commit.

tasks.py
```diff
@@ -423,40 +140,7 @@ def postgresql(c, verbose=False):
...
-        else:
-            run('echo "====> Migration failed"')
-    elif PLATFORM == "Darwin":
-        if run("python manage.py migrate", warn=True, hide=not verbose).ok:
-            run('echo "====> Migrated Django models to lookit db"')
-        else:
-            run('echo "====> Migration failed"')
-    else:
-        run("echo {}".format(MESSAGE_WRONG_PLATFORM))
+    c.run("python manage.py migrate", hide="stdout")
```
okaycj, line 101: @kimberscott This is where the migration happens.

kimberscott: yep - can we just have the docstring say it's doing the migration (not installing postgres/creating the database anymore)?

kimberscott (approved):
looks great! here goes...

tasks.py
okaycj: Whoops, missed that.  Good catch!

PR #681, opened 2021-05-11
Add pre-commit install to invoke setup task

okaycj:
Forgot to install the pre-commit hook (again!!!), therefore I've added it to the setup task within invoke.


kimberscott (approved):
LGTM - thanks for adding this!

PR #682, opened 2021-05-14
Add autoflake to pre-commit

okaycj:
Added `autoflake` to the pre-commit config file.  

Note:  It might be a good idea to double check the source code changes and make sure I didn't make any bad assumptions.  Most notably https://github.com/lookit/lookit-api/pull/682/commits/1820f57a3be5f6f2e601d5d94c2971fa3637f79d#diff-3d71d733cff66ce50defe80ea36983c71130313d82907c89ffbfadf0986071fc


exp/tests/test_response_views.py
```diff
@@ -288,7 +288,6 @@ def test_delete_preview_data(self):
         self.assertEqual(
             self.study.responses.filter(is_preview=True).count(), self.n_previews
         )
-        response = self.client.post(url, {})
```
kimberscott, line 291: this shouldn't be removed - the assigned output isn't used, but making the post request is important

exp/tests/test_response_views.py
```diff
@@ -986,7 +985,7 @@ def test_get_appropriate_children_in_child_csv_as_previewer(self):
         content = csv_response.content.decode("utf-8")
         csv_reader = csv.reader(io.StringIO(content), quoting=csv.QUOTE_ALL)
         csv_body = list(csv_reader)
-        csv_headers = csv_body.pop(0)
```
kimberscott, line 989: likewise here there's a side effect of pop() even though we don't use the assigned csv_headers

exp/views/responses.py
```diff
@@ -1955,7 +1955,6 @@ def post(self, request, *args, **kwargs):
         Downloads study video
         """
         match = self.request.GET.get("match", "")
-        orderby = self.get_ordering()
```
kimberscott, line 1958: this can't be deleted because the view's `orderby` is used, even though it's not used *here* in our code

studies/migrations/0066_new-lab-study-perms-data-migration.py
```diff
@@ -178,8 +177,8 @@ def revert_migration(apps, schema_editor):
         apps: instance of django.apps.registry.Apps
         schema_editor: instance of django.db.backends.base.schema.BaseDatabaseSchemaEditor
     """
-    Lab = apps.get_model("studies", "Lab")
-    Study = apps.get_model("studies", "Study")
```
kimberscott, line 182: it looks like we do use these later?

kimberscott (changes requested):
It seems like removing unused variables is causing trouble because sometimes we're assigning a variable but also making a function call with a side effect. Overall that piece doesn't seem as valuable as the unused imports - it's not really finding cases where there's an unused variable and no reason for it. Can Autoflake be configured to just check for unused imports instead?

okaycj:
> Can Autoflake be configured to just check for unused imports instead?

Yes, but I think the "overzealousness" was more me than the `autoflake`.

exp/tests/test_response_views.py
okaycj: Agreed.  That shouldn't have been removed.  It was removed manually by me, My apologies. It should be back in the code base now.

exp/tests/test_response_views.py
okaycj: Yup, I got a bit delete happy there, my apologies.  It should be back now.

exp/views/responses.py
kimberscott: whoops, as discussed this is fine to delete - I thought it was defined on the class rather than in post

studies/migrations/0066_new-lab-study-perms-data-migration.py
kimberscott: Nevermind, not in revert_migration!

kimberscott (changes requested):
I'd go ahead and remove the unused-variable lines where there's no side effect so we're not tilting our heads at those later and wondering if there was some reason for them, but otherwise looks good to go.

okaycj:
@kimberscott I added another commit (https://github.com/lookit/lookit-api/pull/682/commits/bc36f90dbb8fbdf5d6b9c212b618da4c690c98b7) to remove those lines of code.

kimberscott (approved):
Looks great! This is really nice to have set up :)

PR #683, opened 2021-05-18
Added acceptance comment box

### LINKED ISSUE #610: Allow comments on study acceptance as well as rejection
# TL;DR
Allow reviewer to leave comments on study acceptance as well as rejection

# Narrative
As a study reviewer, I often want to leave final comments along with study acceptance, either to offer softer advice (no changes required, but hey this might come up in piloting or you might consider doing X in the future) or just to say something was nicely done. Currently this has to be done via email/Slack, and any advice does not end up part of the study log.

# Acceptance Criteria

- [ ] Comments can be added with study acceptance
- [ ] Comments still can't be added with pausing/starting/etc., just accept/reject/submit. (It actually wouldn't be unreasonable to consider allowing comments on almost everything to support more of a lab notebook - "we paused because we reached N kids and are going to code now to see how many more we need per condition," etc. - but for now let's just make the minimal change.)
- [ ] Comments are added to the approval decision email
- [ ] Appropriate tests added: this will depend on whether we're refactoring the StudyDetailView first per #518 , see below.

# Implementation notes

- The "update status" request is handled by the exp/views/study.py/StudyDetailView view. This currently handles three unrelated form submissions, which makes very little sense - there should in principle be three separate views for adding/removing users, changing study status, and cloning the study. See https://github.com/lookit/lookit-api/issues/518.  Refactoring now could be a good exercise if you'd prefer to do it first (in which case we'll take the other refactor off the onboarding list) but we could also hold off, we're not making anything worse :)
- Whether to show the comments box is just decided in Javascript in studies/templates/studies/study_detail.html/onStateSelect, and then in exp/views/study.py/update_trigger we check if comments-text was sent with the request. So we probably only need to change this in the template to allow approval comments.
- See studies/models.py/Study/notify_submitter_of_approval for function that sends email about approval (will need to modify that email template to include comments, like for rejection emails)
- Tests: If refactoring the StudyDetailView, now is the time to add basic tests of each post view - adding/updating user perms, changing study state, & cloning study. Otherwise, would tentatively propose adding:
   - Test that no study views are accessible if not a researcher (e.g., test_cannot_see_any_study_views_as_participant analogous to exp/tests/test_study_views.py/test_cannot_see_any_study_views_unauthenticated) 
   - Test that *getting* StudyDetailView requires appropriate permissions (i.e., StudyPermission.READ_STUDY_DETAILS or LabPermission.READ_STUDY_DETAILS for the study's associated lab)

okaycj:
Add textbox to study acceptance model. When comments are left, they are displayed in the comments box and logged.

Closes #610

### DIFF
--- a/exp/tests/test_study_views.py
+++ b/exp/tests/test_study_views.py
@@ -26,6 +26,7 @@ from exp.views.study import (
     ChangeStudyStatusView,
     CloneStudyView,
     ManageResearcherPermissionsView,
+    StudyDetailView,
 )
 from studies.models import Lab, Study, StudyType
 from studies.permissions import LabPermission, StudyPermission
@@ -1169,6 +1170,48 @@ class ManageResearcherPermissionsViewTestCase(TestCase):
             manage_researcher_permissions_view.user_only_admin(mock_study, mock_user)
         )
         mock_user.groups.all.assert_called_once_with()
+
+
+class StudyDetailViewTestCase(TestCase):
+    def test_model(self) -> None:
+        study_list_view = StudyDetailView()
+        self.assertIs(study_list_view.model, Study)
+
+    def test_permissions(self) -> None:
+        self.assertTrue(
+            issubclass(StudyDetailView, ResearcherLoginRequiredMixin),
+            "StudyDetailView must have ResearcherLoginRequiredMixin",
+        )
+        self.assertTrue(
+            issubclass(StudyDetailView, UserPassesTestMixin),
+            "StudyDetailView must have UserPassesTestMixin",
+        )
+
+    def test_test_func(self):
+        manage_researcher_permissions_view = StudyDetailView()
+        self.assertEqual(
+            manage_researcher_permissions_view.test_func,
+            manage_researcher_permissions_view.user_can_see_or_edit_study_details,
+            "StudyDetailView.test_func must be set to StudyDetailView.user_can_see_or_edit_study_details",
+        )
+
+    @patch.object(StudyDetailView, "request", create=True)
+    @patch.object(SingleObjectMixin, "get_object")
+    def test_user_can_see_or_edit_study_details(
+        self, mock_get_object: Mock, mock_request: Mock,
+    ) -> None:
+        mock_request.user.has_study_perms.return_value = True
+        mock_is_researcher = PropertyMock(return_value=True)
+        type(mock_request.user).is_researcher = mock_is_researcher
+
+        study_detail_view = StudyDetailView()
+
+        self.assertIs(study_detail_view.user_can_see_or_edit_study_details(), True)
+
+        mock_request.user.has_study_perms.assert_called_once_with(
+            StudyPermission.READ_STUDY_DETAILS, mock_get_object()
+        )
+        mock_is_researcher.assert_called_with()
 
 
 # TODO: StudyCreateView
--- a/exp/tests/test_study_views.py
+++ b/exp/tests/test_study_views.py
@@ -18,6 +18,7 @@ from django.urls import reverse
 from django.views.generic.detail import SingleObjectMixin
 from django_dynamic_fixture import G
 from guardian.shortcuts import assign_perm, get_objects_for_user
+from parameterized import parameterized
 
 from accounts.backends import TWO_FACTOR_AUTH_SESSION_KEY
 from accounts.models import Child, User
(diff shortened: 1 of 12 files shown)


kimberscott:
It looks  like while approval comments are being emailed, they're not actually stored with the study logs or displayed in the "Study logs" section below. 

Screen Shot 2021-05-19 at 10 45 23 AM

Comments also don't show up here when the study is accepted, which would be nice/consistent if not too hard:

Screen Shot 2021-05-19 at 10 48 49 AM

A few places to look since this functionality is a bit of a maze of hooks that magically happen:

* In studies/models.py/Study._log_action we only create a StudyLog object with extra.comments if the event name is submit, resubmit, or reject. So we want to do that for accept as well.
* There's a remove_rejection_comments_after_approved hook in studies/models.py that clears the comments on a Study when it's approved. This would affect being able to show the comments in the second screenshot above, but shouldn't affect the study logs.

In testing this out I discovered that there's a pre-existing bug/way-things-work that comments are lost if you reject a study directly from an "approved" state (possibly from any state besides "submitted"). That is, the StudyLog has `extra: {"comments": ""}` even if comments are entered. I suspect this is due to the remove_rejection_comments_after_approved hook, because that's running any time the Study is saved - including, possibly, upon trying to save the comments on an approved study as it's being rejected. So it may be resolved if we can find a way around clearing comments whenever an approved Study is saved. I think we could probably just save comments no matter what - i.e. if the comments are empty upon approval, still set comments to "" - and that would keep them up-to-date without the odd workaround.

kimberscott (changes requested):
Just requesting we save the approval comments in the StudyLogs - see other comment with screenshots which I didn't realize I could add as part of the review :)

studies/templates/studies/study_detail.html
```diff
@@ -32,17 +32,22 @@
...
+                    break;
+                case 'submit':
                     stateChangeCommentsInput.attr('placeholder', 'Provide information about peer review and any nonstandard elements here');
+                    break;
+                case 'approve':
+                    stateChangeCommentsInput.attr('placeholder', 'List approval comments here');
+                    break;
+                default:
+                    stateChangeCommentsInput.hide();
                 }
```
kimberscott, line 50: this is much nicer, thank you for changing it as opposed to just tacking on another else if :)

exp/views/study.py
```diff
@@ -421,8 +421,15 @@ def get_context_data(self, **kwargs):
...
         context["study_perms"] = self.request.user.perms_for_study(study)
+        context["comments"] = self.comments(study)
         return context
 
+    def comments(self, study: Study):
+        if not study.lab.approved_to_test:
+            return f"It is not possible to submit or start this study until the lab {study.lab.name} is approved to test."
+        else:
+            return study.comments
+
```
kimberscott, line 432: this is a nice refactoring!

exp/tests/test_study_views.py
```diff
@@ -1143,6 +1144,48 @@ def test_only_one_admin_user_other_admins(self):
...
+        study_detail_view = StudyDetailView()
+
+        self.assertIs(study_detail_view.user_can_see_or_edit_study_details(), True)
+
+        mock_request.user.has_study_perms.assert_called_once_with(
+            StudyPermission.READ_STUDY_DETAILS, mock_get_object()
+        )
+        mock_is_researcher.assert_called_with()
+
+
```
kimberscott, line 1188: Very minor but as we establish patterns for test cases - what we really want to ensure, from a security standpoint, is that we never make changes that let someone who's NOT a researcher, or who IS a researcher but who doesn't have READ_STUDY_DETAILS permissions, can't see this page; and that someone who's a researcher with READ_STUDY_DETAILS permissions on this object can. Can we test that explicitly (along the lines of tests at the top of this file for what happens if you're unauthenticated) or is there a reason (test timing, etc.) we shouldn't be doing that?

kimberscott (changes requested):
This looks great - thank you for all the cleanup hidden inside this change! 

One minor request on the tests - it's more a question of what's the best way to ensure the high-level goal of people only being able to access a view if they're supposed to be. I wouldn't usually hold up the PR for that but figure we're currently setting a pattern for how tests will work.

exp/tests/test_study_views.py
okaycj: You nailed it on the head, I'd rather use mocks to quicken up the test suite speed.  What you've request can be done, but I would argue that we can replicate the same behavior from the client by using mocks.  Are you comfortable using mocks to test what would happen if someone attempted to access the view and didn't have the correct permissions?  Or would you rather I use the client?

okaycj: I've added the commit ef7461c to show what I would like to do.

exp/tests/test_study_views.py
```diff
@@ -1171,6 +1173,57 @@ def test_only_one_admin_user_other_admins(self):
...
+    @patch.object(StudyDetailView, "request", create=True)
+    @patch.object(SingleObjectMixin, "get_object")
+    def test_user_can_see_or_edit_study_details(
+        self,
+        has_study_perms: bool,
+        is_researcher: bool,
+        expected: bool,
+        mock_get_object: Mock,
+        mock_request: Mock,
+    ) -> None:
```
kimberscott, line 1211: ! This is really nice.

kimberscott (approved):
Looks great, thanks for figuring out a way to test permissions more thoroughly without using the client! Going forward this looks like a good pattern.

exp/tests/test_study_views.py
kimberscott: Oh, that makes sense, & sounds like a plan - the way the mocks are set up now seems like a good compromise between testing the exact outcome we want and completing the tests in a reasonable amount of time.

PR #684, opened 2021-05-18
Remove size field from video table

### LINKED ISSUE #583: Large video files not stored in database because size doesn't fit in integer field
**Description**
If the "size" of a video as sent to the renamevideo webhook handler is >2147483647, then no record of the video is created in the database because it doesn't fit in an integer field.

**How to reproduce**
Record a large video file (>=2 GB) during an experiment. Observe that although the video is created and renamed (see Pipe logs and corresponding file on S3) no record is created in the database, so it's not visible to a researcher or the participant.

Alternately, mock by calling Video.from_pipe_payload with an appropriate dict (subbing in existing study & response IDs), e.g. 

```
{ 
    "data": {
        "videoName": "hn7hXDK6DzQpdPzUqlrkzB8oe2ZpVhiv"
        "type": "mp4"
        "payload": "videoStream_<study_UUID>_5-test-trial_<response_UUID>_1503330042572_921",
        "size": 2147483648
}
```

**Expected behavior**
If video is successfully recorded and sent to S3, a record of it should be created in the database.

Production

**Additional context**

See https://sentry.io/share/issue/b5bbd36c219d46a79efa61d071350b09/ for the actual error that alerted me to this. That particular instance I fixed manually by creating a Video record based on the Pipe log for the original filename included in the webhook request as captured by Sentry.

It would be fine to just cap "size" at the max integer. Alternately, since we don't actually USE the size field, we could remove it entirely. It seems potentially useful in the future, but on the same timescale as getting off Pipe, at which point we'd be substantially changing how we obtain it anyway.

okaycj:
A migration to remove the size field from the studies_video table. 

Closes #583


kimberscott (changes requested):
Looks great! We just need to check for usage of "size" in any places where we actually create a Video object. It looks like probably the only place is in Video.from_pipe_payload.

kimberscott (approved):
Looks great! Will be nice to have this edge-case bug resolved so big videos don't just go missing :)

PR #690, opened 2021-05-20
Fix bug when changing permissions ui

okaycj:
When a permission was changed and the user was the only admin the view would throw an assertion error.  This error was caught in the post method of the `StudyDetailView`.  I removed the assert in a recent issue, this caused the last admin on a project to have their permission group changed without error.  

To fix this issue, I changed it from a try/catch to a boolean response.  If there's an error on the `manage_researcher_permissions` method then it should return False and then a 403 error will be given.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ jobs:
     - stage: Build & Test
       script:
         - bash -c 'until pg_isready; do sleep 1; done'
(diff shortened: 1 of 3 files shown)


kimberscott (approved):
This is really nicely done - thanks for digging into what was going on here, and getting the tests in line with the updated setup. I would also not have expected this bug to be so involved!

PR #697, opened 2021-05-24
Update message when video doesn't load

### LINKED ISSUE #262: Consent videos time out after a minute, then look as if there's a problem with the video
**Description**
When using the consent manager on the experimenter interface, after a minute the consent video links time out and the video appears as if there was a problem with recording. So you can effectively only use the consent manager for a minute at a time between reloading.

**How to reproduce**
Go to the consent manager for a study with some responses. Play a consent video ('A'). Wait a minute. Select a different consent video, then 'A' again. You should see a black video and 'Uh oh, we can't load the video.'

**Expected behavior**
The video should either play (by fetching a new link if needed) or say the link has timed out, with an option to refresh/reload. 

**Screenshots**
Screen Shot 2019-04-01 at 4 57 34 PM

On staging as of b336d662719b8b367910ad99b9a9e9222f086e6a

**Additional context**

okaycj:
Update error message when video doesn't load. 

Closes #262

### DIFF
--- a/studies/templates/studies/study_responses_consent_ruling.html
+++ b/studies/templates/studies/study_responses_consent_ruling.html
@@ -284,7 +284,7 @@
 
             $videoSource.on("error", function (event) {
                 if ($videoSource.attr("src").length) {
-                    $currentVideoInfo.addClass("bg-danger").text("Uh oh, we can't load the video.");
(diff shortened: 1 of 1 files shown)


kimberscott:
Yes, this change should definitely "fix" the issue, it's really just changing this string - but worth trying out the whole pipeline just to get familiar with how everything works.

kimberscott (approved):

PR #698, opened 2021-05-24
Google Tag Manager

okaycj:
This PR adds the correct Google Tag Manager code to our HTML templates.  

Open Question:

There are two bits of code.  The first is the normal javascript variant, while the second is a `noscript` version that will trigger when some is running without JS.  Should we leave out the `noscript`?


okaycj:
Closing this issue in favor of #699.  I misspelled the branch name.

PR #699, opened 2021-05-24
Google Tag Manager

okaycj:
This PR adds the correct Google Tag Manager code to our HTML templates.


kimberscott (changes requested):
Yeah, let's leave out the noscript unless you think it's necessary - it seems a bit obnoxious to try to insert an iframe if someone is actually trying to run without JS (although maybe my prior on why someone would do that is skewed to privacy/free-software rather than tech limitations?) and given that participation is fully dependent on Javascript I'm fine with just losing that data.

kimberscott (approved):

PR #707, opened 2021-05-28
Update experiment form help text.

### LINKED ISSUE #125: Specify whether participant can complete study multiple times
*Pain point*: When a child has already participated in a study at least once, the parent sees a standard warning upon participating again, even if it is designed to be a longitudinal study with multiple sessions.

*Acceptance criteria*: Researchers can specify whether participants should see a warning when participating multiple times.

*Implementation notes/Suggestions*: Eventually we want to support particular schedules for longitudinal designs, this is a first step that will cover the immediate problem.

okaycj:
We found out that there is no warning on running a child more than once on an experiment.  

Additionally, update the location of the help text on the experiment edit form. 

Closes #125

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -262,8 +262,8 @@ class StudyForm(ModelForm):
             "exit_url": "Exit URL",
             "criteria": "Participant Eligibility Description",
             "contact_info": "Researcher Contact Information",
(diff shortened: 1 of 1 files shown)


kimberscott (approved):
This is lovely - thank you! I'm sorry for the wild goose chase on the warning message though.

PR #711, opened 2021-06-02
Keeping packages up to date

okaycj:
Update python libraries were available and not pinned.


kimberscott (approved):
Sounds good to me since tests pass! Thanks for keeping on top of this.

Issue #121, opened 2018-10-02
Sort/display available studies by child's eligibility


kimberscott:
*Pain Point*: Available studies are currently displayed to families in a grid in a fixed order; with either complex inclusion criteria or more than a few studies, it is difficult for a parent to tell which studies are appropriate for their own children. 

*High-level acceptance criteria*: A parent can see a list of studies appropriate for a given child without reading through individual study criteria. Parents and users not logged in can still browse all "discoverable" studies. Parents can view a list of studies by child under "past studies."

*Tasks*:

Card design:
- [x] Reconsider current "card" design and agree on sketch to implement. Will want to at minimum make sure thumbnail images are vertically aligned for aesthetics, but may want to consider a full-width card that has the thumbnail to left + info to right for ease of looking through a bunch of results. Leave a space for "No children eligible" / "Jane, Bob" text. Include lab info prominently - e.g. run by the XYZ lab at Institute. Don't include the redundant "see details" link per https://github.mit.edu/Accessibility-FY2021/lookit/issues/17
- [x] Implement new card layout with existing "current studies" page.

Finding studies (Participant-facing study list view):
- [x] Sketch of search/filtering options top or side bar as described below
- [x] Add top or side section with "Find studies for... [dropdown]", "only show studies we haven't done" checkbox, search button, and clear button.
- [x] If user is logged in, show "only show studies we haven't done" checkbox, otherwise don't. If user is logged in, populate dropdown with user's registered (non-deleted) children. 
- [x] If user is not logged in, populate dropdown with babies (under 1), toddlers (1-2), preschoolers (3-4), school-age kids(5-18), adults (18+), and add a "Log in to find studies just right for your child!" prompt
- [x] If user is logged in, display on each study card which children are eligible (e.g. "No children eligible" / "Bob is eligible" "Bob and Jane are eligible" / "Bob, Jane, and Sally are eligible")
- [x] Implement "search" functionality for checking eligibility of specific children when user is logged in - show only studies the selected child is eligible for (using criteria expression as well as age range - we have a utility for this already). 
- [x] Implement "search" functionality for looking for studies for each broad age group when user is not logged in. In this case, rather than checking eligibility of a child, we just want to check whether there's any overlap between the study's min/max age range and the descriptor (baby, toddler, etc.). I.e. study min age <= descriptor max age and study max age >= descriptor min age.
[...]

PR #722, opened 2021-06-09
Update Study list view

okaycj:
Update study list view.

In addition to #121:
- Removed static files that are no longer needed.
- Updated the two base html files to now use the bootstrap template tags

out


web/views.py
```diff
@@ -211,7 +213,16 @@ class StudiesListView(generic.ListView):
...
+
+        qs = super().get_queryset().filter(state="active", public=True)
+        user = self.request.user
+
+        if user.is_anonymous:
+            sort_fn = lambda s: sha1(s.uuid.bytes).hexdigest()
+        else:
+            sort_fn = lambda s: sha1(user.uuid.bytes + s.uuid.bytes).hexdigest()
+
+        return sorted(qs, key=sort_fn)
```
kimberscott, line 225: this is lovely!

kimberscott (changes requested):
Minor questions only throughout. This looks great overall, I'm excited to actually see it live on staging. The new sorting mechanism is a very elegant solution!

web/static/base.css
```diff
@@ -242,87 +242,27 @@ button.account-row {
...
 }
-
-.study-feedback-description-wrapper {
-    display: inline-block;
-    height: 150px;
-    width: 800px;
+.study-list-panel h3 {
+    font-size: large; 
+    font-weight: normal;
 }
```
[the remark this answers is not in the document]
okaycj: Nice catch!  I'll put those back.  Thank you.

web/templates/web/studies-list.html
```diff
@@ -4,96 +4,105 @@
...
-      {% endif %}
-  </div>
-  {% endif %}
+{% if next %}
+<div class="alert alert-danger" role="alert">
+    {% if user.is_authenticated %}
+    <p>
+        {% blocktrans %}
+        Your account doesn't have access to this page. To proceed, please login with an account that has access.
+        {% endblocktrans %}
```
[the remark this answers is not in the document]
okaycj: I can put it back.  For my IDE it was line wrapping those longer template tags.  This was an easy fix.

pyproject.toml
```diff
@@ -51,6 +50,7 @@ gevent = "20.9.0"
 greenlet = "0.4.17"
 psycogreen = "1.0.2"
 sendgrid-django = "4.2.0"
+django-bootstrap3 = "^15.0.0"
```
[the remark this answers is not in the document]
kimberscott: manually is fine!

accounts/forms.py
```diff
@@ -440,3 +440,27 @@ class Meta:
...
         }
+
+
+class StudyListSearchForm(forms.Form):
+    choices = [
+        ("", "Find studies for..."),
+        ("0,1", "babies (under 1)"),
+        ("1,2", "toddlers (1-2)"),
+        ("3,4", "preschoolers (3-4)"),
+        ("5,18", "school-age kids(5-18)"),
```
kimberscott, line 451: space before `(` and 5-17 for consistency

web/views.py
```diff
@@ -200,18 +207,100 @@ def form_valid(self, form):
...
+                    s for s in studies if age_range_eligibility_for_study(age_range, s)
+                ]
+            else:
+                # when user is authenticated, children value is a child pk
+                child = Child.objects.get(pk=children_value)
+
+                if not show_experiments_already_done_value:
+                    studies = self.completed_consent_frame(studies, child)
+
+            studies = [s for s in studies if get_child_eligibility_for_study(child, s)]
```
kimberscott, line 257: I think we don't want to do this filtering if the user is anonymous, right? only if authenticated?

web/views.py
```diff
@@ -200,18 +207,100 @@ def form_valid(self, form):
...
+        user = self.request.user
+
+        if user.is_authenticated:
+            children = user.children.filter(deleted=False)
+
+            # add eligible children to study object
+            for study in studies:
+                study.eligible_children = create_string_listing_children(
+                    [c for c in children if get_child_eligibility_for_study(c, study)]
+                )
```
kimberscott, line 288: that's cool - how does it work to add stuff to the study that isn't part of what it'd have per the database? we can just add arbitrary fields?

web/views.py
```diff
@@ -200,18 +207,100 @@ def form_valid(self, form):
...
+        if children_value:
+            if user.is_anonymous:
+                # when user is anonymous, children value is an age range.
+                age_range = [int(c) for c in children_value.split(",")]
+                studies = [
+                    s for s in studies if age_range_eligibility_for_study(age_range, s)
+                ]
+            else:
+                # when user is authenticated, children value is a child pk
+                child = Child.objects.get(pk=children_value)
```
kimberscott, line 252: should check somewhere that this child pk is actually associated with the logged-in account. Otherwise you could send sequential pks and determine eligibility for any arbitrary kid... probably not a security issue on its own, but in combination with any similar unintended abilities it could be (e.g. getting a list of nicknames and approximate birthdates of kids who definitely have autism since they're eligible for such-and-such study)

web/templates/web/studies-list.html
```diff
@@ -4,96 +4,125 @@
...
+                    <div class="text-center">
+                        <a class="btn {% if study.eligible_children %}btn-success{% endif %} btn-default text-center" href="{% url 'web:study-detail' uuid=study.uuid %}" role="button">
+                            {% trans "See details" %}
+                        </a>
                     </div>
                 </div>
             </div>
+            <div class="col-md-9 right">
+                <h1>{{study.name}}</h1>
+                <h2>{{study.lab.name}}</h2>
```
kimberscott, line 77: do we have space here to add (Institution)? A lot of lab names are very similar and the institution is an easier way to distinguish

accounts/queries.py
```diff
@@ -88,6 +88,14 @@
 QUERY_DSL_PARSER = Lark(QUERY_GRAMMAR, parser="earley")
 
 
+def age_range_eligibility_for_study(child_age_range, study) -> bool:
+    study_start, study_end = study_age_range(study)
+    child_start = child_age_range[0] * 365
+    child_end = child_age_range[-1] * 365
+
+    return study_start <= child_end and study_end >= child_start
```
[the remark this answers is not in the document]
okaycj: Me too!

studies/migrations/0071_auto_20210629_1434.py
```diff
@@ -0,0 +1,25 @@
...
+                "base_manager_name": "related_manager",
+                "permissions": (
+                    (
+                        "view_all_response_data_in_analytics",
+                        "View all response data in analytics",
+                    ),
+                ),
+            },
+        ),
+    ]
```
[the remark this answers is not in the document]
okaycj: Before I get too far into this I have to admit I still need to do some research on this.  Basically, there are existing migrations to add `ordering` in the meta class using the `AlterModelOptions` class.  This can be seen [here](https://github.com/lookit/lookit-api/blob/e390e5fc536b9d0a47a976184a469791a3e51f25/studies/migrations/0017_auto_20170710_1449.py#L15-L21) and in other migrations.  What I don't know is "why?".  

To answer your other two questions, yes and yes.

web/views.py
okaycj: Python supports monkey patching.  As long as we don't override an existing value (that we might need) this should work well for us.

web/templates/web/_navigation.html
```diff
@@ -10,47 +11,44 @@
...
     <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
       <ul class="nav navbar-nav navbar-left">
-        <li><a href="{% url 'web:studies-list' %}">{% trans "Studies" %}</a></li>
-        <li><a href="{% url 'web:faq' %}">{% trans "FAQ" %}</a></li>
-        <li><a href="{% url 'web:scientists' %}">{% trans "The Scientists" %}</a></li>
-        <li><a href="{% url 'web:resources' %}">{% trans "Resources" %}</a></li>
+        {% nav_item request 'web:studies-list' 'Studies' %}
+        {% nav_item request 'web:faq' 'FAQ' %}
+        {% nav_item request 'web:scientists' 'The Scientists' %}
+        {% nav_item request 'web:resources' 'Resources' %}
```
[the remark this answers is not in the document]
okaycj: The translation should be handled in `nav_item()` within `web/templatetags/web_extras.py`.  Unless I am mistaken, the underscore should do the translation and then pass that back to the template.

web/views.py
```diff
@@ -200,18 +207,100 @@ def form_valid(self, form):
...
+                    [c for c in children if get_child_eligibility_for_study(c, study)]
+                )
+
+    def search_form(self):
+        if self.request.POST:
+            form = self.form_class(self.request.POST, user=self.request.user)
+            if form.is_valid():
+                return form
+
+        return self.form_class(user=self.request.user)
```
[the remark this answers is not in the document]
okaycj: I did some digging around and I found [FormMixin](https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-editing/#formmixin).  This might be a better pattern to follow and it looks like it's in other views.  Should I take the time to do this refactor?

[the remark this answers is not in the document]
okaycj: Upon reflection, I think I'd like to do this refactor, if you're willing.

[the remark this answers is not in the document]
kimberscott: Sounds good to me if you're up for it! I wouldn't count on my feeling of "this is odd" - a lot of Django seems odd to me still - but if there's a pattern that makes more sense to you too then let's go for it. Sorry for the delay.

studies/migrations/0071_auto_20210629_1434.py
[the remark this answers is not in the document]
kimberscott: ah-hah, I see the absence of "ordering" under options is actually what's removing it here. I was expecting to see the removed thing named like if we were removing a field. Sounds good!

web/templates/web/_navigation.html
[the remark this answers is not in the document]
kimberscott: Oh, I missed the underscore - makes sense! We might want to just confirm that when we run `makemessages` any changes make sense to us (e.g. just adding things we actually added to translate)

web/views.py
kimberscott: this one still needs to be addressed, right? line 257 should be indented to only happen when user is authenticated?

web/views.py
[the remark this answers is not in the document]
okaycj: I wasn't aware of these mixins.  This is a good opportunity to dig in and learn one of them.

web/views.py
kimberscott: Nevermind, I'm sorry! I got confused by GitHub and was looking at an outdated version.

kimberscott (commented):
Everything I mentioned before has been addressed - looks great! (...and thanks for your patience with code review covering grammar.) I'm just holding off on officially "approving" since I know you're still working on the refactor/tests.

kimberscott:
Notes from (local environment) QA:

* "We are not running any studies at this time" - text should be different if this is the result of a search (e.g. "No studies found"). We could actually change it for all cases if that's simpler.

* On narrower windows the layout is awkward - can we at least set a max size for the thumbnail?
  Screen Shot 2021-07-01 at 11 51 32 AM

* The "show experiments already done" checkbox only applies if a child is selected - can we disable that field if no child is selected? 

* I played around with the layout a bit in a separate branch and have minor tweaks to see if you like them - see PR to this branch.

exp/templates/exp/_navigation.html
```diff
@@ -7,48 +9,48 @@
...
+            <li><a href="{% url 'logout' %}">Logout</a></li>
+          </ul>
+        </li>
         {% else %} {# /if request.user.is_authenticated #}
-            {% if 'login' not in request.path %}
-                <li><a href="{% url 'login' %}?next={{ request.get_full_path.urlencode }}">Login</a></li>
-            {% endif %}
+        {% if 'login' not in request.path %}
+        <li><a href="{% url 'login' %}?next={{ request.get_full_path.urlencode }}">Login</a></li>
+        {% endif %}
```
kimberscott, line 53: why is the formatting now doing this? (don't go too deep down this rabbit hole, just curious if there's an obvious answer)

web/static/base.css
```diff
@@ -38,10 +38,6 @@ body {
     padding: 1em 0;
 
 }
-.lookit-page-title {
-    background: #337ab7;
-    color: #FFF;
-}
```
kimberscott, line 44: thank you for being so thorough as to get rid of css we're not using :)

web/templates/web/_navigation.html
```diff
@@ -10,47 +11,44 @@
...
+
       <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-2">
         <ul class="nav navbar-nav navbar-right">
           {% if user.is_authenticated %}
             {% if user.is_researcher %}
-              <li><a href="{% url 'accounts:manage-account' %}">{% trans "My Account" %}</a></li>
+            {% nav_item request 'accounts:manage-account' 'My Account' %}
             {% else %}
-              <li><a href="{% url 'accounts:manage-account' %}">{% trans "Hi" %} {{ user.nickname }}</a></li>
+            {% nav_item request 'accounts:manage-account' 'Hi '|add:user.nickname %}
```
kimberscott, line 37: is this translated?

web/views.py
```diff
@@ -200,18 +208,114 @@ def form_valid(self, form):
...
+
+    def completed_consent_frame(self, studies, child):
+        return [
+            r.study
+            for r in Response.objects.filter(
+                study__in=studies, child=child, completed_consent_frame=True,
+            ).distinct("study_id")
+        ]
+
+    def child_eligibility(self, studies):
```
kimberscott, line 301: very minor nitpick - can we call this set_eligible_children_per_study or some similar (better) name that describes the side effect here?

web/tests.py
```diff
@@ -442,6 +454,264 @@ def test_get_success_url_not_deleted_children(self, mock_request):
...
+
+        mock_user_b = MagicMock(name="b")
+        type(mock_user_b).uuid = PropertyMock(
+            return_value=uuid.UUID("{33333333-1234-5678-1234-567812345678}")
+        )
+
+        mock_user_c = MagicMock(name="c")
+        type(mock_user_c).uuid = PropertyMock(
+            return_value=uuid.UUID("{22222222-1234-5678-1234-567812345678}")
+        )
```
kimberscott, line 705: can you add a comment indicating what this is checking for?

web/views.py
```diff
@@ -200,18 +208,114 @@ def form_valid(self, form):
...
+        if child_value:
+            if user.is_anonymous:
+                # when user is anonymous, child value is an age range.
+                age_range = [int(c) for c in child_value.split(",")]
+                studies = [
+                    s for s in studies if age_range_eligibility_for_study(age_range, s)
+                ]
+            else:
+                # when user is authenticated, child value is a child pk
+                child = Child.objects.get(pk=child_value, user=user)
```
kimberscott, line 260: this is nice, so far as I can tell this completely takes care of the "pk of a child not associated with this user sent" security issue?

kimberscott (changes requested):
very very minor changes requested only! This looks great overall and I'm very excited for it. Do you know what's going on with the Travis builds?

kimberscott:
Per discussion - let's change the "show experiments already done" to "hide studies we've done"

web/views.py
okaycj: It does.  The child object returned must have both the pk of the child and the user object.  The user object is obtained from the the user be authenticated.

exp/templates/exp/_navigation.html
okaycj: I don't know why the default formatter for vscode does anything.  But to me, it looks like it just removed the whitespace in front of those lines.  I can put the whitespace back if prefer.  

As a side note, I wish I could find a reasonable universal Django template format software.

web/templates/web/_navigation.html
okaycj: No.  I'll need to take some time to look into this.

exp/templates/exp/_navigation.html
kimberscott: no worries, just curious!

web/templates/web/_navigation.html
okaycj: Per our conversation, replace "Hi..." with "My Account"

PR #726, opened 2021-06-15
Update dockerfile for running tests

okaycj:
I named the dockerfile for running unit tests `travis.Dockerfile` as it was intended for travis-ci.  I have now learned that we're running these tests in GCP and the name doesn't make sense.


kimberscott (approved):
LGTM! Sorry for the delay.

PR #727, opened 2021-06-29
Stopgap solution to staging deploy 

okaycj:
This is a stopgap solution to getting the studies list feature onto staging.


kimberscott (approved):
looks good to me as a stopgap! :)

PR #728, opened 2021-06-29
Stopgap staging deploy

okaycj:
The same stopgap solution, but with graphviz libraries.


kimberscott (approved):
looks even better :) sorry not to be helpful in catching things!

PR #730, opened 2021-07-01
Participant/redesign studies list view ks review

kimberscott:
Minor suggestions for redesign of studies list view that were easier to write out than put in comments.


web/templates/web/studies-list.html
```diff
@@ -56,27 +56,32 @@
...
+                <h1>{{study.name}}</h1>
+                <p><em>{{study.lab.name}} ({{study.lab.institution}})</em></p>
+            </div>
+            <div class="col-sm-2 col-xs-12">
+                <div class="text-center">
+                    <a class="btn {% if study.eligible_children %}btn-success{% endif %} btn-default text-center" href="{% url 'web:study-detail' uuid=study.uuid %}" role="button">
+                        {% trans "See details" %}
+                    </a>
+                </div>
+            </div>
```
okaycj, line 70: This might be a bit too knick picky but, the "see details" button is a bit tight on the bottom when the window is very small.  If this is intended, then I can get over it. 

Screen Shot 2021-07-01 at 1 53 33 PM

okaycj (changes requested):
This looks good to me!  I have one comment, but super low concerns with it.

web/templates/web/studies-list.html
kimberscott: Thanks for catching that! Here's what it looks like now:

Screen Shot 2021-07-01 at 3 25 46 PM

okaycj: I like it.

okaycj:
Sorry for the accidental branch commit.  Everything should be back to the way it was.

okaycj (approved):
LGTM!  💯

PR #731, opened 2021-07-08
Install dev dependencies only for tests

okaycj:
Here is my latest attempt at getting two docker images:

- One image for running tests
- A second image for for pushing to GCR and running in production.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -50,8 +50,24 @@ steps:
     '.'
   ]
 
+- id: build-image
+  name: gcr.io/cloud-builders/docker
+  args: [
+    'build',
+    '--tag',
+    'lookit-test',
+    '--build-arg',
+    'GIT_COMMIT=$COMMIT_SHA',
+    '--build-arg',
+    'GIT_TAG=$TAG_NAME',
+    '--network=cloudbuild',
+    '--file',
+    'dev.Dockerfile'
+    '.'
+  ]
+
 - id: run-tests
-  name: lookit
+  name: lookit-test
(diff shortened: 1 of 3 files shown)


cloudbuild.yaml
```diff
@@ -82,6 +82,8 @@ steps:
     'BRANCH_NAME=$BRANCH_NAME',
     '--env',
     'TAG_NAME=$TAG_NAME',
+    '--env',
+    'POETRY_INSTALL_ARG="--no-dev"'
```
jtrouth, line 86: This argument needs to go in the 'build-image' step that starts on line 39 rather than this deployment step. You'll want to pass it with the --build-arg flag instead of --env.

jtrouth (changes requested):
The Dockerfile looks good, but changes are needed to cloudbuild.yaml.

kimberscott (approved):
I'm "approving" to defer entirely to @jtrouth on this one :)

okaycj:
@jtrouth Hey, I am just checking with you and letting you know this is still on my radar.

jtrouth:
@okaycj This should be good as soon as the change I requested to the cloudbuild.yaml is made. Since the poetry install argument is used in the container build process it needs to be passed in the container build step rather than the deployment step.

jtrouth:
@okaycj Did you push your changes up? I'm not seeing a new commit.

okaycj:
> @okaycj Did you push your changes up? I'm not seeing a new commit.

You will see the changes in commit 9119db5.

jtrouth (approved):

PR #733, opened 2021-07-08
Update travis ci config

### LINKED ISSUE #729: Does Travis CI need to build a docker image?
# TL;DR
Pulling images from docker hub might be taking too much time and wasting Travis CI credits.

# Narrative
Currently, Travis CI builds a docker image and installs our app in it.  The process of pulling down the python image is time consuming and wasting our gifted credits on travis-ci.com.  Additionally, docker hub is limiting downloads for logged in users, causing false fail reporting.  This would effect the Travis CI badges we have in our readme.  

Could we not build the image?

Alternatively, we could just install the necessary python packages in the travis ci provided python environment and run the test suite.  We could also add call to `pre-commit` to verify that those checks are passing as well.

okaycj:
Remove building and running tests/linting within a docker container in travis-ci.  Instead, uses the native environment that is provided.  

Closes #729


kimberscott (changes requested):
If we're leaving a commented-out config file, can we just also leave one line of explanation for why we're moving to the simpler non-docker setup for Travis? Otherwise looks great! (Just merge, don't wait on official signoff :) )

okaycj:
> If we're leaving a commented-out config file, can we just also leave one line of explanation for why we're moving to the simpler non-docker setup for Travis? Otherwise looks great! (Just merge, don't wait on official signoff :) )

Whoops,  I'll remove that and merge it in.

PR #735, opened 2021-07-08
Install without using poetry

### LINKED ISSUE #689: Use pip install requirements instead of poetry
**TL;DR**: Install our project within the Dockerfile by using `pip install .`.

**Narrative**
Currently we have to install poetry to install our project's libraries.  In order for poetry to be installed it has to install its own python libraries.  These libraries might (unlikely) conflict with our own library requirements.  Additionally, poetry is intended as a development tool, not a deployment tool.  

To avoid any possible issues, we should be deploying our project with `pip install .`, which would require a reorganization of our directory structure.  

**Acceptance Criteria**

- [ ] Tests
- [ ] Scoping
- [ ] Alternatives Considered
- [ ] Rico sign-off

**Implementation Notes**
This is possible, but messy.

okaycj:
This is an attempt to be able to install our application without poetry.  

Todos:

1. Test that this application can be ran by uwsgi when installed into virtualenv.
2. Is it possible to install using `pip install .` with Poetry not installed?

Closes #689

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,18 @@
 name = "lookit-api"
 version = "0.1.0"
 description = ""
-authors = [""]
+authors = ["MIT <mit@mit.com>"]
+
+packages = [
+    { include = "accounts" },
+    { include = "api" },
+    { include = "exp" },
+    { include = "project" },
+    { include = "studies" },
+    { include = "web" },
+]
+
+include = ['build/*', "ember_build/**/*", "locale/**/*"]
 
 [tool.poetry.dependencies]
 python = "3.8.3"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -2,7 +2,18 @@
 name = "lookit-api"
 version = "0.1.0"
 description = ""
-authors = [""]
+authors = ["MIT <mit@mit.com>"]
```
kimberscott, line 5: Can this just be MIT? (mit.com isn't a domain that makes sense to use as a placeholder - we could always use lookit@mit.edu if we want a durable-ish reference that actually goes to us)

okaycj: I meant for it to be "...mit.edu".  I'll update it to what you suggested above.

okaycj:
With the recent changes to how we deploy using poetry and the complexity this way of deploying would introduce, this change doesn't seem like something we should be pushing forward.  Closing issue and PR.

PR #746, opened 2021-07-12
External study type support

### LINKED ISSUE #626: External study type support
# TL;DR
Add a new "external" type of study which is not hosted on Lookit

# Narrative
As a researcher, I may want to run studies on another platform, including synchronous Zoom-based studies, but recruit participants through Lookit. Running through Lookit rather than only posting on e.g. CHS also gives me the ability to independently manage when my study is collecting data, and (pending #121) will allow participants to more easily find my study if they're eligible. 

# To do - Lookit director
* [x] Provide copy for "going from Lookit to an external site" - indicating to participant that we now have less control over content, & that child/account data will be shared with that research team. (CJ can show you where this is located)
* [x] Provide copy for "Past Studies" page to indicate whether response is confirmed
* [x] Revisions to Terms of Use and/or Access Agreement to reflect external study option. (The Terms should be updated as a flat page, the changes committed to the research-resources repo, and the community notified as described in the onboarding document. The agreement can just be replaced in research-resources and people signing new agreements will use the new one - but it may not be necessary to make changes, which would be nice in terms of existing agreements clearly supporting external studies.) See onboarding doc for details about changing terms of use. Also change terms of use quiz.

# Overall setup notes

When a participant clicks to participate in an external study, in addition to sending them to the site, Lookit should create a minimal response reflecting their interest in participating. Researchers should be able to see basic info on anyone with a response to an external study, without requiring that the participant has gotten through a consent frame or consent has been confirmed - in contrast to studies run on Lookit. Researchers should be able to confirm whether a particular response is actually associated with a record from their external system and leave feedback that shows up on Lookit, as for a Lookit response. 

It doesn't need to be possible to "participate" in the same external study more than once with the same child - Lookit can track "participating at all" and the external site can handle any longitudinal aspects.

Initial plan for confirmed field: _We discussed the mechanism for marking responses as "confirmed" - it would be nice, for participant experience especially, to distinguish between studies someone participated in and studies someone just clicked on to look into. But even if we withheld access to data download until confirmation, some researchers wouldn't do it - they'd just want to use Lookit for recruitment and collect all the data they need externally. We landed on letting researchers optionally mark confirmation but making the default "confirmed" - so they can say no, this person didn't actually do the study, and we could take that into account for e.g. study announcements and display._
Update 8/19: We don't actually know that this is adding value for researchers or participants, or that anyone will go to the extra trouble of un-checking confirmed on Lookit. Holding off on this until there is a clear use case or other data being integrated from outside studies - so no confirmed field on responses, no interface for editing it, no filtering on confirmed.

In general, we want to check whether a *study* is external, not check responses individually; this may affect both how to display things on a view and which responses researchers should have access to. (We don't want to do a lot of mixing and matching here - checking responses some places and study types others - which will be a bit brittle.) In some cases (e.g. message participants) there will be minimal or no changes to views for external studies. In others  there will be only very minor changes and we can deal with if/else logic (e.g., researcher-facing study detail view - don't show the "build experiment runner" button for external studies) as a separate view would be overkill. But where we get into major changes we should go ahead and have separate views for internal and external studies, especially to make this more maintainable if there is eventually another experiment type. 

# Acceptance Criteria

### Study type table

* [x] Add an additional study 'type' (like Ember frame player, which requires a repo URL & commit SHA) to represent external studies via a migration. For external studies, collect (a) scheduled vs. any time, (b) signup/participation URL. Note that we'll want to be able to reference this specific type somehow (e.g. to filter for external studies), so we'll need to store a reference to this record.

### Response table

* [x] Add either a field for response type (i.e. study type at time of data collection) OR is_external (false by default). Note that we're thinking here about having exactly two study types, but we might eventually have additional types (a Lookit-hosted scheduled study, for instance, or another experiment runner type like jsPsych).
* [x] Add "confirmed" Boolean field reflecting whether the researcher has confirmed participation in the external study.

### Study detail view (participant-facing) + preview view (researcher-facing)

* [x] Don't require external studies to have experiment runner built in order to preview/participate
* [x] Clearly display that study is being run outside of Lookit for external studies. Still have parent select a child and click to participate. Indicate to parents that clicking to participate via Lookit will make it possible for researchers to contact them & will share profile information with researchers.
* [x] For external studies, customize button text based on whether study is synchronous - e.g. "Participate now!" vs. "Schedule a time to participate" (possibly "Share my profile and..." per above, but that may be too long)
* [x] For external studies, don't allow creating more than one response per child. (The parent can be directed to the study again, just don't need to make a new response.) 
* [x] Set up "participate now!" button to redirect to the participation site (with url param for child HASHED ID added, e.g. `?child=ABCDEF`) and to create a minimal "response" when the participant clicks to participate in an external study with a specific child

### Study list view (participant-facing)

* [x] Mark (can be reasonably subtle) on cards whether study is external
* [x] Allow filtering on Lookit/external studies (e.g. checkbox for "Include studies hosted outside Lookit" or allow selecting/deselecting each of "Lookit studies" and "Studies outside of Lookit")
* [x] Allow filtering on synchronous/asynchronous studies (e.g. checkboxes to show studies where we can participate... [ ] right now, [ ] at a scheduled time with a researcher. Count Lookit studies as synchronous.)
* [x] When filtering out studies the participant has already done, don't require completed_consent_frame on external ones; instead require confirmed.

### Study participation proxy view & preview proxy view

* [x] Eliminate for external studies (if study active but external, show informative error message). Note: we could alternately try to extend the proxy views to actually serve as proxies for the external websites - but I think at least for now we want to make the "you're leaving Lookit, we're not as responsible for what goes on out there" distinction clearer. 

### Study detail view (researcher-facing)

* [x] For external studies, don't display build status or build experiment runner button
* [x] For external studies, don't display link to "Review consent"

### Study edit view (researcher-facing)

* [x] For external studies, disable or don't display exit URL, protocol configuration, and protocol generator function
* [x] Remove exit URL from model.
* [x] Don't allow changing the type of a study, at least once there are any responses. (If allowing changes until responses exist, clearly indicate that this is the case in e.g. help text about the experiment type.)

### Study API 
* [x] Add hard-coded exit URL to always return history list view. 

### Study build

* [x] Make sure attempt to build external study fails gracefully/sensibly

### Study list view (researcher-facing)

* [x] For external studies, show expressed interest / confirmed participation counts instead of current complete/incomplete, consent approved/pending. 

### Previous studies page (participant-facing)

* [x] Divide into Lookit studies and external studies tabs
* [x] Don't require completed_consent_frame to show external studies
* [x] Display participation in external studies as "pending" or "confirmed" with brief explanations

### Study data access views [researcher-facing]

* [x] From "all responses" allow access to all external responses (and associated child/parent data), even without confirmed consent.
* [x] From "all responses" view, omit frame data & frame data dictionary downloads for external studies. Have the frame data downloads return a sensible error. Response data JSON, response data CSV, child data can all stay.
* [x] Add 'confirmed' status of responses to response data JSON & response data CSV. 
* [x] From "demographic snapshots" view, allow access to demographic snapshots associated with any external response, not filtered on consent. 
* [x] Remove "Videos" link for external studies OR clearly indicate it's not applicable.
* [x] Individual responses view: display pared-down info for external responses, and adjust columns of table as appropriate.
* [x] Display and allow editing of "confirmed" status from individual responses view. (Default should be confirmed as discussed above.) 
* [x] Edit language throughout (individual responses, all responses, demographic snapshots) regarding there being N responses as appropriate for external studies ("no responses with confirmed consent" vs "no responses yet", etc.)

### Messaging participants

* [x] Message participants view should allow contacting anyone with a response to this study for external studies (this may be the same as for regular Lookit experiment runner studies)

### Consent manager

* [x] No consent coding view for external studies (view returns error, no link from study detail page) OR at least clearly mark on consent manager that it's not applicable to external studies.

### Sending study announcements

* [x] Use a different constant to represent how many announcements are sent per day about external studies (this can be the same for now! just making sure it's easy to adjust. long-range, this is the sort of thing it might make sense to move to a table so we could edit it in the DB instead of redeploying - how are constants like this typically handled?)
* [x] When checking for previous participation, don't require completed_consent_frame on external responses - always count the participant as already having participated if there's a response.

## Tests:

To fill in!

okaycj:
This PR is to add support for external studies.   

We're going to merge this PR into staging to begin QA.  There is still work that needs to be done tests.


studies/migrations/0074_add_confirmed_to_response.py
```diff
@@ -0,0 +1,16 @@
...
+
+    dependencies = [("studies", "0073_add_study_type_to_response")]
+
+    operations = [
+        migrations.AddField(
+            model_name="response",
+            name="confirmed",
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
kimberscott, line 16: might want to squash with previous migration (at the point of actually merging in). Hooray for the first steps towards external studies today!

PR #754, opened 2021-07-16
Update logic for filter "hide studies we've done"

okaycj:
This PR fixes a bug that was introduced with the recent update to the studies list page filter.  This update should now correctly only display studies that have responses where the field `completed_consent_frame` is False.

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -233,7 +233,7 @@ class StudiesListView(generic.ListView, PaginatorMixin, FormView):
         user = self.request.user
         page = self.request.GET.get("page", 1)
 
-        qs = super().get_queryset().filter(state="active", public=True)
+        studies = super().get_queryset().filter(state="active", public=True)
 
         # values from session
         search_value = session.get("search", "")
@@ -243,10 +243,7 @@ class StudiesListView(generic.ListView, PaginatorMixin, FormView):
         )
 
         if search_value:
-            qs = qs.filter(name__icontains=search_value)
-
-        # convert to list as it's no longer being treated as a queryset
-        studies = list(qs)
+            studies = studies.filter(name__icontains=search_value)
 
         if child_value:
             if user.is_anonymous:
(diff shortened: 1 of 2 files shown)


web/views.py
```diff
@@ -290,11 +292,11 @@ def get_initial(self):
...
 
-    def completed_consent_frame(self, studies, child):
+    def studies_without_completed_consent_frame(self, studies, child):
         return [
             r.study
             for r in Response.objects.filter(
-                study__in=studies, child=child, completed_consent_frame=True
+                study__in=studies, child=child, completed_consent_frame=False
             ).distinct("study_id")
         ]
```
kimberscott, line 301: this is unfortunately not quite the logic we want: what we need is studies for which there is *no* response with a completed consent frame. This is fetching studies where there are responses that don't have completed consent frames. More concretely, say that there are the following studies:
* Study 1: baby has no responses
* Study 2: baby started study, but didn't complete consent frame
* Study 3: baby started study but didn't complete consent frame one time, then tried again and completed consent frame
* Study 4: baby started study and completed consent frame

When we hide studies we've done before, we want to show studies 1 and 2. I think this logic is going to show studies 2 and 3 instead.

kimberscott (changes requested):
Still needs a slight change in the logic. I could be confused, though, so please double check - booleans are hard :)

web/views.py
okaycj: Hide these:
1. Any complete consent frame flags on any responses.  There could be more than one response. 
2. Show studies that have no responses.

okaycj: I have updated the query in 1d23c60.

web/views.py
```diff
@@ -290,13 +289,9 @@ def get_initial(self):
...
-    def completed_consent_frame(self, studies, child):
-        return [
-            r.study
-            for r in Response.objects.filter(
-                study__in=studies, child=child, completed_consent_frame=True
-            ).distinct("study_id")
-        ]
+    def studies_without_completed_consent_frame(self, studies, child):
+        query = Q(responses__child=child, responses__completed_consent_frame=True)
+        return studies.exclude(query)
```
kimberscott, line 294: lovely! looks like this logic required doing things substantially differently, but it's done very cleanly :)

kimberscott (approved):
LGTM!

PR #755, opened 2021-07-19
Move log in message to search bar

okaycj:
This is to move the "log in message" to the search bar.


kimberscott (approved):
LGTM! Thanks :)

PR #756, opened 2021-07-19
Update age range in search filter

okaycj:
Update age range in search filter to cover all ages throughout childhood.  For example, before the age range "1 - 2" and "3 - 4" doesn't include the entire year "2".  This has been fixed by changing age range "1 - 2" to actually cover until age "3".  This has been update will all age ranges.


kimberscott (approved):
Wonderful, thanks!

PR #762, opened 2021-07-22
Cleared search form will no longer select checkbox

okaycj:
Fixes a bug where the checkbox in the search bar would be selected when the clear button is pressed.  

Note:  This bug shines a light on the fact that we don't have a Splinter/Ghost Inspector/Selenium test suite.


kimberscott (approved):
Thanks for catching this! And yes, I think there's an issue somewhere for Selenium(-type) tests.

PR #765, opened 2021-07-27
No longer remove search form values from session

okaycj:
Two issues came out of the recent push to master:

1.  The css looked a bit odd, mainly the border between the left and right sections on the studies list cards.  I removed this border and some unnecessary padding. 
2. The filter was removed when a user changes pages.  This made for an odd experience.  The solution was to no remove the filter values from session and the user will have to press the clear button if they want the for cleared.  The browser's reload button will no longer clear the filter form.

### DIFF
--- a/web/static/base.css
+++ b/web/static/base.css
@@ -243,10 +243,6 @@ button.account-row {
     background-color: #f8f8f8;
 }
 
-.panel.study-list .right {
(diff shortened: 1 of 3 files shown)


kimberscott (approved):
Looks good, thanks for the CSS changes as well!

PR #767, opened 2021-08-01
Add form validation to the search bar.

okaycj:
This PR adds validation to the search bar form.  This should prevent the error recorded in sentry here:

https://sentry.io/organizations/massachusetts-institute-of-technology/issues/2537850350/?environment=production&project=299962&referrer=alert_email

### DIFF
--- a/web/views.py
+++ b/web/views.py
@@ -224,7 +224,9 @@ class StudiesListView(generic.ListView, PaginatorMixin, FormView):
             for field, value in form.clean().items():
                 request.session[field] = value
 
-        return self.form_valid(form)
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
 
     def get_queryset(self):
         session = self.request.session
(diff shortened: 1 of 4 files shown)


web/views.py
```diff
@@ -221,12 +221,10 @@ def post(self, request, *args, **kwargs):
...
         if form.is_valid():
-            for field in self.form_class().fields:
-                request.session[field] = form.data.get(field, "")
+            for field, value in form.clean().items():
+                request.session[field] = value
 
-            return self.form_valid(form)
-        else:
-            return self.form_invalid(form)
+        return self.form_valid(form)
```
kimberscott, line 227: just leave a quick comment here about the intent so we're not confused later about why we're returning form_valid(form) regardless of whether the form is actually valid

web/views.py
```diff
@@ -238,9 +236,8 @@ def get_queryset(self):
         # values from session
         search_value = session.get("search", "")
         child_value = session.get("child", "")
-        hide_studies_we_have_done_value = (
-            session.get("hide_studies_we_have_done", "") == "on"
-        )
+        hide_studies_we_have_done_value = session.get("hide_studies_we_have_done", "")
+        print(hide_studies_we_have_done_value)
```
kimberscott, line 240: print was probably just for debugging?

kimberscott (changes requested):
A few comments and just need to update the test to expect the form will successfully load with invalid input - but this is really nicely restructured!

web/views.py
okaycj: I refactored this to have the "form_invalid()" call back in.  Thank you for calling this out.

web/views.py
okaycj: Whoops! Removed. Good eye!

PR #769, opened 2021-08-04
Merge search bar bug fix into master.

okaycj:
As the title says, merge the latest search bar bug fix into master.


kimberscott (approved):
Let's go for it! If we can the develop/master history sorted out so that we're merging in only the few commits that are actually new that'd be great though. Rebase develop on master (and force push) after this maybe - would that work?

PR #775, opened 2021-08-19
Fix don't show studies that a child has attempted

okaycj:
Yet another attempt to fix this filter.  The issue this time laid with the confusing documentation around "exclude".  In the end, looking up the studies we want to exclude and then excluding them was a more correct.


kimberscott (approved):
Thanks for the additional tests & sorry not to have caught that exclude behavior the first time around!

PR #783, opened 2021-09-07
 Mostly complete first draft of internationalization and translation 

rhodricusack:
Summary of primary changes

- translations of password validation hints are not shipped with django, so make sure these are in the .po files so our translators can provide them 
- if LoginRequiredMixin redirects to login URL, make sure it adds language code (web/mixins.py) 
- lots of snagging on missed words and phrases in i8n
- convert most flat pages into templates, so their contents are translated 
- many new .po and corresponding .mo files


locale/ja/LC_MESSAGES/django.po
```diff
@@ -1,149 +1,121 @@
...
 
+#. Shall I put the period "。" at the end of a sentence even when the original English sentence does not have a period?
 #: accounts/forms.py:90
-msgid ""
-"There's an update about a study we participated in (for example, results are "
-"published)"
-msgstr "参加した研究の更新があります（例えば、結果が公開されています）。"
+#, fuzzy
+msgid "There's an update about a study we participated in (for example, results are published)"
+msgstr "参加した研究の更新があります（例えば、結果が公開されています）"
```
okaycj: I am learning here, but from [this stack overflow](https://stackoverflow.com/a/1377396) should we be leaning away from fuzzy translations?

locale/ja/LC_MESSAGES/django.po
```diff
@@ -1381,14 +1254,8 @@ msgid "We are not running any studies at this time!"
...
-msgid ""
-"Looking for more ways to contribute to research from home? Check out <a href="
-"\"https://childrenhelpingscience.com/\" target=\"_blank\" rel=\"noreferrer "
-"noopener\">Children Helping Science</a> for even more studies!"
-msgstr ""
-"自宅で研究に貢献する方法をお探しですか?<a href=\"https://"
-"childrenhelpingscience.com/\" target=\"_blank\" rel=\"noreferrer noopener"
-"\">Children Helping Science</a>では、さらに多くの研究をご覧いただけます。"
+msgid "Looking for more ways to contribute to research from home? Check out <a href=\"https://childrenhelpingscience.com/\" target=\"_blank\" rel=\"noreferrer noopener\">Children Helping Science</a> for even more studies!"
+msgstr "自宅で研究に貢献する方法をお探しですか?<a href=\"https://childrenhelpingscience.com/\" target=\"_blank\" rel=\"noreferrer noopener\">Children Helping Science</a>では、さらに多くの研究をご覧いただけます。"
```
okaycj: This might be something on our end, but should we excluding HTML from translations?

Procfile
```diff
@@ -0,0 +1 @@
+web: gunicorn project.wsgi
```
okaycj, line 1: You guys are running on heroku!  Nice, I wish we could've been deployed here.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+-r ./requirements/defaults.txt
+-r ./requirements/dev.txt
```
okaycj: I can understand if this is a requirement of Heroku, but we're no longer maintaining these files.  Is there a solution were you could be using poetry?  I'd hate for our work with dependabot and keeping up with security updates to be over looked.

web/views.py
```diff
@@ -222,6 +224,9 @@ class StudiesHistoryView(LoginRequiredMixin, generic.ListView):
     template_name = "web/studies-history.html"
     model = Study
 
+    # i18n for redirect
+    login_url = '/' + get_language() + '/login.html'
+
```
okaycj: This might be unnecessary.  The build in translation view appears to be working.  For example, https://lookit-staging.mit.edu/es/login/.  Maybe I am misunderstanding something.

web/mixins.py
```diff
@@ -2,14 +2,15 @@
...
         lang=get_language_from_request(request, check_path=True)
-        if lang:
-            print('Switching login URL from this %s'%self.login_url)
+        if lang:   
+            # If it isn't already there, add this to the front of the login URL
             if not self.login_url[:(2+len(lang))] == ('/' + lang +'/'):
                 self.login_url='/' + lang + self.login_url
-            print(' to this %s'%self.login_url)
+        # Continue as normal
         return super().dispatch(request, *args, **kwargs)
```
okaycj: I love this new login mixin.  Could we clean up the unnecessary commits that are just prints?

web/views.py
```diff
@@ -292,7 +292,7 @@ def dispatch(self, request, *args, **kwargs):
             return super().dispatch(request)
         else:
             return HttpResponseForbidden(
-                _("The study %s is not currently collecting data - the study is either completed or paused. If you think this is an error, please contact %s")%(study.name,study.contact_info)
+                _(f"The study {study.name} is not currently collecting data - the study is either completed or paused. If you think this is an error, please contact {study.contact_info}")
             )
```
okaycj, line 312: If I am remembering correctly, I ran into this issue while I was working on recent feature.  I ended breaking up the string using concatenation and translating the "unchanging" parts of the string.  For example:
```
_("The study ") + study.name + _("is not currently collecting data - the study is either completed or paused. If you think this is an error, please contact") + study.contact_info
```
I understand the variable won't be translated in the above version, but maybe you guys have a preference.  I could go either way, I just want to be consistant and make the lives of the translators are easy as possible.

locale/ja/LC_MESSAGES/django.po
okaycj: I am just now realizing these commits are out of date.  I am going to mark them as resolved and review the code as a whole, instead of commit by commit.  My apologies.

requirements.txt
okaycj: Resolving outdated commit.

accounts/models.py
```diff
@@ -40,6 +40,7 @@
     StudyPermission,
 )
 
+from django.utils.translation import ugettext_lazy as _
```
okaycj, line 43: why did we make the decision to move to `ugettext_lazy` here?  Everywhere else is `gettext_lazy`.  Additionally, [this stack overflow](https://stackoverflow.com/a/4164683) post suggests that the lazy variant shoudn't be used in models.  What has been your experience?

web/views.py
rhodricusack: Sometimes the location of the inserted component shifts between languages, and if someone just translates "The study" and "is not currently collecting..." then the concatenated result might not make sense. I'm not sure about this specific sentence, but we had this problem in another location with Japanese recently. It is probably less error prone to translate whole sentences, as long as our translators are well-schooled in not translating the {study.name} tag.

rhodricusack:
> @rhodricusack As I stated in above in a review, I hadn't realized that I was referencing commits that were out of date. I will review the code as a whole, instead of reviewing commit by commit as I normally would.
> 
> My apologies for the confusion.

If you do have any obvious pointers on best practice, to improve the readability of my commits, please do so! As I was debugging via heroku at the beginning of the process, I think I committed every small change, even those I was about to role back, which is probably rather unmanagable.

Procfile
okaycj: That's unfortunate.

accounts/signals.py
```diff
@@ -14,7 +68,7 @@ def post_migrate_create_flatpages(sender, **kwargs):
         <div class="main">
         	<div class="home-jumbotron">
         		<div class="content">
-        			<h1>Lookit<br>
+        			<h1>Lookit MBAH<br>
         			<small>the online child lab</small></h1>
```
okaycj, line 72: This should be reverted back.

web/views.py
okaycj: I think going forward, we'll stick to your "method" of translating text.  I'll am going to put in a issue to confirm that all translated text doesn't take on my "incorrect concatenation" I mentioned above.   

https://github.com/lookit/lookit-api/issues/840

accounts/signals.py
```diff
@@ -1,8 +1,59 @@
...
+
+    site.domain = settings.SITE_DOMAIN
+    site.name = settings.SITE_NAME
+
+    site.save()
+    FlatPage = apps.get_model("flatpages.FlatPage")
+
+    # Path to raw github files. English in this folder, others in locale/[langcode]/
+    github_root = "https://github.com/manybabies/mbah-research-resources/raw/master/lookit%20static%20pages"
+
```
okaycj, line 26: It looks like we translating the flatpages to static pages and then hosting them on github.

If this is true, then why don't we just remove the concept of flatpages all together and replace them with templates.  This would allow the current translation process of using `.po` files to continue and we wouldn't have to go through MBAH to make updates.

accounts/templates/accounts/researcher-registration.html
```diff
@@ -13,7 +13,7 @@ <h1 class="panel-title">Create Researcher Account</h1>
                     {% csrf_token %}
                     {% bootstrap_form form %}
                     {% buttons %}
-                        <p> By clicking 'Create Account', you agree that you have read and accepted the <a href="/termsofuse/" target="_blank">Terms of Use</a> and <a href="/privacy/" target="_blank">Privacy Statement</a>.
+                        <p> By clicking 'Create Account', you agree that you have read and accepted the <a href="../termsofuse/" target="_blank">Terms of Use</a> and <a href="../privacy/" target="_blank">Privacy Statement</a>.
```
okaycj, line 16: I think I am going to do some research here, but I am still on the idea that we should be using the [URL tag](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#url) to generate urls.  Since translations (and the translations view) are handled by Django, it would seem like the urls would also be updated on HTML creation.

PR #788, opened 2021-09-08
Updates for external studies

okaycj:
This PR will contain additional test cases as well as fixes from initial QA of staging.

### DIFF
--- a/studies/models.py
+++ b/studies/models.py
@@ -1,6 +1,7 @@
 import logging
 import uuid
 from datetime import datetime
+from enum import Enum
 
 import boto3
 import dateutil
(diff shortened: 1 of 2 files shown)


mekline (approved):
Only question before you merge this is about the content of the tests you factored out - do you now test the same things, just separately for the structure of internal/external, or are there tests that either don't seem to apply or that are new for external? If either, I'd just like to understand the content of what's changing.

okaycj:
> Only question before you merge this is about the content of the tests you factored out - do you now test the same things, just separately for the structure of internal/external, or are there tests that either don't seem to apply or that are new for external? If either, I'd just like to understand the content of what's changing.

There are two sets of tests I add with this PR.  

1. I added some code to the model to identify which study type I am working with.  These are simple tests where I create a study of an expected study type and verify the type.  
2. I wanted to make sure the SQL query that finds which persons to email about studies was working for external studies.  This query wasn't something I was sure I got correct, so the tests were to verify my logic.  

The focus of all the tests from this feature have been on the changes within the feature.  The above tests are not all of the test for this feature.  Those tests were included in the feature PR.  There should have been and still needs to be done, a set of tests that verify the security of these changes.  At least, we need to determine if such security tests are even needed or covered in other tests.  

Let me know if you'd like more information or if you'd like a walk through of the code.

PR #795, opened 2021-09-13
Update studies list view search bar

### LINKED ISSUE #785: Alter tab name & selection on Studies list view
If at all possible, can we please have a second drop-down next to/near the "find study for" dropdown that has the following options:

[Default] Show studies that...
...happen on Lookit
...happen somewhere else

The first option should limit what's listed to internal studies (same as current "Lookit studies" tab), the second should limit to external studies (same as current "External studies" tab)

Instead of the current set of 5 tabs, I'd like for there to be 3 tabs, with additional text for each. 

***
"All studies" - Synch and async studies, respecting selection of internal/external

Text: Please note you'll need a laptop or desktop computer (not a mobile device) running Chrome or Firefox to participate, unless a specific study says otherwise.

***
Studies happening right now - Show sync/unscheduled studies only, respecting selection of internal/external

Text: You and your child can participate in these studies right now by choosing a study and then clicking "Participate." Please note you'll need a laptop or desktop computer (not a mobile device) running Chrome or Firefox to participate, unless a specific study says otherwise.

***
Scheduled studies - Show async/scheduled studies only, respecting selection of internal/external

Text: You and your child can participate in these studies by scheduling a time to meet with a researcher (usually over video conferencing). Choose a study and then click "Participate" to sign up for a study session in the future.

okaycj:
This is an update to the search bar for the Studies List view.  This closes #785.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
 
 install:
   - poetry env info
-  - poetry install --quiet > /dev/null
+  - poetry install
 
 script:
   - pre-commit run --all-files
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -464,9 +464,9 @@ class StudyListSearchForm(forms.Form):
         adults = ("18,999", _("adults (18+)"))
 
     class StudyLocation(FormChoiceEnum):
-        empty = ("0", "Show studies that are...")
-        lookit = ("1", "...happening on Lookit")
-        external = ("2", "...happening on other websites")
+        empty = ("0", _("Show studies that are..."))
+        lookit = ("1", _("...happening on Lookit"))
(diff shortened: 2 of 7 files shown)


mekline (changes requested):
Requesting a small language change for the StudyLocation choices.  If it reads:

Show studies that are...
...happening on Lookit
...happening on other websites

Then after you fill both dropdowns the grammar will make sense!  I just realized this. EG "Find studies for...toddlers(1-2), happening on Lookit"

mekline:
Travis CI didn't enjoy this edit; I'm assuming you know why better than I do!

okaycj:
> Travis CI didn't enjoy this edit; I'm assuming you know why better than I do!

I don't, but I'll figure it out.

PR #818, opened 2021-10-06
Remove mo files

okaycj:
Now that we've got the job in GCP to `compilemessages` on deploy (lookit/lookit-orchestrator#25), all `mo` translation files shouldn't be in the repository.


mekline (approved):
All makes sense, thank you!

PR #822, opened 2021-10-11
Test for multiple response for external study type

okaycj:
This PR covers the test case for there being multiple responses for an external study.


okaycj:
@mekline 

I agree with this for frame player studies.  This limitation is only applied to external studies.  I don't know of a reason a researcher would need more then one response.  I've imagining the responses on an external study gives the researcher access to demographic data.

PR #825, opened 2021-10-11
update str method for ConsentRuling

okaycj:
Removed some unnecessary code from the ConstentRuling model.  The `__str__` method was handling a case were the arbiter might not be present, this should never happen.


okaycj:
> After conversation, let's make an issue for this.

Issue #829 was created.  

@mekline Should we accept this change or close this PR?

mekline (approved):
Let's keep this change! IMO an error/default NA responses is better indicator something's off than a string 'None' silently entering the db...

PR #862, opened 2021-11-10
Show external and scheduled in study edit form

### LINKED ISSUE #861: Edit Survey form should show selected values for external and scheduled
# TL;DR
Right now, the external/scheduled values are hidden from researchers in the edit form, because these values are not modifiable after creation.  Instead, they should appear but be grayed out, so the researcher can see what they are, in parallel to how they appeared on the create form page.

# Narrative
As a researcher, I want to know the values I set for my study when I created it, even if I can no longer modify some of those values.  The Create and Edit forms should be as close to identical as possible so I can understand where all my study parameters are.

# Acceptance Criteria

All 4 boxes appear on the edit form, with the values that the researcher previously set pre-set (just as all other fields in the form are set to current values).  

Both the External and Scheduled boxes should be greyed out/unchangeable, regardless of their values (this is a different behavior from the greying-out of scheduling on the Create form, which is annoying but necessary)

# Implementation Notes
Leave important notes/suggestions here for implementation (important files, suggested approaches, etc.

This is the current Create Study checkbox structure; the Edit Study checkboxes should look the same but have the different behavior described above (all 4 boxes set to current value; External and Scheduled unmodified)

Screen Shot 2021-11-09 at 11 31 40 AM
)

okaycj:
This PR is to show the "External" and "Scheduled" on the Study Edit Form.  

Closes #861

Screen Shot 2021-11-10 at 11 15 27 AM
Screen Shot 2021-11-10 at 11 15 31 AM
Screen Shot 2021-11-10 at 11 15 37 AM
Screen Shot 2021-11-10 at 11 15 40 AM

### DIFF
--- a/exp/tests/test_study_views.py
+++ b/exp/tests/test_study_views.py
@@ -416,9 +416,8 @@ class StudyViewsTestCase(TestCase):
             "Study build was not invalidated after editing metadata",
         )
 
+    @patch("django.views.generic.edit.ModelFormMixin.form_valid")
     @patch("exp.views.mixins.StudyTypeMixin.validate_and_fetch_metadata")
-    @patch("exp.views.study.HttpResponseRedirect")
-    @patch("exp.views.study.reverse")
     @patch("exp.views.study.StudyUpdateView.get_form")
     @patch.object(StudyUpdateView, "request", create=True)
     @patch.object(SingleObjectMixin, "get_object")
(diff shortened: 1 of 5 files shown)


mekline (approved):
I'll be substantially more helpful for feedback viewing the changes on develop, so let's go ahead :)

PR #880, opened 2021-12-13
Remove eligibility template

okaycj:
I have been wrestling with the decision to remove this template for a couple of weeks.  I am unable to find how it's used in the current site.  I don't want to as there's good code in this template, but it keep getting tagged in sonar cloud.  It should probably go at this point.


kimberscott:
(shh I'm not here but just fyi it's not used! This was added as part of the initial work to allow specification of eligibility criteria and turned out after a lot of discussion not to actually address all expected use cases (e.g. disjoint sets), and would have been a lot more work to implement fully than the "criteria expression" solution we settled on. Probably GitHub has better history of the discussion but this is just to confirm you're not deleting anything necessary.)

mekline (approved):
As per Kim, adios to this template!

PR #882, opened 2021-12-15
Move flatpages to templates

okaycj:
This PR is to remove existing flat pages from the our site so that they may be translated.


mekline (changes requested):
The logic of the reorganization basically makes sense to me, but can you please demo me how the translation model works e.g. on the home.html page? 

Sonarcloud is reporting 'code smells' and a few other things, so please let me know if any of those warrant discussion/explanation.  Do you generally find you can correct all these issues, or should be discussing/documenting what happens with those it makes sense to leave?

okaycj:
> Sonarcloud is reporting 'code smells' and a few other things, so please let me know if any of those warrant discussion/explanation. Do you generally find you can correct all these issues, or should be discussing/documenting what happens with those it makes sense to leave?

I'll fix these issues and update the PR.

okaycj:
> The logic of the reorganization basically makes sense to me, but can you please demo me how the translation model works e.g. on the home.html page?

Sure thing!  I'll put up some images of the home page translated.

mekline:
We can do a first pass in standup today also!

okaycj:
This works is being done in #904.  Closing.

PR #883, opened 2021-12-20
Move develop into Master

okaycj:
This PR covers many major updates:
- Django version bump to 3.2
- Fix to email table
- Many fixes from sonarcloud.io


mekline (approved):
Good to go! We've discussed all of these previously, and staging seems to be rendering the formerly static pages as I'd expect!

PR #895, opened 2022-02-08
Merge develop into master

okaycj:
This PR will merge in recent updates from dependabot.


mekline (changes requested):
(Just requesting changes bc there are reported tests failing)

okaycj:
> (Just requesting changes bc there are reported tests failing)

That makes sense.  I don't think I've had the google trigger fail before.  They've been slow all day, I hope this won't be an issue.  I re-ran the trigger.

okaycj:
> They've been slow all day, I hope this won't be an issue.

It turned out to be an issue.  I needed to update how we install/run poetry on docker build.

okaycj:
Confirmed acceptance from @mekline via slack.

PR #904, opened 2022-02-15
Contributions from MBAH: Converting flatpages to templates, snagging

rhodricusack:
I've re-instated the flatpages-from-templates, which can now be translated. 

I also had to tweak some settings to re-enable translation.

I snagged one issue - the translation of the field label "Nickname" on the participant signup form.


accounts/forms.py
```diff
@@ -182,7 +182,8 @@ class Meta(LowercaseUsernameUserCreationForm.Meta):
 
 
 class ParticipantSignupForm(LowercaseUsernameUserCreationForm):
-    nickname = forms.CharField(required=True, max_length=255)
+# Rhodri: I don't know why this was here - nickname is in User model
+#    nickname = forms.CharField(required=True, max_length=255)
```
okaycj, line 187: This doesn't need to be here.  `nickname` is defined in the meta section of the form class.  Please remove it.

web/templates/flatpages/default.html
```diff
@@ -61,8 +61,8 @@
                         <li><a href="https://accessibility.mit.edu/">{% trans "Accessibility" %}</a></li>
 						<li><a href="/contact_us/">{% trans "Contact us" %}</a></li>
 						<li>{% trans "Connect" %}:
-						<a href="https://www.facebook.com/lookit.mit.edu" target="_blank" rel="noopener"><i class="fa fa-facebook fa-lg social-link"></i> </a>
-						<a href="https://www.instagram.com/babiesoflookit/" target="_blank" rel="noopener"><i class="fa fa-instagram fa-lg social-link"></i> </a>
+						<a href="https://www.facebook.com/lookit.mit.edu" target="_blank"><i class="fa fa-facebook fa-lg social-link"></i> </a>
+						<a href="https://www.instagram.com/babiesoflookit/" target="_blank"><i class="fa fa-instagram fa-lg social-link"></i> </a>
```
okaycj, line 66: There's a security issue with `target="_blank"` links. `rel="noopener"` is the fix for this issue and should remain.

web/urls.py
```diff
@@ -40,34 +40,13 @@
...
-    path("", flatpages_views.flatpage, dict(url=""), name="home"),
+    path("", views.HomeView.as_view(), name="home"),
+    path("faq/", views.FAQView.as_view(), name="faq"),
+    path("privacy/", views.PrivacyView.as_view(), name="privacy"),
+    path("scientists/", views.ScientistsView.as_view(), name="scientists"),
+    path("contact_us/", views.ContactView.as_view(), name="contact"),
+
+# Remaining flat pages
+    path("resources/", flatpages_views.flatpage , dict(url="/en-us/resources/"), name="resources"),
+    path("termsofuse/", flatpages_views.flatpage , dict(url="/en-us/termsofuse/"), name="termsofuser"),
```
okaycj: Would it be too much of a lift to port over the remaining flatpages to templates?  If so, let me know and I can on that work.

okaycj (changes requested):
Thank you for this work.  Only small changes are required here.  Generally, I would like the construct of flatpages to removed from this project.  Here a few things that still need to be done:

1. Give the directory that the flatpage templates are in a different name.
2. Remove flatpages plugin from settings. 
3. Remove flatpages from database.

web/urls.py
rhodricusack: No problem, I can do this! We just didn't do them because we don't plan to translate them - the resources because they're very country specific, and the terms-of-use because we're not lawyers.

Something on my mind: do you think we should get rid of the "flatpages" directory for the templates. This is going to hang around like an old legacy, long after it is relevant.

okaycj: Thanks for do this.  Yes, I do think we should get rid of the directory, after they've been all moved to templates.

accounts/forms.py
rhodricusack: done

web/templates/flatpages/default.html
rhodricusack: done

web/urls.py
rhodricusack: Done!

rhodricusack:
I have done all except
4. Remove flatpage tables from database
which I think has to be done at your end?

okaycj:
> I have done all except
> 4. Remove flatpage tables from database
> which I think has to be done at your end?

The table will need to be removed via a migration.  I can push a commit if you'd rather, but for sure this is something you can do.  Let me know if you would you like to pair program on this issue.

rhodricusack:
It would be good to see how to remove the table in the migrations if we can match our schedules. Alternatively, if it is quick, go ahead and I will look at the result-

okaycj (approved):
This looks good to me!  Thank you for all your hard work.  💯

PR #908, opened 2022-02-23
Lab view

### LINKED ISSUE #620: Option to display only studies from a specific lab
# TL;DR
Provide a way to see only studies from a given lab on Experiment.

# Narrative
As a staff member, it's helpful to be able to see what other studies a lab is running when reviewing one study or corresponding with lab members. 

As a researcher, it's helpful to see studies grouped by lab in the case of the Sandbox vs. Demo vs. actual labs.

# Acceptance Criteria

- [x] Add a study list view with custom URL slug showing only that lab's (active, discoverable) studies
- [x] New column on lab object for custom url; initially populated with the lab's display name reduced to [a-z0-9] plus hyphen
- [x] New field on the lab detail page https://lookit.mit.edu/exp/labs/n/ below Lab Website, named "Custom URL", and appearing as "https://lookit.mit.edu/studies/mycustomslug" (or whatever the right address for the search is)
- [x] New field added to Edit Lab page, below Lab Website.  Add descriptions below each field so it looks like this:

##### Lab Website
[Text entry field w/ current value]
*A link to an external website, such as your university lab page*

##### Custom URL
https://lookit.mit.edu/studies/ [Text entry field]
*A unique URL (https://lookit.mit.edu/studies/mylabname) that will show discoverable, active studies for this lab only*

- [x] The Custom URL field on the edit page should enforce uniqueness however the Lab Name field enforces it

- [x] Test that correct studies show up in filtered study list - should show only discoverable, active studies as with the normal study view page.

okaycj:
This PR adds a view that utilizes a slug set by labs to generate a page that contains only their experiments.  This view is a small modification on the current study list page and all the study list page functionality should work for this view as well.  

Closes #620


mekline (changes requested):
Oh shoot! I accidentally modified the acceptance criteria on a very old issue rather than the new one.  See https://github.com/lookit/lookit-api/issues/620

From my perspective the critical change would be to add the help text I specified there for both the lab website and custom URL, and change the name of the slug field (at least for the label, I don't mind what it's called in the db) - the reason for this is that researchers won't necessarily know what a "slug" is. To avoid chaos we need to cue them where to put their actual external website in one box, and what will happen to the string they enter into the other (i.e. result in a link lookit.mit.edu/mycoolslug that they can visit)

PR #926, opened 2022-03-02
Move Staging into Production

okaycj:
Moving a few fixes into Production:
1. Researcher registration link should be more visible
2. General accessibility updates

### DIFF
--- a/exp/static/exp/css/app.css
+++ b/exp/static/exp/css/app.css
@@ -50,6 +50,23 @@ a.state-link {
     border-right: 1px solid #ffffff;
 }
 
+nav.navbar-default.navbar-static-top ul.nav.navbar-nav li a{
+    color:#666666;
+}
+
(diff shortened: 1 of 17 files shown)


mekline (changes requested):
I spent a bunch of time drafting a comment about this in my mind, but never passed to you! I think the researcher-form notification is still too hard to read. Can you put a line break in, with the following text?

(1) Ready to participate in a Lookit study? Create a family account here!

(2) To make a researcher account, please use [this registration form] instead.

okaycj:
> I spent a bunch of time drafting a comment about this in my mind, but never passed to you! I think the researcher-form notification is still too hard to read. Can you put a line break in, with the following text?
> 
> (1) Ready to participate in a Lookit study? Create a family account here!
> 
> (2) To make a researcher account, please use [this registration form] instead.

PR #938 was created to facilitate change request.

mekline (approved):
Yay!

PR #935, opened 2022-03-07
Update for study list cards

### LINKED ISSUE #920: Study card redesign
-  [ ] Small cards
-  [ ]  No University, compensation info
-  [ ] New field: single sentence purpose section

okaycj:
Update study list cards to reflect CHS changes. 

Screen Shot 2022-03-07 at 12 44 05 PM

 
Closes #920

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 
 script:
   - pre-commit run --all-files
-  - python manage.py test
+  - python manage.py test --failfast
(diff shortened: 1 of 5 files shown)


mekline (approved):
Is the double-title the result of a study not having an image set? I think studies will always have images so not terribly concerned, just want to make sure I know what's happening here.

okaycj:
> Is the double-title the result of a study not having an image set?

When an image is no loaded, for whatever reason, the alt text is displayed instead.  The alt text for these images is the title of the study.  Your correct in your assumption that that all studies require an image.

PR #939, opened 2022-03-09
Move redesign from staging to production

okaycj:
This update moves a few changes to our website moving us closer to the CHS merger. 

1. Update home page with CHS logo.
2. Update study list page to more closely resemble CHS's study list page. 
3. Update Scientist's page to include more people and copy update.


mekline (approved):
All looks as expected so far; happy to have another look when the study card refresh is ready also!

PR #945, opened 2022-03-23
Upgrade python

okaycj:
Installing `Asdf` for local development environments has proven to be problematic.  To remove this necessity, I've updated the project to Python 3.9.x.  This version of Python should match the version installed by brew.  The lack of patch version might prove to be an issue in the future, but we'll cross that bridge when it comes.


mekline (approved):
Fingers crossed!

PR #946, opened 2022-03-24
Fix login next

okaycj:
Update login so that the current page is stored in the query string.

Login from study's detail page:
one

Login from navigation bar:
two


mekline (approved):
Looks great, and those screenshot gifs are very helpful to see what's going on!

PR #952, opened 2022-03-29
Add command to create site for local dev env

okaycj:
Add command to create entry in site database table during local environment setup. 

The Python package `Black` was updated as the pinned version was broken.


okaycj:
@mekline This doesn't change anything public facing.  I am just going to push this to development.

PR #972, opened 2022-05-26
Set a db password in the example env file

okaycj:
Historically, there's been a empty string as the password for the DB in the local environment.  This PR set the password to the normal default of "postgres" for the local environment.  Small change, but necessary for docker compose script.  Also, less confusing.


becky-gilbert (approved):
LGTM!

PR #976, opened 2022-06-08
Add rabbitmq host to web service

okaycj:
This PR covers 4 very small changes:

- Add the RabbitMQ host to the web service within the docker compose script.
- No longer deletes the database on `clean` make target. 
- Removes the docker images on `clean` make target.
- Exposes the database port to host machine so that we can use a database IDE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@ serve:
 	docker compose pull && docker compose up --build
 
 clean:
-	rm -rf ./data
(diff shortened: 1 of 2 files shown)


becky-gilbert (approved):
I'm no expert in this stuff (docker, rabbitmq) so I shouldn't be relied on to catch issues with the code changes. Also I don't have enough context about the goal of the changes here to say whether or not they make sense. That said, I don't see any obvious problems so LGTM!

PR #977, opened 2022-06-14
Add child hash id to response

okaycj:
### Summary
To provide the child hashed id, we needed to first generate it and then inject it into the response API.  To inject the field in the API we're using `SerializerMethodField`.  The function to generate the hashed id, required the Response model object to be in a dictionary form that I was unable to recreate.  To complete this task a bit more quickly, I just wrong a parallel function to the existing one and a unit test to verify that they work the same.  

### Off Topic Work
I am continually working on the docker compose script.  The latest change was to define a temp directory within the container.  This was to remove a warning.

### Additional Reading
- http://www.tomchristie.com/rest-framework-2-docs/api-guide/fields#serializermethodfield

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,3 +35,6 @@ local-certs:
 	mkdir -p certs 
 	mkcert -install
 	cd certs && mkcert local_lookit.mit.edu
+
+test:
+	docker compose run --rm web poetry run ./manage.py test --failfast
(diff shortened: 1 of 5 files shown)


becky-gilbert (approved):
Usual disclaimer: I'm not the best person to review. That said, it all looks very sensible to me!

PR #978, opened 2022-06-22
Added targets for the database

okaycj:
A small update to the makefile.  There are now targets for getting access to the psql shell and a target to pipe in SQL file for data import.  These two commands were added to the recent documentation update.


becky-gilbert (approved):
LGTM!

PR #980, opened 2022-06-23
Schuduled box editable

### LINKED ISSUE #927: Make Scheduled box editable in existing studies
**Description**
We set both External and Scheduled to be fixed after creation, because a single study object should not collect data crossing modes like this - this would be expected to be a separate study + review process. External/Internal drastically affects the initial info you fill in and results in different structures, so should stay fixed.  

*However* a stray click on the Schedule box can go unnoticed, and people should be able to alter this like any other field (but would still get a question about it if they did so with a live study already collecting data)

**How to reproduce**
Attempt to change Scheduled value on any Edit Study view

**Expected behavior**
Box can be checked and unchecked, identical in behavior to e.g. Discoverable box

**Additional context**
(First modification to the inital external studies rollout)

okaycj:
This PR will allow external studies' `scheduled` checkbox to be changed after save.  

Closes #927

edit-scheduled

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -329,7 +329,6 @@ class StudyEditForm(StudyForm):
     def __init__(self, user=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["external"].disabled = True
-        self.fields["scheduled"].disabled = True
         self.fields["structure"].help_text = PROTOCOL_HELP_TEXT_EDIT
         # Restrict ability to edit study lab based on user permissions
         can_change_lab = user.has_study_perms(
--- a/studies/templates/studies/_study_fields.html
+++ b/studies/templates/studies/_study_fields.html
@@ -114,7 +114,7 @@
     function updateScheduled(){
         const external = document.querySelector('#id_external')
         const scheduled = document.querySelector('#id_scheduled')
-        scheduled.disabled = external.disabled || !external.checked
+        scheduled.disabled = !external.checked
     }
 
     function updateScheduling (scheduledCheckBox) {            
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -10,7 +10,13 @@ WORKDIR /code
 
 COPY pyproject.toml poetry.lock /code/
 
-RUN apt-get clean \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends locales \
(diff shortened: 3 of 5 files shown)


studies/forms.py
```diff
@@ -329,7 +329,6 @@ class StudyEditForm(StudyForm):
     def __init__(self, user=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["external"].disabled = True
-        self.fields["scheduled"].disabled = True
```
becky-gilbert, line 332: Does `StudyEditForm` control the form's initial settings when a study is first created, or when the Edit button is clicked after the study has been created, or both? I'm guessing it's the latter, and thus the External checkbox still needs to be disabled because it was set when the study was created and can't change after that. Is that right?

studies/templates/studies/_study_fields.html
```diff
@@ -114,7 +114,7 @@
     function updateScheduled(){
         const external = document.querySelector('#id_external')
         const scheduled = document.querySelector('#id_scheduled')
-        scheduled.disabled = external.disabled || !external.checked
```
becky-gilbert, line 117: Makes sense!

becky-gilbert (approved):
Looks good! I think we could add tests for the external/scheduled checkbox behavior but I don't think that needs to hold up merging - I can just add it to a to-do list.
BTW I see that you included a commit that fixes the footer. Looks like it was an issue with a translation string accidentally including a link to the CHS site, rather than just the text around it. Fine with me if you want to add that fix in this PR, just flagging it in case the commit was unintentional 😃

okaycj:
> Fine with me if you want to add that fix in this PR, just flagging it in case the commit was unintentional

Oops, it was intentional.  I should have included that change in the description of the PR.  Thanks for calling it out.

studies/forms.py
okaycj: I would disagree.  I think tests should go with the PR.  I didn't include a PR, because I felt that this was a UI change that can't be tested with our current testing framework.  But you're correct, this is a change with the form class and should have a test to go with this change.

becky-gilbert: > I would disagree. I think tests should go with the PR. 

Ok, good to know! 😃 

> I felt that this was a UI change that can't be tested with our current testing framework

Ah ok, I thought we could test UI changes - in that case ignore me (or add a form class test if that's relevant/possible).

PR #981, opened 2022-07-05
Country view and new demographics fields

### LINKED ISSUE #790: Update other family identity
# TL;DR
"Other" by itself isn't enough, there is a need to capture the actual identity.

# Narrative
Researchers (ManyBabies) would like to provide a blank field when other is selected to give people a chance to record their actual identity.  

# Acceptance Criteria

- [ ] Tests for new feature
- [ ] QA on new feature

# Implementation Notes
We need to investigate how we're currently recording this data and make sure that there's reasonable place to store this data.

### LINKED ISSUE #796: Update demographic data form to be more international-aware
# TL;DR
The current demographics questionnaire is confusing/not relevant to people outside the united states; there are higher and lower-effort ways we can mitigate this! Will require consult with experts/researchers who do data collection internationally. MBAH brought this critique and would like a better version! 

# Narrative
"As a User, I would like to see questions in the demographic survey that make sense for people in my country, or at least signal when they aren't relevant."

<details>
<summary>Feature Request fields</summary>
<p>
**With this `details` wrapper, you can just use this ticket to record your findings.
When you're ready to turn this into an implementation request, delete the scoping label and
this wrapper, leaving only the filled out information.**

# Acceptance Criteria

- [ ] Tests for new feature
- [ ] QA on new feature
- [ ] Sign off from Kim

# Implementation Notes
Leave important notes/suggestions here for implementation (important files, suggested approaches, etc.)

</p>
</details>

okaycj:
This PR updates the Demographics form to cover the following:

1. Force country selection before showing form.
2. Make some fields conditional on country. 
3. Remove no longer needed fields. 

Closes #796
Closes #790


okaycj:
> accounts/forms.py and accounts/migrations/0054_rename_deprecated_fields.py
> 
> I just don't fully follow the logic of how (1) the presence or absence of the race_identification item is handled and (2) how the database migration is handled alongside these form changes. Can you please walk us through (1) how the logic works in the codebase and (2) what will happen when these changes hit master?

There are two things going on here... First, I thought you had wanted this field removed and so I removed the field.  The migration in this PR has the field removed from the database.  Second, I've since figured out that you actually don't want the field remove, but you want it renamed.  

Before we move these changes to staging, I will need to fix the migration so that the field is renamed.

okaycj:
> api/tests/test_demographics.py
> 
>     * What's going on with the transition from languages_spoken_at_home to additional comments 1 and 2? I generally understand this is to handle the deprecated fields but unclear what the api user will now see for these requests. (Also, any idea why French, English, Spanish were hard coded in this way?)

These are tests.  The text is hard coded to then be checked in an assertion.  I replaced a check on the `languages_spoken_at_home' field to another field that exists in the database.  I picked `additional_comments` randomly as it seemed reasonable.  

What the user sees is handled in `api/serializers.py`.  I removed these fields from the API response.  They'll need to be put back as these fields are not removed, but renamed.

mekline:
OK great! knowing that it's just replacement with a random different field
for testing answers all of my questions :)

becky-gilbert:
@okaycj on `update-demographics-form` branch, for `number_of_guardians`, when I select "Another number, or explain below", the text box that appears is labeled "Guardians explanation" with default text "Guardians explanation". Based on Melissa's spreadsheet I think the question title should be "Share more about your children's parents/guardians:" (not sure about default text?). 
Screen Shot 2022-07-12 at 10 58 02 AM
I would've suggested a specific change but couldn't see where the `guardian_explanations` question title and default text is set in the code.

becky-gilbert:
One more thing I noticed: everything works as expected when I select a country on the first demographics page (account/demographics/country). If I select US and save then I see the two US-dependent questions (`state` and `race_ethnicity_identification`), and if I select non-US then I don't see those questions.

However, if I first select a non-US country, save my response, and then _change country_ to US on the demographics  update page (account/demographics), then I do see the state question appear but I don't see the race/ethnicity question appear. Similarly, if I start with the US on account/demographics/country and then change to non-US on account/demographics, then the state question is correctly hidden but the race/ethnicity question is still shown. 

I think the race/ethnicity question should always show/hide depending on the US/non-US country selection, even when it's changed on the demographics update form. Is that right?

mekline:
Becky your proposal sounds good to me as well!

becky-gilbert:
@okaycj I just committed a change to show/hide the race/ethnicity question based on country: 0d2adbf02c64db8168b4d4b0800bb065735a720f Could you take a look please?

And if you agree that the `guardians_explanation` question title and default text should be changed, then please let me know where to find this in the code and I can update it.

web/static/js/demographics-form.js
```diff
@@ -22,15 +22,20 @@ $gender.on('change', function () {
...
     if ($(this)[0].value === 'US') {
         showField('state');
+        showField('race_ethnicity_identification');
     } else {
         hideField('state');
         $(`#id_state`)[0].value = '';
+        hideField('race_ethnicity_identification');
+        $(`input[name="race_ethnicity_identification"`).each(function() {
+            this.checked = false;
+        });
```
okaycj, line 38: Very nice.  💯

okaycj: After looking more into this, there's more that needs to done or discussed.  The template will remove these fields when the a non-US country is selected.  This means that if you save the form with a non-US country and attempt to change to US the fields can't be show.  They're not there.  

The reverse is true.  If you save the form as US, you can toggle the fields.  They exist.  

Do we want to have the field toggle-able (maybe not a word ;) )? Or we want the fields not exist when a the rest of the world sees this form?

okaycj: >What about referring to/thinking of this field as "usa_race..." and setting
>the initial value to 'NA'/Not applicable for all forms outside the US?  So
>if someone started on the USA and toggled, the (actual) value given would
>remain. If someone started elsewhere the values would be set to NA on first
>save. If they then toggled to US, they would see a set of non-checked boxes
>for that question, and could update if desired.  Does that make sense?

@mekline Feel free to make comments in these conversation boxes.  It makes life easier when needing to context switch.  

I think you're suggesting we clear the selection when country is toggled from non-US to US.  Great suggestion.  We're currently doing that to some degree.  I think we can formally implement it.  

I am inferring you'd prefer we toggle the fields and clear them when shown?

PR #990, opened 2022-07-12
Show study link on external studies

### LINKED ISSUE #970: Study link not displayed for external studies
**Description**
In an external study, this study link is not displayed even after the study status is set to "Active."
Instead, there is a prompt "Your study link will show up here when you have built the experiment runner." However, the experiment runner is not used for external studies, so it is not possible to build the experiment runner.

**How to reproduce**
The study link is not shown for **active external studies** with the following properties:
- Not scheduled, not discoverable
- Not scheduled, discoverable
- Scheduled, not discoverable*
- Scheduled, discoverable*

*These have not been yet confirmed, since it is also not possible to change an active external study from not scheduled to scheduled: when attempting to check the "scheduled" box the cursur becomes a red circle with a line through it.

**Expected behavior**
External studies should not reference the experiment builder.
Before the study is active, the text can read that the study link will be available once the study is active; the study link should be then be displayed when the study status is changed to "Active."

It should also be possible to check the "scheduled" box to change an external study which is set to "not scheduled" to be "scheduled"

**Screenshots**
image

any

**Additional context**
There is also an external study field for the Exit URL, but can this be determined by Lookit if the user is directed to an external site?

okaycj:
Update the conditional in the template to show the study link when the study type is external.

Closes #970

Screen Shot 2022-07-12 at 9 23 26 AM

### DIFF
--- a/studies/templates/studies/study_detail.html
+++ b/studies/templates/studies/study_detail.html
@@ -270,7 +270,7 @@
                                 </p>
                             </div>
                             <div class="col-xs-12">
-                                {% if study.built %}
+                                {% if study.built or study.study_type.is_external %}
                                     <span>
                                         <label>Study link:</label>
                                     </span>


okaycj:
> Issue #970 also mentioned (in the middle of the text) that the study checkbox should be editable between scheduled & unscheduled for an external study. Has the weird behavior around making those changes also been fixed somehwere else?
> 
> (actual study link change makes perfect sense! Just requesting changes in case further updates are needed for the checkbox issue)

If I am not mistaken, PR #980 should've taken care of this issue.

mekline (approved):

Issue #973, opened 2022-05-31
Rhodri/Sho broken link issue?


mekline:
Seen during the MBAH zoom call 5/30 - 

This page on the Japanese version of the site:
https://lookit.mit.edu/ja/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/

Clicking the blue button (participate now) leads to an XML error:

This XML file does not appear to have any style information associated with it. The document tree is shown below.
<Error>
<Code>AccessDenied</Code>
<Message>Access denied.</Message>
<Details>Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object.</Details>
</Error>

May also appear at this link: 
https://lookit.mit.edu/ja/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/095c1558-56ab-4d1d-94e8-16082e51c300/

okaycj:
@becky-gilbert So it looks like any link that forcefully translated (adding the language to the URL) doesn't work. 

Examples:
- https://lookit.mit.edu/es/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/c958e6dd-0d4d-4e00-b3c0-feabcd5c2533/
- https://lookit.mit.edu/fr/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/c958e6dd-0d4d-4e00-b3c0-feabcd5c2533/
- https://lookit.mit.edu/ko/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/c958e6dd-0d4d-4e00-b3c0-feabcd5c2533/

If you remove the language from the URL, the link works.

Example:
 - https://lookit.mit.edu/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/c958e6dd-0d4d-4e00-b3c0-feabcd5c2533/

Still unsure of the solution.

mekline:
What is the expected behavior here? @rhodricusack might be able to chime in - once a participant clicks "participate" (or relevant translated text) to launch the Lookit study player, are there expected differences between languages?   Another way to ask this would be whether you (Rhodri) have also been working on changes to https://github.com/lookit/ember-lookit-frameplayer

mekline:
If not, then @okaycj @becky-gilbert the expected behavior should be "launch experiment builder with no difference in regard to language, then exit the experiment builder back to a link with the same language setting/URL as before"

[...]

PR #1008, opened 2022-07-29
remove locale from ember study path (fixes #1000)

### LINKED ISSUE #1000: Error in study links generated from translated Lookit site (locale in URL)
**Description**

When a user is on a translated version of the Lookit site (i.e. with a locale in the URL), and they click on the "Participate Now" button, the study page opens but produces an error. See #973. 

**How to reproduce**

1. Go a translated version of the Lookit site and find a study page, e.g. https://lookit.mit.edu/it/studies/c3dc8dc6-ee80-4399-afba-d852225712ae/

2. Select a child and click the (translated) "Participate Now" button.

3. Study page will show the error: "The XML file does not appear to have any style information associated with it. The document tree is shown below." along with console errors, including: "Failed to load resource: the server responded with a status of 403 ()".

The study loads when the locale is removed from the URL that the user is directed to.

**Screenshots**

Screen Shot 2022-07-21 at 1 20 36 PM

Screen Shot 2022-07-21 at 1 20 59 PM

**Expected behavior**

The study should load when a user is on a translated version of the Lookit site and clicks on a "Participate Now" button.

Some might expect the study to not only load but also be automatically translated into the language indicated by the locale in the Lookit site URL. This issue is only intended to fix the study links that the Lookit site generates. The possibilities for dynamically translating studies based on locale will be addressed in a separate issue.

becky-gilbert:
Closes #1000 

Also addresses part of #973.

**Implementation Notes**

I tried using the [revproxy `rewrite` attribute](https://django-revproxy.readthedocs.io/en/stable/proxyview.html#revproxy.views.ProxyView.rewrite), but couldn't get it to work (gives me a "Too many values to unpack" Python error [here](https://github.com/jazzband/django-revproxy/blob/fb323b82e848cc4173aedd38306953c7fac00f57/revproxy/views.py#L55)). I'm pretty sure this error with `rewrite` is also the reason why the `PreviewProxyView` and `ExperimentProxyView` `dispatch` methods already edit the `path` - see this comment:
https://github.com/lookit/lookit-api/blob/8e7c8ea1b1e8fb1e65680b20e0b73529bf581415/exp/views/study.py#L896-L897

**Proposed Solution**

Now, in addition to manually changing the `paths`, the `dispatch` method checks for a locale (language code) in the URL and, if necessary, modifies the request before the request is passed on. ~The method for checking for locale currently just looks to see if `split` returns an additional value before the study UUID. We could use regex instead.~ I switched to regex.

When testing locally, I get the same (expected) error when trying to preview a study with and without a locale in the URL, and the request URLs are the same.

study_preview_no_locale
study_preview_locale

And same thing when trying to participate rather than preview:

study_participate_no_locale
study_participate_locale

I'm not sure how much more testing I can do locally?

**TO DO**
- [ ] On the staging server, test that the following conditions redirect to a functioning ember study:
  - No locale in URL - preview study
  - Locale in URL - preview study
  - No locale in URL - participate in study
  - Locale in URL - particpate in study
- [ ] Write a test for study redirection with and without locale in Lookit site URL?


becky-gilbert:
@okaycj this isn't ready for a proper review yet, but it'd be great to get your thoughts on what to do next (test on staging server?). And I'm not sure yet why the Travis CI check failed.

okaycj:
> And I'm not sure yet why the Travis CI check failed.

Formatting failed.  (You can run `poetry run pre-commit run --all-files` to fix.)

exp/views/study.py
```diff
@@ -905,6 +905,18 @@ def dispatch(self, request, *args, **kwargs):
...
+            In this case we need to add an extra returned string before study_uuid to capture the values correctly,
+            and then re-write the request path so that it points to a working study URL.
+            """
+            if study_uuid == "studies":
+                _, _, exp, studies, study_uuid, child_uuid, _, _, *rest = request.path.split("/")
+                path_no_locale = "/"+"/".join([exp, studies, study_uuid, child_uuid, "preview/"])
+                request.path = path_no_locale
+                request.path_info = path_no_locale
+                request.META['HTTP_REFERER'] = request.META["BASE_URL"] + path_no_locale
+                request.META['PATH_INFO'] = path_no_locale
```
okaycj, line 919: This is bit of a finicky suggestion, but could we remove the duplicate code?

exp/views/study.py
```diff
@@ -905,6 +905,18 @@ def dispatch(self, request, *args, **kwargs):
             return HttpResponseRedirect(get_external_url(study, response))
         else:
             _, _, _, study_uuid, _, _, _, *rest = request.path.split("/")
+            """If locale (language code) is present in the URL, then there will be an extra element
+            returned from split("/") before the study UUID, so the UUID variable will be "studies".
+            In this case we need to add an extra returned string before study_uuid to capture the values correctly,
+            and then re-write the request path so that it points to a working study URL.
+            """
+            if study_uuid == "studies":
```
okaycj, line 913: I know this isn't the code you originally wrote, but `study_uuid` is confusion.  Could we rename this value?

exp/views/study.py
```diff
@@ -905,6 +905,18 @@ def dispatch(self, request, *args, **kwargs):
             return HttpResponseRedirect(get_external_url(study, response))
         else:
             _, _, _, study_uuid, _, _, _, *rest = request.path.split("/")
```
okaycj, line 907: I think I know the answer to this question, but I would love a better solution.  Is there a less odd way to extract the path directory values?  Could the [`pathlib`](https://docs.python.org/3/library/pathlib.html) library help in anyway?  Feel free to push back on this.  I am fairly sure the answer is "Not worth our time".

okaycj (changes requested):
💯 This is a very nice PR.  Thank you!  Just of few comments and we're good to start testing on staging.

okaycj:
@becky-gilbert As a suggestion, you could replace the text in the description "This PR is to fix https://github.com/lookit/lookit-api/issues/1000 and part of https://github.com/lookit/lookit-api/issues/973." With "Closes #1000" and the issue will close when this PR is merged.  You might already know this.

exp/views/study.py
becky-gilbert: 🤦‍♀️  Not finicky! Will do.

exp/views/study.py
becky-gilbert: Absolutely.

becky-gilbert:
> > And I'm not sure yet why the Travis CI check failed.
> 
> Formatting failed. (You can run `poetry run pre-commit run --all-files` to fix.)

Thanks! (And sorry - I thought that pre-commit formatting was done automatically)

exp/views/study.py
becky-gilbert: Ah, so looking at this again - I think the `study_uuid` variable name makes sense here, because when there's no locale in the URL, this variable _is_ the study UUID:
```python
_, _, _, study_uuid, _, _, _, *rest = request.path.split("/")
```
But if there is a locale in the URL, then splitting the path returns an extra value that shifts the value-index mapping by 1. Which is why I check here to see if `study_uuid` is actually "studies".

Maybe instead of splitting the path on "/" here, we should use regex to get the locale (if it exists) and study UUID. Thoughts?

exp/views/study.py
becky-gilbert: I had a look at pathLib and that looks like it mostly exists to deal with OS-specific file path issues. It does have the [`parts` method](https://docs.python.org/3/library/pathlib.html#accessing-individual-parts), but that looks like it'll be doing the same thing as `.split("/")` in our case. Let me know if you think there's something I missed.

I'm now thinking regex is the best way to go. I could check whether the request path matches the pattern <root>/studies/<study_UUID>/... or <root>/<locale>/studies/<study_UUID>/..., and do the rest of the manipulation based on that.

exp/views/study.py
becky-gilbert: @okaycj I switched from this `split` method to regex, so there's no longer a variable called `study_uuid` that might _not_ be the study UUID. So I'll mark this one as resolved.

exp/views/study.py
becky-gilbert: @okaycj I switched to regex to check the path structure and extract values. Can you please review?

exp/views/study.py
becky-gilbert: @okaycj I removed the change to `HTTP_REFERER` because it was uncessary. But I think the changes to the request `path`, `path_info` and `META.PATH_INFO` are all necessary. (It's hard to know for sure because the study links produce an error in development anyway - I think I'll have to test on staging).

So I don't think there's code duplication here, unless I've misunderstood?

exp/views/study.py
```diff
@@ -897,42 +898,30 @@ def dispatch(self, request, *args, **kwargs):
...
-                    _,
-                    exp,
-                    studies,
-                    study_uuid,
-                    child_uuid,
-                    _,
-                    _,
-                    *rest,
-                ) = request.path.split("/")
+            locale_pattern = r"/(?P<locale>[a-zA-Z-].+)/exp/studies/"+str(study_uuid)+"/"+str(child_uuid)+"/preview/(?P<rest>.*?)"
```
okaycj: F-Strings might clean this up a bit.

okaycj (changes requested):
👍 This is fairly perfect.  One suggestion and I think we're all done here.  Thank you.

exp/views/study.py
becky-gilbert: Agree! I switched to f-strings for the regex patterns and re-constructing the URLs, and I think it's more readable now. Resolved?

becky-gilbert:
@okaycj Let me know if you have any other suggestions. Otherwise I think this is ready for testing on staging, but I'll wait for you to merge (since you wanted to merge other stuff first, I think).

@mekline @okaycj also I haven't forgotten about writing tests. Normally I'd want to include the tests with the PR containing relevant code change, but I haven't written django tests before and it might take me a while to figure out. If we're in a hurry to fix the broken links then I could write the tests separately so as to not hold things up (plus the tests are mostly to catch potential problems down the line, e.g. changes to hard-coded parts of the paths). Regardless, I'll manually test everything on the staging server.

mekline:
@becky-gilbert holding the tests for after this release would be great, just so we can get people moving again!

becky-gilbert:
Ok thanks @mekline!

okaycj (approved):
💯 Let's get this out!  Good job and thank you again.

PR #1011, opened 2022-08-04
Submit form declarations

### LINKED ISSUE #996: Add declarations to submit form
# TL;DR
To make reviewing submitted studies less of a load, we're going to add a few values to the submit form were the experimenter can declare elements we'd want to review. 

# Narrative
Adding declarations would allow reviewers a hint at what to expect when looking through submitted studies.  The existing submit form has a text box that should remain.  Underneath, there will be additional check boxes that will allow the experimenter to declare aspects of their study reviewers would want to take a close look at.  

# Acceptance Criteria
- [ ] Submit form has updated copy
- [ ] Submit form has updated fields
- [ ] When Submit form is submitted, values are saved to database. 

# Implementation Notes
Study log doesn't currently operate as a traditional Django form.  There is a call to the ORM to create an entry in the database.  This pattern is repeated a few times throughout the code, but for an example see [here](https://github.com/lookit/lookit-api/blob/develop/studies/models.py#L760).  

So, as to not stray too far from the existing pattern, we can save the above checkbox values to the jsonp field "extra" using the following key values:

- peer_review_name,
- changes_description,
- issue_deception,
- issue_length,
- issue_consent,
- issue_externaldata,
- issue_sensitive,
- issues_description

The copy for the form should be:
```
# Are you sure you want to change the status of [STUDY NAME]?

Please list the researchers outside your group who have provided feedback on your study prior to submission, and the changes you made in response to peer feedback. (If this is a revision to an active study, list new changes made after approval):

[Larger textbox, the size of the current solo textbox]

Please note here any elements of your study that require additional review (see Terms of Use):

[ ] Deception/withholding of information

[ ] Sessions longer than 30 minutes

[ ] Departures from the standard Lookit template consent form or recorded confirmation statement

[ ] Incorporation of participant information from other sources (e.g. a school ID number)

[ ] Collection of personally identifying information such as email addresses or zip codes

[ ] Sensitive topics or questions (e.g. political subjects, race, bullying, mental health)

If you checked any of the boxes above, please describe below:

[Larger textbox, the size of the current solo textbox]

```

Created from #992

okaycj:
This PR covers adding declarations to the study submit modal, recording the declarations in the study log, and showing the declarations in the UI. 

Closes #996

declarations


becky-gilbert (approved):
Looks great to me! 
Sorry that I don't have any specific feedback on the implementation. I did look through the code and the parts that I understand all look very sensible. And I appreciate the clean-up work, like moving the study detail JS into a separate file and fixing the HTML problems 👍

PR #1014, opened 2022-08-11
Declarations in study comments

### LINKED ISSUE #1009: Add text to comment field on submit form action
When the new submit form is submitted (#996) , we are now saving some new fields that can be used for future differentiation of reviewing.  In addition, we should add text to the existing Comment field so that something sensible appears in the study log and "comments" box. 

The text should read as follows: 

```
`changes_description`
```

When non of the issues_XXX fields are TRUE (aka, very similar to the current behavior!). When at least on of those issue checkboxes is checked, it should read:

```
`changes_description`

Potential issues:  `identifyinginfo, consent, sensitive` #i.e. a list of the 'issues' fields set to TRUE

`issues_description`
```

Image

okaycj:
This adds declaration to the comments box in the study details page. 

Closes #1009

comments

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -447,7 +447,7 @@ class StudyDetailView(
             comments_text.append(
                 f"It is not possible to submit or start this study until the lab {study.lab.name} is approved to test."
             )
-        else:
+        elif study.comments is not None:
             comments_text.append(study.comments)
 
         # if study is submitted, see if there are any declarations to display
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -394,7 +394,7 @@ class StudyDetailView(
         """
         context = super(StudyDetailView, self).get_context_data(**kwargs)
 
-        study = context["study"]
+        study: Study = context["study"]
         admin_group = study.admin_group
 
         context["triggers"] = get_permitted_triggers(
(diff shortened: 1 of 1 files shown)


becky-gilbert:
Looks great! Just to clarify a few things (for my own benefit): 
I think the study's `comments` and `comments_extra` (which includes `declarations` and `issues_description`) are all managed separately, and this `comments` method just controls the display of text in a "comments" box on the study page. Is that right?
Does the `comments_text` (or some version of the same info) appear in the study log as well? I'm wondering because it seems like this comments box will contain just the latest study comments (which makes sense), but it'd probably be useful for reviewers to see the whole history of declarations/explanations that went along with any prior submissions.

mekline (approved):

okaycj:
@mekline 

> I think the study's `comments` and `comments_extra` (which includes `declarations` and `issues_description`) are all managed separately, and this `comments` method just controls the display of text in a "comments" box on the study page. Is that right?

Yes.  Let me know if you'd like more detail.

> Does the `comments_text` (or some version of the same info) appear in the study log as well? I'm wondering because it seems like this comments box will contain just the latest study comments (which makes sense), but it'd probably be useful for reviewers to see the whole history of declarations/explanations that went along with any prior submissions.

Yes.  I will appear after the list of declarations.  I'll put a link to the previous PR (#1011) so that you may review.  Let me know if the GIF isn't clear.  I'd be happy to demonstrate if needed.

becky-gilbert:
> I'll put a link to the previous PR (https://github.com/lookit/lookit-api/pull/1011) so that you may review. Let me know if the GIF isn't clear. I'd be happy to demonstrate if needed.

Ah yes, that PR answers my question. Thanks!

becky-gilbert (approved):
🚀

PR #1016, opened 2022-08-15
Remove locale from study links by manually setting language

becky-gilbert:
## Background
This is to address the issue of broken study links when the user starts the session from the Lookit website that contains a locale code in the path: #973, #1000.

I tried manually removing the locale from the request path for study sessions in this PR: #1008. This does change the request path correclty, but when testing on the staging server, it didn't work: the locale is added back into the path. 

I think the Locale Middleware is altering the request _after_ it's been modified by the ProxyView dispatch methods ([see here](https://docs.djangoproject.com/en/4.1/topics/http/middleware/)). If so then the fix will probably need to involve Django's translation tools and/or the Locale Middleware.

## Current change
This PR is to see if it's possible to remove the locale from the study session URLs by manually setting the request's language code and [activating](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.translation.activate) the default translation language.

If this works, then I will try removing the earlier modifications to the request path from #1008.

If this doesn't work, then the next step might be trying to figure out how to make URL pattern exceptions with [`i18n_patterns`](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#language-prefix-in-url-patterns). I haven't been able to figure that out yet and I'm not sure if it's possible. It does look like it's possible to re-write/customize the `i18n_patterns` function: https://stackoverflow.com/a/54786316.

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -928,9 +928,9 @@ class PreviewProxyView(ResearcherLoginRequiredMixin, UserPassesTestMixin, ProxyV
         path replacement manually. Great! Just wonderful.
         """
 
-        """Manually set the language code to the default language, which isn't prefixed in URLs, 
-        to stop the locale middleware from adding a language/locale prefix back into the study URL 
-        after the request has been dispatched"""
+        # Manually set the language code to the default language, which isn't prefixed in URLs,
+        # to stop the locale middleware from adding a language/locale prefix back into the study URL
+        # after the request has been dispatched
         translation.activate(settings.LANGUAGE_CODE)
         request.LANGUAGE_CODE = settings.LANGUAGE_CODE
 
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Not sure why this PR is listing my earlier commits - I obviously did something wrong! 😣 
Maybe have a look at the The Files Changed tab rather than going through all of the commits. Or just this one: [e171135](https://github.com/lookit/lookit-api/pull/1016/commits/e171135a99c1ff2b42427b38d69ad136f42cb7c2).

okaycj:
> Not sure why this PR is listing my earlier commits - I obviously did something wrong! 😣 Maybe have a look at the The Files Changed tab rather than going through all of the commits. Or just this one: [e171135](https://github.com/lookit/lookit-api/pull/1016/commits/e171135a99c1ff2b42427b38d69ad136f42cb7c2).

Try rebasing this PR against develop.

becky-gilbert:
> Try rebasing this PR against develop.

That did it, thanks! I did have to force push but don't think I broke/lost anything. Won't make a habit of it!

mekline (approved):
Test away!

okaycj:
> That did it, thanks! I did have to force push but don't think I broke/lost anything. Won't make a habit of it!

Rebasing means you'll have to force push.  When it comes to PRs, force pushing is always fine.  I do it constantly.

becky-gilbert:
> Rebasing means you'll have to force push. When it comes to PRs, force pushing is always fine. I do it constantly.

Ah ok, thanks! Yes I'd only googled quickly and saw some scary warnings about rebasing when the work has been pushed to a public repo. But after reading a little more, it seems like it's only a problem if the branch/PR is potentially being used/edited by someone else.

PR #1017, opened 2022-08-19
Update to reviewing tab

okaycj:
Updates to experimenter study list view. 

Closes #991 
Closes #863
Closes #612

PR #1018, opened 2022-08-30
Fix study links with locale

becky-gilbert:
This is an attempt at #973 and #1000. This is just for testing on staging, since I'm not able to test this locally.

After my last attempt to fix this issue on develop (#1016), I found that, when the URL contains a locale, clicking the study link _does_ load the ember study, but then it produces a 'Page not found' error before making some of the requests that originate from the study JS (specifically the vendor.js file, I think). The requests that aren't made are those starting with `/api/v1/`.

Normal sequence of requests (i.e. no locale in the URL) are shown below. Requests with the red box around them are absent when there's a locale in the Lookit site URL:

staging_no_locale_first_request_after_404

Page content and requests with a locale in the Lookit site URL:

staging_ember_request_locale_404

So the problem might be with the creation of paths to the study asset files? 

I'm going to merge this without review so that I can test ASAP. But please feel free to provide thoughts/feedback anyway.

BTW I added the fix for the study declarations key error (totally separate issue), just because this was causing problems for me. Happy to fix the error in some other way, I just needed a temporary solution.

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -452,7 +452,8 @@ class StudyDetailView(
             comments_text.append(study.comments)
 
         # if study is submitted, see if there are any declarations to display
-        if study.state == "submitted":
+        declarations_exist = study.comments_extra.get("declarations")
+        if study.state == "submitted" and declarations_exist:
 
             declarations_state = ", ".join(
                 DECLARATIONS["submit"][k]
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Update: I've reverted my attempts to fix the errors related to declarations and comments_extra being None, since CJ is fixing this on a separate branch. I'll merge this into develop after #1017 is merged.

becky-gilbert:
Changed my mind, I don't think I need to wait for the bug fix from CJ's branch before testing this. So I'll merge now so that I can test ASAP.

PR #1023, opened 2022-09-02
Declarations in study submit and comments; fix study links with locale

becky-gilbert:
This PR is to merge `develop` into `master`. 

_Note: will edit this to add closing keywords later, once I know which issues should close vs remain open._

The `develop` branch contains:
1. These merged PRs that have already been reviewed:
- #1011 
- #1008 (portions of this change were removed in a later commit)
- #1015 
- #1014 

2. These merged PRs that were not reviewed, but that contain changes that were removed in a later commit (because they did not fix the locale problem on staging):
- #1016 
- #1018 

3. This merged PR and commit, which still need to be reviewed:
- #1020 
- https://github.com/lookit/lookit-api/commit/72ad0b6063f090361fda1dfb3313f64dcebe7db2 (direct push rather than PR to increase the google cloud build timeout)

So it might be easiest to review the final file change diffs, rather than reviewing individual commits, since some code changes were made but later removed.


exp/views/study.py
```diff
@@ -927,6 +928,12 @@ def dispatch(self, request, *args, **kwargs):
         path replacement manually. Great! Just wonderful.
         """
 
+        # Manually set the language code to the default language, which isn't prefixed in URLs,
+        # to stop the locale middleware from adding a language/locale prefix back into the study URL
+        # after the request has been dispatched
+        translation.activate(settings.LANGUAGE_CODE)
+        request.LANGUAGE_CODE = settings.LANGUAGE_CODE
+
```
okaycj: Such a good idea.  Nice work!

web/views.py
```diff
@@ -611,9 +611,28 @@ class ExperimentAssetsProxyView(LoginRequiredMixin, ProxyView):
...
+        # Check if locale (language code) is present in the URL.
+        # If so, we need to re-write the request path without the locale
+        # so that it points to a working study asset URL.
+        locale_pattern = rf"/(?P<locale>[a-zA-Z-].+)/studies/{uuid}/(?P<rest>.*)$"
+        path_match = re.match(locale_pattern, request.path)
+        if path_match:
+            path_no_locale = rf"/studies/{uuid}/{path_match.group('rest')}"
+            request.path = path_no_locale
+            request.path_info = path_no_locale
+            request.META["PATH_INFO"] = path_no_locale
```
okaycj: If the language is set to the default language, do we still need to check if there's a language code in the URL?

web/views.py
```diff
@@ -680,34 +662,27 @@ def dispatch(self, request, *args, **kwargs):
         path replacement manually.
         """
 
-        # Manually set the language code to the default language, which isn't prefixed in URLs,
-        # to stop the locale middleware from adding a language/locale prefix back into the study URL
-        # after the request has been dispatched
-        translation.activate(settings.LANGUAGE_CODE)
-        request.LANGUAGE_CODE = settings.LANGUAGE_CODE
-
```
okaycj: I presume that by removing this logic it wasn't solving our problem.  Which is a shame.

web/views.py
```diff
@@ -680,34 +662,27 @@ def dispatch(self, request, *args, **kwargs):
...
-            debug_path = rf"{settings.EXPERIMENT_BASE_URL}{request.path_info}"
-            return redirect(debug_path)
-        else:
-            path = f"{study_uuid}/index.html"
-            return super().dispatch(request, path)
+            url = f"{settings.EXPERIMENT_BASE_URL}{path}"
+            return redirect(url)
+
+        path = f"{study_uuid}/index.html"
+        return super().dispatch(request, path)
```
okaycj, line 688: I don't think we should change it or put any time in researching, but do you know why the "index.html" needs to be in the URL?

cloudbuild.yaml
```diff
@@ -147,4 +147,4 @@ secrets:
   secretEnv:
     SENTRY_AUTH_TOKEN: CiQAVYLHlum+JbxvLUXADsLiiIwXjocXFGg5SAsxKyvAkDRjEaESaABnxmN4TReBWF0cY+XPS8xytlGSUhClZewwH7ttpBFKC65+Y0a6W8XIsTjyQMyVzWiLxSOef46bmI1Bo8JIbh/ErIRWm3rPBNFs843z52oF8M93+miMUlt8UOWmIgyq1U/fjThhq0UI
 
-timeout: 1200s
+timeout: 1800s
```
okaycj, line 150: Do you know why it is necessary to increase the timeout?

okaycj (approved):
This all works.  I am super excited to see this get moved into production.  Just a few questions and then let's get this merged!

web/views.py
becky-gilbert: Yep, doesn't work and I still don't know why. It would've been such a nice solution.

cloudbuild.yaml
becky-gilbert: No I don't know why. The two commits that timed out were 4c911de and f2f14ff. For each of the longer builds, I looked at the timing of each step and it looks like 3 of the steps took about twice as long as to complete they normally do: build-image, build-test-image, and run-tests. I have no idea why the changes I made would cause this. Interestingly, the build time went back to normal after your last merge into develop (97e690c), so maybe it was a temporary thing? We could revert this back to 1200s if future builds continue to stay under 20 minutes.

web/views.py
becky-gilbert: No, sorry.

becky-gilbert:
@okaycj @mekline Is this ready to merge?

PR #1035, opened 2022-10-31
Move node version to 12 for EFP

okaycj:
The docker file for building experiments is held in this repo.  We needed the version of node increased from 10 to 12 to support AWS SDK.


becky-gilbert (approved):
Great, thanks! After merging, do we just need to test the study build on staging? Let me know how I can help 👍

PR #1041, opened 2022-11-22
Filter participants by uuid as well as nickname

### LINKED ISSUE #919: Filter participants by global ID
**Description**
As a researcher, I search for individual participants from the participants tab (https://lookit.mit.edu/exp/participants). I can search by nickname, but these often contain users' real names, so I don't store this field in our local data. I want to search by user global ID instead.

**How to reproduce**
From the participants tab, copy and paste any of the global IDs on the first page of results into the "Filter by nickname" box. This returns the "no results found" text.

**Expected behavior**
A global ID or partial global ID string entered into the Filter box should return any Users with matching ID, displaying as it currently does with nickname, global ID, and last active date.

**Additional context**
Reported/requested by Liz L.

okaycj:
On experiment participant's page, filter by participant's UUID as well as their nickname. 

Closes #919

### DIFF
--- a/accounts/templates/accounts/participant_list.html
+++ b/accounts/templates/accounts/participant_list.html
@@ -28,7 +28,7 @@
                     <input id="search-participants"
                            class="form-control"
                            name="match"
-                           placeholder="Filter by nickname"
+                           placeholder="Filter by nickname or user global ID"
                            size="50"
                            type="text"/>
                     <input type="hidden" name="sort" value="{{ sort }}"/>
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Hi @okaycj, I just tested this and it works ✅ 

One minor suggestion: we could modify the placeholder text in the search bar to make it clear that the user can search by UUID as well as nickname. 

Here's how it looks now:
Screenshot 2022-11-23 at 9 58 47 AM

Proposed change:
Screenshot 2022-11-23 at 10 04 21 AM

If you're happy with this then I'll commit it to this branch.

becky-gilbert (approved):

PR #1042, opened 2022-11-28
Proxy redirect auth

okaycj:
We recently made a change were we put in a redirect to an experiment.  This redirect circumvented the Django middle-ware and therefore didn't authenticate the proxy view before a user would see a study.  This PR should fix this issue.


exp/views/study.py
```diff
@@ -972,14 +979,16 @@ def user_can_view_previews(self):
         except Study.DoesNotExist:
             return False
 
+        print(child.user, request.user)
```
becky-gilbert, line 982: Just checking that you meant to leave this in. Or was it just for testing?

becky-gilbert (approved):

becky-gilbert:
Thanks for catching this issue @okaycj - looks good to me! 
Am I right in thinking that I can't test that this solution works locally because of the way that the site redirects to the local Ember server based on `DEBUG` and `ENVIRONMENT` settings? That is, when running locally I can always visit an Ember study page even when I'm not logged in to the Lookit site, right? If so I'd be happy to test this once it's on staging.

exp/views/study.py
okaycj: Whoops.  I'll put in a change for that.

PR #1044, opened 2022-12-05
Study priority

### LINKED ISSUE #1024: Manual Study Order Feature for lab home pages
# TL;DR
CRADLE: Create a feature that allows labs to manually order their studies on the lab home page created for #620. 

# Narrative

From Ian: It will be helpful to visually show participants the suggested order for study completion. We will also need this feature to work with #625's eligibility criteria modification, as we want this order to adapt to remaining studies for the particular participant. 

As a researcher, I want to be able to assign priority to individual studies in my lab, rather than accepting the random/hashed ordering that's automatically produced on the Search and Lab pages. 

# Acceptance Criteria

- [x] New field appears on Edit Study page allowing a 'priority' integer between 1 and 99 to be assigned. This corresponds to a new variable on a study object called 'study_queue_priority.' 
- [x] By default, all studies (including existing studies) are given a priority of 1 (highest) - the logic here is that, if a lab is not ranking studies, a nefarious lab member would have to modify someone *elses* study, rather than their own, to try and sneakily jump ahead of the other studies. 
- [x] On Search and Lab pages, studies continue to be displayed according to whether they are discoverable, and if the user selects from the drop downs, according to if a particular child is eligible, if the study takes place on or off Lookit, etc. 
- [x] On the main Search page, sorting continues to be by hash only (status quo). 
- [x] On the Lab page, sorting is first by priority, and then by hash.  This means that if a set of studies-to-be-shown contains three with priority 1 and two with priority 10, then they would be displayed as [priority 1 studies ordered by hash]-[priority 10 studies ordered by hash]

# Implementation Notes

It's likely that accomplishing the final 2 criteria will require modifying existing sorting code, and we should proceed with the idea that future changes to sorting could come up in the future - so for instance this might involve defining a set of rankings across both of these page views, where the Search page is set to ["Sort by hash = first", "Sort by lab priority = "no"], and the Lab page is set to ["Sort by lab priority = first", "Sort by hash = second"]

okaycj:
Add a 'Priority' field to study table and then sort the studies on the lab pages by first this value and then by uuid/hash.  

Closes #1024


becky-gilbert (commented):
Looks great! I tested locally and the studies are being ordered (and re-ordered) correctly based on assigned priority 👍 
I wonder if it would help to show the assigned priority value next to the slider, to make it easier for researchers to ensure that they're prioritizing studies correctly? (Might be easier to compare numbers versus fine-grained positions on a slider). Happy to do this if you want, and I could always implement it later. I'm not sure if this can be done just via `forms.py`, but I could write a JS function in `_study_type.html`.
Also, do you know why the Travis CI build failed? I had a quick look but I'm not sure. Looks like the first error is `fatal: early EOF`?

mekline:
I agree on showing the current value!  It would be fine to have this just listed as a number in the form rather than as part of the slider element, if that's easier.  

We also need help text to explain what this is!! So all together, it might look like this: 

#### Priority
Current priority: 1
`---------slider----------`
*This affects how studies are shown at your lab's custom URL, not the main study page. If you leave all studies at priority 1, all of your lab's active/discoverable studies will be shown in a randomized order for each participant. If you set the priority of this study to 100, then it will appear last in the list. You can find your lab's custom URL from the [labs page](https://lookit.mit.edu/exp/labs/).*

mekline:
Note this new feature incurs some documentation debt!  Added a corresponding issue on the docs repo: https://github.com/lookit/lookit-docs/issues/277

becky-gilbert:
> Priority
Current priority: 1
---------slider----------
This affects how studies are shown at your lab's custom URL, not the main study page. If you leave all studies at priority 1, all of your lab's active/discoverable studies will be shown in a randomized order. If you set the priority of this study to 100, then it will appear last in the list. You can find your lab's custom URL from the [labs page](https://lookit.mit.edu/exp/labs/).

I have a few suggested changes to Melissa's suggestion for the form text above. Feel free to ignore if you disagree!

1. Even though it's in Melissa's explanation already ,we might want to emphasize that this value only affects the lab page and not the main study page, by e.g. renaming the header "Lab Page Priority", adding "on your lab page" to the explanation text. 

2. It might help to have labels on each end of the slider, e.g. "Highest (show study first)" and "Lowest (show study last)".

3. We could add a link to the documentation about this in the explanation blurb, since there are some nuances to this feature that don't really make sense to cover in detail here.

So perhaps something like this:

**Lab Page Priority**
Current priority: 1
    ---------slider----------
Highest &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Lowest
This affects how studies are ordered at your lab's custom URL, not the main study page. If you leave all studies at priority 1, all of your lab's active/discoverable studies will be shown in a randomized order on your lab page. If you set the priority of this study to 100, then it will appear last in the list on your lab page. You can find your lab's custom URL from the [labs page](https://lookit.mit.edu/exp/labs/). For more info, see the documentation on study prioritization (\<link here\>).

studies/forms.py
```diff
@@ -322,6 +327,7 @@ class Meta:
                 For more information on experiment runner types, please
                 <a href={STUDY_TYPE_HELP_LINK}>see the documentation.</a></p>""",
             "structure": PROTOCOL_HELP_TEXT_INITIAL,
+            "priority": f"This affects how studies are ordered at your lab's custom URL, not the main study page. If you leave all studies at priority 1, all of your lab's active/discoverable studies will be shown in a randomized order on your lab page. If you set the priority of this study to 100, then it will appear last in the list on your lab page. You can find your lab's custom URL from the <a href='/exp/labs/'>labs page</a>. For more info, see the documentation on study prioritization (link here).",
```
becky-gilbert, line 330: There are some discrepancies between the text here and the UI. I'm not sure which is correct but we should fix one or the other.
- The text here says "If you leave all studies at priority 1..." but it looks like the default value is 99.
- The text here says "If you set the priority of this study to 100..." but it looks like the max value is 99.

becky-gilbert (commented):
I just flagged some discrepancies between the help text below the priority slider and the UI/settings. This probably happened because my suggestions for the help text weren't in line with Melissa's specs for how this should work. Sorry!

studies/forms.py
okaycj: Yup, your correct.  I had made a note to myself in the description, but it didn't include the default value.  I'll update both those values.

becky-gilbert:
Just a reminder that we should make sure that whatever we do here lines up with the corresponding changes in the docs about ordering studies on the lab pages: https://github.com/lookit/lookit-docs/pull/282

Specifically, I wrote:

> Setting the study's priority value to 1 (the default) will mean it is shown *first* on your page, while setting the priority to 99 will mean it is shown *last*. You can use any integers between these values for more fine-grained control over ordering. If any of your lab's studies share the same priority value, then presentation order will be random within that set.

Let me know if that's not how it actually works and I'll edit the docs!

studies/forms.py
```diff
@@ -322,6 +327,7 @@ class Meta:
                 For more information on experiment runner types, please
                 <a href={STUDY_TYPE_HELP_LINK}>see the documentation.</a></p>""",
             "structure": PROTOCOL_HELP_TEXT_INITIAL,
+            "priority": f"This affects how studies are ordered at your lab's custom URL, not the main study page. If you leave all studies at priority 99, all of your lab's active/discoverable studies will be shown in a randomized order on your lab page. If you set the priority of this study to 1, then it will appear last in the list on your lab page. You can find your lab's custom URL from the <a href='/exp/labs/'>labs page</a>. For more info, see the documentation on <a href='https://lookit.readthedocs.io/en/develop/researchers-manage-org.html#ordering-studies-on-your-lab-page'>study prioritization</a>.",
```
becky-gilbert, line 330: Looks great, thanks for adding the links.
I think there's still a discrepancy between this part of the form and the way that it actually works. Specifically I think the 'Highest' and 'Lowest' labels on the slider need to be the other way around, so that the left is labelled 'Lowest' and the right is labelled 'Highest'.

And to make the mapping between values and priority a little clearer, I wonder if we can edit the text like this:

> If you leave all studies at the highest priority (99), then all...

and

> If you lower the priority of this study to 1, then...

studies/templates/studies/_study_fields.html
```diff
@@ -191,11 +191,41 @@
...
+
+        // Add current priority element
+        currentPriority.id = 'current-priority';
+        priority_parent.insertBefore(currentPriority, priority);
+
+        // Add priority labels
+        highest.innerHTML = 'Highest';
+        lowest.innerHTML = 'Lowest';
+        lowest.classList.add('priority-lowest');
+        priority_parent.insertBefore(highest, help_block);
```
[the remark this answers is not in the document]
okaycj: That's a bit embarrassing....

PR #1053, opened 2023-01-04
Add hewbrew to languages

### LINKED ISSUE #1052: Add Hebrew to list of languages
On a family request, please add Hebrew to the list of language options anywhere it appears in the demographic form (I believe both the Family and Child pages...)

okaycj:
This add Hebrew to the list of languages for a child. 

Closes #1052

Screenshot 2023-01-04 at 12 37 44 PM

### DIFF
--- a/studies/fields.py
+++ b/studies/fields.py
@@ -69,6 +69,7 @@ LANGUAGES = (
     ("gu", _("Gujarati")),
     ("hak", _("Hakka")),
     ("ha", _("Hausa")),
+    ("iw", _("Hebrew")),
     ("hi", _("Hindi")),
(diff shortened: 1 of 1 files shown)


becky-gilbert (approved):
Looks good 👍 
This probably doesn't matter, but [this wikipedia list](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) says that the ISO code for Hebrew has changed from 'iw' to 'he'. And it looks like django also uses 'he': https://github.com/lookit/lookit-api/blob/e5a398d0655d9321d4566a0615c3b5248dc4c7be/project/settings.py#L226

PR #1063, opened 2023-01-19
Image size validation

### LINKED ISSUE #1038: Pad study (and lab) images

okaycj:
Image now need to be square.  To enforce this,  a validation was added to the form after submit.  

Closes #1038


mekline:
This will fix things going forward, but not existing images, right?  Two ways I see to address that: you could *also* add padding to the images where they appear (css or some bootstrap property? I don't know...), or we could do a one-time amendment of the images on existing live studies. 

The latter would be a pain, but I would approach it by giving people a ~1wk window to do this themselves, then amending just what's left.  (For minimal back-end pain, we could do this by exercising super-user privileges to simply re-upload pictures that passed the new validation). 

Since the former would require more work at the end of an already tight shape I might vote for the latter solution, with the caveat that the upcoming transitions to mobile might require taking another look at this fairly soon...

@okaycj if you are good with the 'manually update existing images' plan I can write to to the userbase today to get this started.

okaycj:
> @okaycj if you are good with the 'manually update existing images' plan I can write to to the userbase today to get this started.

I think this is the best solution.  If we were to edit people images for them, they could be cropped in unkind ways.

becky-gilbert (approved):
Looks and works great!
I have one minor suggestion - an edit to the image help text on `studies/forms.py` (line 316):
> "image": "This is the image participants will see when browsing studies. Please make sure that your image file dimensions are square and the size is less than 1 MB.",

PR #1072, opened 2023-01-25
Replace multiselectfield

okaycj:
To get ready to update Django (#1051), the django-multiselectfield package needs to be removed.  By all accounts, the [library](https://github.com/goinnn/django-multiselectfield) hasn't been updated since the beginning of 2021.  

In order to replace the multiselect field with a bit field, the string values had to be compared to the list of choices and get the index.  The index is masked and then or'd with all selections for that field.  

The US Ethnicity field is the only field that needed to be migrated away from multiselect. 

Screenshot 2023-01-25 at 2 44 21 PM


mekline:
Can you confirm that other checkbox choices, like the big list of languages people can choose when creating/modifying a Child, are not affected by this? Just wasn't clear from what you wrote above.

mekline (commented):
Can you confirm that other checkbox choices, like the big list of languages people can choose when creating/modifying a Child, are not affected by this? Just wasn't clear from what you wrote above.

becky-gilbert:
> Can you confirm that other checkbox choices, like the big list of languages people can choose when creating/modifying a Child, are not affected by this? Just wasn't clear from what you wrote above.

@mekline I can confirm that the US ethnicity question is the only one that uses `multiselectfield`. The other multi-select questions, like child languages and existing conditions, are already using the `BitField` type that CJ is switching to here.

@okaycj happy to approve this pending the fix to circleci build-and-test 🙂 

Also I noticed that the question text that is currently live on staging/develop looks different to that in your screenshot above:

Screenshot 2023-04-25 at 10 23 59 AM

Note the slight difference in question text: 
> What racial group(s) does your family identify with/belong to? Please select any that apply to someone in your children's immediate family.

And the help text below the question, which is missing in the screenshot above:
> Please select the appropriate responses for everyone in your children's immediate family.

But it doesn't look like you've changed/removed the question or help text in this PR, so maybe the screenshot above (from Jan 25th) is just out of date?

okaycj:
This issue seems to have been resolved by the Multiselectfield's version 0.1.13.  Closing PR.

PR #1073, opened 2023-01-26
Update google tag code

### LINKED ISSUE #1056: Implement ad tracking snippet
# TL;DR

We minimize the amount of tracking we do on Lookit, but it appears to be no longer possible to set up a google ad campaign without some form of 'conversion metric' that depends on behavior on the site you direct people to.  So, to make the new ad campaign work, we need to insert the snippet below into those URLs that will be tracked by the google ad campaign. 

# Acceptance Criteria

- [ ] Prior to implementation, write up a short (1-2 paragraph MAX) description of what the code does, and how it interacts with a participant's computer. Edit this issue so we have a record of this in case we need to roll back, or in case anyone asks about this.
- [ ] Add the code snippet in the <head> of the following URLS (which are the plausible landing page or endpoint for conversion tracking):

- https://lookit.mit.edu/
- https://lookit.mit.edu/signup/
- https://lookit.mit.edu/account/demographics/
- https://lookit.mit.edu/account/children/
- https://lookit.mit.edu/studies/

# Implementation Notes
Here's the code snippet: 

```
<!-- Google tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-XMVJEVKPVV">
</script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'G-XMVJEVKPVV');
</script>
```

okaycj:
Update to google manager tag code.

### DIFF
--- a/project/settings.py
+++ b/project/settings.py
@@ -15,6 +15,8 @@ from pathlib import Path
 
 from django.contrib.messages import constants as messages
 
+GOOGLE_TAG_MANAGER_ID = os.environ.get("GOOGLE_TAG_MANAGER_ID", "")
+
 ENVIRONMENT = os.environ.get("ENVIRONMENT", "develop")
 
 # Build paths inside the project as str(Path(BASE_DIR, ...).resolve()) or os.path.join(BASE_DIR, ...).
--- a/web/templatetags/web_extras.py
+++ b/web/templatetags/web_extras.py
@@ -1,4 +1,3 @@
-import os
 import textwrap
 from typing import Text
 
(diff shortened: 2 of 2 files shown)


mekline:
As far as I can tell this is just what we need - the google ad interface provides a way to test the setup, but I think will require me to keep using the main lookit.mit.edu domain. Can you move this into production so I can try that out?

mekline (approved):

okaycj:
> As far as I can tell this is just what we need - the google ad interface provides a way to test the setup, but I think will require me to keep using the main lookit.mit.edu domain. Can you move this into production so I can try that out?

Absolutely,  I am currently working on updating orchestrator to get the Google tag ID into the environment variables.

PR #1076, opened 2023-01-27
Add js to fix selection of participation fields

okaycj:
Paticipation fields are mutli selection fields.  The default method in this type of field forces the user to use the command button + mouse click to unselect a study.  This PR adds JS to allow the user to toggle selections with a mouse click only.

Additionally, This PR adds a list of selected studies below each of the Participation fields.  

out


mekline (approved):

becky-gilbert (commented):
Works perfectly! 👍

PR #1098, opened 2023-02-09
Bootstrap 5 home

okaycj:
This PR is for the BS5 updates to the home page.  

Screenshot 2023-02-09 at 11 43 18 AM
Screenshot 2023-02-09 at 11 43 25 AM


becky-gilbert:
Looks great!!
At the risk of being annoying.... would it be possible to review the PRs for merging different bits of the site (login, home etc.) into a bootstrap 5 base branch (rather than develop), so that it's a little easier to see the changes that are specific to the view that you're working on? Unless that totally messes up your workflow.

becky-gilbert (approved):

becky-gilbert:
Oh one minor comment: I kind of like the colors (gray rather than blue) and grouping/spacing of the nav bar items in the current version:

Screenshot 2023-02-09 at 4 08 00 PM

Screenshot 2023-02-09 at 4 06 53 PM

Though if it were up to me, the nav bar items would be in three evenly-spaced groups, like this:
>        Home    |    Studies   FAQ   The Scientists   Resources    |    Sign Up   Login    

>     Home  Experimenter    |    Studies  FAQ  The Scientists  Resources     |    My Account  My Past Studies  Logout

(Ok so it's hard to get the spacing right in the text above, but the main idea would be to center the Studies-Resources group between the left-most and right-most items.)

mekline:
I agree with Becky about the top grey bar!  Ideally, I would love to push our work from this sprint into production right afterward (IE the week of 2/27) OR early in the subsequent sprint (first week of March), which means we should limit the CSS changes we work on this sprint to MVP.  

Let's keep a running list of CSS changes for 'first round' and later on - I can start a new issue but let me know if there's a better way to do this to respect your workflows!

@okaycj is this timeline doable?

mekline (commented):
In an ideal world I'd love to include the grey top bar; I also probably want to think about the icon colors for clipboard/coffee/balloon, but do *not* mind leaving them as is for now.

becky-gilbert:
> Let's keep a running list of CSS changes for 'first round' and later on - I can start a new issue but let me know if there's a better way to do this to respect your workflows!

It's fine with me to put this running list of changes in a separate issue (so that CJ can carry on with all of the views). I'd be happy to tackle the gray bar issue and other minor/'second round' things.

PR #1101, opened 2023-02-10
Static file reorganization - fixes #852

### LINKED ISSUE #852: Collecting static directories into one directory
**TL;DR**: All static files should be under the same parent directory. 

**Narrative**
When generating the integrity hash for some JS files, I found static files in many different locations.   It would be nicer and easier to find the files if they were under the same static directory.  

**Acceptance Criteria**
- [x] All static files are under the static directory
- [x] `collectstatic` works correctly on development

**Implementation Notes**
Moving the files and renaming their location within the templates seems straight forward.  I am unsure how GCP will react to change in static file location.

becky-gilbert:
Reorganization of static files: see #852 for details. 
This is a work-in-progress - I will update as I work on the remaining to-do items.


okaycj (approved):
LGTM!

becky-gilbert:
@okaycj I just realized that if I merge this into `develop` now, then it might break the `develop` branch after we merge `bootstrap-5` (because this includes a change to the static file directory settings). So I'd like to merge `bootstrap-5` into `develop` first, then I'll update this branch with those changes and test/things as needed.

PR #1108, opened 2023-02-14
Summary char limit - fixes #1075

### LINKED ISSUE #1075: Add/change length validation for preview summary text
# TL;DR
Researchers are beginning to write summary text that is too long to fit nicely on the search page!  

# Narrative

As a family, I want to see a short, clear description of what I might learn about, not a wall of text that breaks up the search page.
As a reviewer (who will soon not be doing as much reviewer), I don't want to repeatedly tell researchers to make this text shorter.

# Acceptance Criteria

Accept when the new study/edit study form returns an error message telling the user to make the summary shorter if they submit a too-long text string!

# Implementation Notes

 The help text already tells researchers to keep this text under 100 words, but it would be okay if the validation gave them a bit of leeway, e.g. 150 words - the goal is brevity, not a tight limit.

becky-gilbert:
As discussed in #1075, this PR is to reduce the text input limit for study preview summaries from 500 to 300 characters.

Here is what the new limit looks like (test study):

Screenshot 2023-02-14 at 1 56 07 PM

The text input field will now stop accepting characters after 300. The limit is also indicated via the preview summary help text:

Screenshot 2023-02-14 at 2 03 39 PM

This will not change anything for existing studies that already have a preview summary between 300 and 500 characters. However, if the researcher edits the study and tries to save it, they will get a form validation error and banner message:

Screenshot 2023-02-14 at 2 21 19 PM
Screenshot 2023-02-14 at 2 21 41 PM

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -339,7 +339,7 @@ class StudyForm(ModelForm):
             "lab": "Which lab this study will be affiliated with",
             "image": "This is the image participants will see when browsing studies. Please make sure that your image file dimensions are square and the size is less than 1 MB.",
             "exit_url": "Specify the page where you want to send your participants after they've completed the study. (The 'Past studies' page on Lookit is a good default option.)",
-            "preview_summary": "This is the text participants will see when browsing studies. Please keep your description under 100 words.",
+            "preview_summary": "This is the text participants will see when browsing studies. The limit is 300 characters.",
             "short_description": "Describe what happens during your study here. This should give families a concrete idea of what they will be doing - e.g., reading a story together and answering questions, watching a short video, playing a game about numbers. If you are running a scheduled study, make sure to include a description of how they will sign up and access the study session.",
             "purpose": "Explain the purpose of your study here. This should address what question this study answers AND why that is an interesting or important question, in layperson-friendly terms.",
             "contact_info": "This should give the name of the PI for your study, and an email address where the PI or study staff can be reached with questions. Format: PIs Name (contact: youremail@lab.edu)",
--- a/studies/models.py
+++ b/studies/models.py
@@ -296,7 +296,7 @@ class Study(models.Model):
     priority = models.IntegerField(
         validators=[MinValueValidator(1), MaxValueValidator(99)], default=99
     )
-    preview_summary = models.CharField(max_length=500, default="")
+    preview_summary = models.CharField(max_length=300, default="")
     short_description = models.TextField()
     purpose = models.TextField()
     criteria = models.TextField()


okaycj (changes requested):
Nice work! If the model is updated, there will usually be a need for a migration.  Would you mind running `makemigration` and include it in this PR?

mekline (approved):

becky-gilbert:
> If the model is updated, there will usually be a need for a migration. Would you mind running makemigration and include it in this PR?

@okaycj Done! But when I ran `make migrate` after producing this migration file, I get the error: "django.db.utils.DataError: value too long for type character varying(300)". Looks like it's not happy with existing preview summaries in the database that are longer than the character limit. Did I need to do anything to fix this?

okaycj:
> @okaycj Done! But when I ran `make migrate` after producing this migration file, I get the error: "django.db.utils.DataError: value too long for type character varying(300)". Looks like it's not happy with existing preview summaries in the database that are longer than the character limit. Did I need to do anything to fix this?

If I do the following I have no issues:
1. Remove DB
2. Add DB from staging
3. Run migrations

Here's the output:
```
Projects/lookit-api - [summary-char-limit] » make migrate
docker compose run --rm web poetry run ./manage.py migrate
[+] Running 1/0
 ⠿ Container lookit-api-db  Running                                                                                                                                                                    0.0s
-------------------Why yes, we are using local assets!-------------------------
Operations to perform:
  Apply all migrations: accounts, admin, auth, authtoken, contenttypes, django_celery_beat, guardian, sessions, sites, studies, web
Running migrations:
  Applying studies.0089_auto_20230217_1855... OK
Projects/lookit-api - [summary-char-limit] »
```

becky-gilbert:
I've updated this PR to use the approach discussed in #1075. 
The new migration file finds any studies with preview summaries greater than 300 characters, and for each of these:
1. reduces the preview summary length to 297
2. adds "..." to the end of the preview summary to indicate that it has been clipped
3. copies the entire original preview summary text into the start of the study's short description section. 

Here's what this change looks like:

# Before migration

## Example preview summary - 445 characters
pre_migration_preview

## Example description
pre_migration_description

# After migration 

## Preview summary - 300 characters
post_migration_preview

## Description - original summary appended to the start
post_migration_description

The only studies that are affected are those with preview summaries greater than 300 characters.

@mekline just checking that the last 2 screenshots are what you had in mind, and that are you're still happy with 300 characters as the limit?

# Implementation notes:

@okaycj I'm using print statements to get the names and IDs of studies that are affected by the data alteration, so that we can notify researchers and they can check/edit their summaries/descriptions after this change. Let me know if there's a better way to identify the studies that are affected.

I added a reverse function so that we can undo these changes if necessary, but it's not perfect. After identifying study summaries that have been clipped, it finds the original (300+ character) preview summary by grabbing the text from the start of the study description up to the first newline character after the 300th character. The idea behind this method is to locate the end of the original summary text by finding the newline character that was added after the summary when it was copied into the start of the description. But this means that, if the original preview summary happened to contain a newline between the 300th and 500th characters, then the reverse function will split the description text in the wrong place. I think this scenario isn't super likely to occur, plus we'll have a list of the studies that are affected by this reversal, so we can notify researchers and have them check their summary and description text.
However, maybe we'd rather not have a reversal function than one that doesn't work in all scenarios - let me know if that's the case and I can remove it.

mekline (approved):
Correctly interpreted @becky, thank you!  Let's remember to announce this change to the userbase when it goes into production

okaycj:
> @okaycj I'm using print statements to get the names and IDs of studies that are affected by the data alteration, so that we can notify researchers and they can check/edit their summaries/descriptions after this change. Let me know if there's a better way to identify the studies that are affected.

That is very much a reasonable way to do this, except how are you going to get those values from production?

becky-gilbert:
> That is very much a reasonable way to do this, except how are you going to get those values from production?

@okaycj I was hoping the print statements would show up in the GCP cloud console under Kubernetes -> Workloads -> production-migration -> Logs. In that area I can see some logs about the last migration on production (e.g. "Applying studies.0089_alter_lab_slug... OK"). But I realize the `print` out location might depend on the environment settings, and I might just be seeing it in the same place as the other migration logs when things are running in local/debug mode. Do you know for sure? If not, maybe we can try it on staging?

PR #1119, opened 2023-02-17
Bootstrap 5 studies history bug fix

becky-gilbert:
This fixes a bug on `bootstrap-5` in the `studies/history` view where the message "You have not yet participated in any external studies." message is shown when the user views their history of _internal_ studies and the list is empty. 

Old:
Screenshot 2023-02-17 at 11 27 06 AM

I think this was happening because, when the study list is empty, this line in the template file always returns false (presumably because `study` or `study_type` doesn't exist?).
```
{% if study.study_type.is_ember_frame_player %}
```

This PR fixes the problem by replacing the conditional for study type with a filter method for the determining the selected tab (currently just Lookit vs external, though it will continue to work if we add more tabs), so that the form can use the tab selection to set the appropriate message text.

Fix:
Screenshot 2023-02-17 at 2 18 23 PM

### DIFF
--- a/web/templates/web/studies-history.html
+++ b/web/templates/web/studies-history.html
@@ -3,6 +3,7 @@
 {% load bootstrap_icons %}
 {% load i18n %}
 {% load static %}
+{% load web_extras %}
 {% block title %}
     {% trans "Past Studies" %}
 {% endblock title %}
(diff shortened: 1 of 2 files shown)


becky-gilbert:
@okaycj ok for me to merge this one?

PR #1121, opened 2023-02-27
Bootstrap 5 exp study detail

okaycj:
This is the experimenter's study detail view update for BS5.  This is still a WIP.


mekline:
This looks good! @okaycj, could you please write a new issue to add back in the colors (success/warn/error) that currently show under the study status bar? No need to change for now.

mekline:
Also, what's up with the "Recruitment stats" item on the nav bar? I'd prefer we keep this view as (not) prominent as it currently is, because I'm not sure about its future...

mekline (approved):

becky-gilbert:
> could you please write a new issue to add back in the colors (success/warn/error) that currently show under the study status bar? No need to change for now.

I went ahead and made this change for study and build status. I also fixed a bug that was causing the study status state/color (success/warning/danger) to be used for all items in that section.

> what's up with the "Recruitment stats" item on the nav bar? I'd prefer we keep this view as (not) prominent as it currently is, because I'm not sure about its future...

@mekline not sure what you mean? "Recruitment Stats" looks the same as the other nav bar items to me, and it's also shown on the study detail pages on production.

becky-gilbert (approved):

mekline:
Doh! I forgot that this was an experimenter view and was imagining that families would see this for some reason. Thanks for correcting me :)

PR #1132, opened 2023-03-10
Bootstrap 5 response views

becky-gilbert:
Changes to the experimenter responses views for the bootstrap 5 update. 
This PR includes the URLs that correspond to the 4 tabs on the study responses page (`/exp/studies/<int>/responses`):
- exp/studies/int:pk/responses/
-  exp/studies/int:pk/responses/all/
-  exp/studies/int:pk/responses/demographics/
-  exp/studies/int:pk/responses/attachments/

As well as the URLs used for downloading files, adding/updating feedback, deleting preview data, and :
-  exp/studies/int:pk/responses/download/
-  exp/studies/int:pk/responses/feedback/
-  exp/studies/int:pk/responses/all/download_json/
-  exp/studies/int:pk/responses/all/download_summary_csv/
-  exp/studies/int:pk/responses/all/download_summary_dict_csv/
-  exp/studies/int:pk/responses/all/download_summary_children_csv/
-  exp/studies/int:pk/responses/all/download_summary_children_dict_csv/
-  exp/studies/int:pk/responses/all/collision_check/
-  exp/studies/int:pk/responses/all/download_frame_zip_csv/
-  exp/studies/int:pk/responses/all/download_frame_dict_csv/
-  exp/studies/int:pk/responses/demographics/download_json/
-  exp/studies/int:pk/responses/demographics/download_csv/
-  exp/studies/int:pk/responses/demographics/download_csv_dict/
-  exp/studies/int:pk/responses/videos/int:video/
-  exp/studies/int:pk/responses/delete_preview/

Individual responses:

With existing feedback that can be edited:

All responses:

Demographics:

Videos:


studies/templates/studies/_all_json_and_csv_data.html
```diff
@@ -104,55 +86,46 @@ <h1>All Responses</h1>
...
-                                            type="submit"
-                                            formaction="{% url 'exp:study-responses-download-json' pk=study.id %}"
-                                            {% if not n_responses %} disabled{% endif %}>
-                                        <span> <i class="fa fa-download"></i> (JSON) </span>
-                                    </button>
+                                    {% if n_responses %}
+                                        {% bootstrap_button bs_icon_download|add:" JSON" button_type="submit" button_class="btn btn-primary" id="download-all-data-json" formaction=url_download_all_json %}
+                                    {% else %}
+                                        {% bootstrap_button bs_icon_download|add:" JSON" button_type="submit" button_class="btn btn-primary" id="download-all-data-json" formaction=url_download_all_json disabled="disabled" %}
+                                    {% endif %}
```
becky-gilbert, line 96: @okaycj is there a better way to do this? The only thing that differs between the buttons in the if/else blocks is the disabled tag, which used to be set via a conditional (`{% if not n_responses %}...`). But I wasn't sure how to use a conditional here since I'm already inside a template tag. Maybe I could create a variable?

studies/templates/studies/study_attachments.html
```diff
@@ -2,71 +2,55 @@
...
 {% endblock title %}
 {% block head %}
     {{ block.super }}
-    <script type="application/javascript">
-        $(document).ready(function () {
-            if ("{{ match }}" !== '') {
-                document.getElementById("search-attachments").value = "{{ match }}";
-            }
-        });
-    </script>
```
becky-gilbert, line 16: @okaycj I know you mentioned removing this in other places. I think this can be safely removed here too, but not 100% sure.

becky-gilbert:
@okaycj I'm having trouble generating the video data needed to test the video-related aspects of these views, namely the "Videos" section for selected responses on the "Individual responses" tab, and the video download buttons on the "Videos" tab. If you have a study with video data in your local environment, would you mind testing those two things for me?

studies/templates/studies/_all_json_and_csv_data.html
okaycj: We've already spoken about this, but I'll link [the solution here](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#yesno).

studies/templates/studies/_all_json_and_csv_data.html
```diff
@@ -104,55 +86,46 @@ <h1>All Responses</h1>
...
-                                <p class="help-text">
+                                <p class="fst-italic text-muted mt-2">
                                     The response overview file gives high-level information about each response - the account and child IDs, consent approval information, condition assignment, and information about the child such as gender and languages spoken. There is one row per response. This can be used in conjunction with frame data (below) to avoid having to parse JSON in your analysis.
                                 </p>
                             </div>
-                            <div class="col-md-5 col-sm-6">
-                                <div class="pull-right download-button">
+                            <div class="col">
+                                <div class="text-end my-3 download-button">
                                     Data
```
okaycj, line 110: Wow!  I totally missed "text-end".  Nice find.

mekline (approved):

okaycj (approved):
LGTM!  I know you have a couple of changes you still want to make.  Feel free to squash/merge after you're satisfied.

studies/templates/studies/_all_json_and_csv_data.html
becky-gilbert: @okaycj I tried this and couldn't get it to work 😞 First I tried this:
```
{% bootstrap_button ... disabled=n_responses|yesno:",disabled" %}
```
but that doesn't work because the presence of the disabled attribute itself is what makes the button disabled, so the button is always disabled and there's no string value (empty, false etc.) that will enable it.

I also tried this:
```
{% bootstrap_button ... n_responses|yesno:",disabled='disabled'" %}
```
but that gives me this error:
> 'bootstrap_button' received some positional argument(s) after some keyword argument(s)

I'm stumped and need to move on, so I'm going to leave this as is and merge it. But if you can suggest a way to improve this I will definitely come back and fix it!

PR #1137, opened 2023-03-11
Bootstrap 5 2FA setup

becky-gilbert:
BS5 changes for the 2fa set up page.

Screenshot 2023-03-10 at 11 43 40 PM


okaycj (approved):
💯!  I like the css file per view.  I think I'll continue this pattern forward.

PR #1142, opened 2023-03-15
remove duplicate files from `web/base.html`

becky-gilbert:
It looks like we accidentally introduced some duplication of JS and CSS file loading lines (same URL and integrity hash) in the `head` of `web/templates/web/base.html` on `develop`. This PR removes the duplicates.


okaycj (approved):
💯 LGTM!

PR #1143, opened 2023-03-20
Python sass

okaycj:
This add a script to django management to generate the custom css from scss files.  This script will check the django-bootstrap5 package for BS5 version.  Additionally, it'll alert the user if the custom css file has changed.  

TODO:
1. figure out how to handle secondary elements
2. Implement the rest of the requested style changes

Updated home page.  
Screenshot 2023-03-20 at 12 48 39 PM
Screenshot 2023-03-20 at 12 48 42 PM
FAQ

Screenshot 2023-03-20 at 12 48 49 PM


mekline (approved):
(Approved pending the CI/CD checks passing as usual)

Makefile
```diff
@@ -57,3 +57,6 @@ collectstatic:
 
 lint: 
 	poetry install --sync && poetry run pre-commit run --all-files
+
+css:
+	poetry run ./manage.py custom_bootstrap5
```
becky-gilbert, line 62: Do you think this should always run at the start of the `collectstatic` make command (above), since it generates a file that gets stored in `web/static`? If so, can I add `make css` or the corresponding line to the start of `collectstatic`?

becky-gilbert (approved):

Makefile
okaycj: We answered this question in another conversation.  We've decided to commit the custom css file to our repo, so there is no need to worry about the timing of these commands.

PR #1149, opened 2023-03-22
BS5 button classes

becky-gilbert:
This PR makes the button classes/styling consistent across views. It moves the primary and secondary button classes into template tags in `web_extras` and replaces all primary/secondary button classes with values from these template tags. We can easily change the specific styling for the primary/secondary buttons. The point of this PR is to make them all consistent.
- Primary (positive action) buttons: **CHS blue background and black text**
- Secondary (cancel and 'less important') buttons: **light gray background and dark gray text/border**
- **There are no more green buttons**.
- Navigation bar on the experimenter manage studies page now matches the main navigation bar (i.e. gray with a filled button for the active nav item, rather than link styling).

**Green/Success**

Please let me know if we want to add any green buttons back in (either for all positive action buttons or for a subset of them) and I can do that.
We do still use the green/success color scheme to signal success in other places, like study status ('experiment runner built'), banner messages ('saved successfully'), and video consent approval. 

**Red/Danger**

There are still a few 'danger' (red) button types for things like deleting account info, removing researcher access from a study, and saving a study after it has been approved. 
Developer note: I didn't create a template tag for this category since it's only used in a few places and it just uses the standard 'btn-danger' class, but I can do so if we want.

**Examples**

- Positive action buttons ("create", "save", "download") are styled the same. 
- Cancel buttons and 'less important' buttons ("update child", "update experimenter runner") are styled the same.

Screenshot 2023-03-22 at 9 56 44 AM
Screenshot 2023-03-22 at 9 53 13 AM
Screenshot 2023-03-22 at 9 49 33 AM

- Manage researchers for a study - the 'add' button is now blue. 
- We're still using red (danger) for delete/remove.

Screenshot 2023-03-22 at 10 01 18 AM

- Fixed the issue where "add child" button style was inconsistent in account/children vs account/add-child/, and "update child" button was inconsistent (dark gray vs light gray buttons elsewhere). We're still using red (danger) for 'delete'.

Screenshot 2023-03-22 at 9 49 54 AM
Screenshot 2023-03-22 at 10 53 50 AM

Screenshot 2023-03-22 at 9 50 35 AM

- Study status buttons have been 'demoted' from primary to secondary

Screenshot 2023-03-22 at 9 47 44 AM

- Navigation items on the experimenter manage study page now match the main navigation bar.

Screenshot 2023-03-22 at 4 22 24 PM

### DIFF
--- a/accounts/templates/accounts/2fa-setup.html
+++ b/accounts/templates/accounts/2fa-setup.html
@@ -2,10 +2,12 @@
 {% load django_bootstrap5 %}
 {% load bootstrap_icons %}
 {% load static %}
+{% load web_extras %}
 {% block title %}
     Set up Two Factor Authentication
 {% endblock title %}
 {% block content %}
+    {% button_primary_classes as btn_primary_classes %}
     {% bootstrap_messages %}
     <div class="row">
         <div class="col">
@@ -28,7 +30,7 @@
                                     {% csrf_token %}
                                     {% bootstrap_form_errors form %}
                                     {% bootstrap_form form %}
-                                    {% bootstrap_button "Test Code" button_type="submit" button_class="btn-primary" %}
+                                    {% bootstrap_button "Test Code" button_type="submit" button_class=btn_primary_classes %}
                                 </form>
                             </li>
                         </ol>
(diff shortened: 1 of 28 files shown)


mekline (approved):
Great!

okaycj (approved):
💯

PR #1150, opened 2023-03-23
BS5 content and navbar spacing

becky-gilbert:
This PR is to standardize (1) the spacing between the navigation bar, page heading(s) and/or the rest of the page content, and (2) the indentation of content relative to the navigation bar. Example screenshots below.

I've created template tags for BS5 classes for the page's container div, main title/heading (h1) and subheadings. However so far I've only implemented these in a small subset of pages in the web app. I will add more pages to this PR ASAP as I get them done.

I also made a small change to the state drop-down selection on the Resources page: I added "Select a state" to the default/blank option.

Screenshot 2023-03-22 at 11 17 33 PM
Screenshot 2023-03-22 at 11 17 49 PM
Screenshot 2023-03-22 at 11 18 01 PM
Screenshot 2023-03-22 at 11 18 11 PM

PR #1152, opened 2023-03-23
Fix registration views

okaycj:
This PR is to fix the (Overlooked) registration views.  

Screenshot 2023-03-23 at 3 54 13 PM


becky-gilbert (approved):
I went through the forgot/reset password flow. Works and looks great!

Thanks also for fixing the duplicate form error messages.

The one minor thing (that you already know about) is the lack of spacing between the nav bar and cards:
Screenshot 2023-03-23 at 4 43 23 PM

This should be fixed when #1150 is merged 👍

PR #1154, opened 2023-03-27
Limit python version to only 3.9

okaycj:
Sometimes when we're installing local environments we forget to set the version of python.  This change will throw an error if a version other then '3.9' is used.


mekline (approved):
Nice process improvement!

PR #1166, opened 2023-04-04
Found error where variable wasn't set

### LINKED ISSUE #1163: Superuser can't access lab pages
**Description**

My superuser account has lost the ability to access lab pages; this functionality is needed so I can go in and approve new labs for testing as they finish their various paperwork/requirements. 

This is a new issue as of the most recent version, because I access links like these at least weekly. 

**How to reproduce**

One my account (will be curious to know what happens on other SU accounts), I can access:

https://lookit.mit.edu/exp/labs/2/ <- the demo lab
https://lookit.mit.edu/exp/labs/3/ <- the actual ECCL research lab 

(I am an admin-level member of both of these labs.)

I cannot access: 

https://lookit.mit.edu/exp/labs/4/
https://lookit.mit.edu/exp/labs/150/

I get a Server Error (500) on these. 

I can't see it right now, but I believe I am *not* listed as a "lab member" on these/all other labs (which is the appropriate behavior), but am nevertheless able to see & modify due to my superpowers.

**Expected behavior**

A superuser account can access & modify information for all labs on the site, even when not a member of that lab. 

**Additional context**

Just that this is definitely new/was not a bug as of ~2 weeks ago.

okaycj:
On local (Staging) data, there was a template error were a variable wasn't set.  After making sure value was set, errors seem to be fixed. 

Closes #1163

### DIFF
--- a/studies/templates/studies/lab_detail.html
+++ b/studies/templates/studies/lab_detail.html
@@ -19,6 +19,8 @@
         {% set_variable "<div>You have joined this lab.</div>" as lab_join_message %}
     {% elif requested_this_lab %}
         {% set_variable "<div>You have requested to join this lab.</div>" as lab_join_message %}
+    {% else %}
+        {% set_variable "" as lab_join_message %}
     {% endif %}
     {% set_variable link_edit_lab|add:link_lab_members|add:lab_join_message as page_title_right_content %}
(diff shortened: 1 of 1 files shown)


becky-gilbert (approved):
Thanks for catching and fixing this!

PR #1169, opened 2023-04-06
Empty text

okaycj:
This PR adds template function to properly format list "empty text".

Screenshot 2023-04-06 at 11 39 29 AM
Screenshot 2023-04-06 at 11 39 35 AM
Screenshot 2023-04-06 at 11 41 01 AM


accounts/templates/accounts/participant_detail.html
```diff
@@ -82,15 +83,15 @@
                         {% for birthday in user.latest_demographics.child_birthdays %}
                             {{ birthday | timesince|default:"N/A" }}
                             {% if not forloop.last %};{% endif %}
-                        {% empty %}{% trans "No Response" %}
+                        {% empty %}
+                            {% trans "No Response" as trans_no_response %}
+                        {% empty_text trans_no_response %}
```
becky-gilbert, line 88: I think we should go back to the non-template-tag version in this case, i.e. `<td>No Response</td>` to match the other NA values in the rest of the table.

becky-gilbert (changes requested):
This looks great! 
I did find one minor issue. In `accounts/participant_detail.html`, there's an `empty_text` template tag for one of the empty values (Children current ages), but this doesn't match the other NA values on this page, which are all added as a default "No Response" in `<td>` tags. So I think we should not use the `empty_text` template tag for "Child current ages" in order to match the rest of the table.
Here's an example showing a `empty_text`-templated 'No Response` value for "Children current ages", along with some other non-templated "No Response" values.
Screenshot 2023-04-07 at 4 25 43 PM
If it's ok with you, I'll make this change now and then approve the PR.

becky-gilbert:
I made my requested change here: [48df787](https://github.com/lookit/lookit-api/pull/1169/commits/48df787d1f684ae392d19d29644ef917ec39a0aa)

Here's what the participant detail page looks like now:
Screenshot 2023-04-07 at 4 41 27 PM

I think it's ready to merge!

becky-gilbert (approved):

mekline (approved):

PR #1170, opened 2023-04-06
Change the icons on the home page to red

okaycj:
Add appropriate sass to add red theme to our css.  Add new "text-red" class to icons on home page. 

Screenshot 2023-04-06 at 12 44 33 PM

### DIFF
--- a/scss/base.scss
+++ b/scss/base.scss
@@ -1,9 +1,15 @@
-@import "custom";
 @import "2fa-setup";
 @import "all-json-and-csv-data";
 @import "navigation";
 @import "study-fields";
 @import "study-responses";
 @import "superuser";
-// This import should always be last.
+
+// Add imports to adjust default colors and theme maps
+@import "bootstrap-5.2.0/scss/functions";
+@import "custom-variables";
+@import "bootstrap-5.2.0/scss/variables";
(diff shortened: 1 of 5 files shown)


becky-gilbert:
@okaycj jcould you explain why we need all of the changes in `web/static/custom_bootstrap5.css`, instead of just changing the default `red` value to match the CHS red (i.e. `$red: #ff5f5c;` in `scss/custom-variables.scss`). Is it just so that we could use a new custom class (`text-red`) rather than `text-danger` class to modify the icon color, or is there more to it than that? (It's fine if this is just to create the new `text-red` class, I just want to make sure I understand!)

becky-gilbert:
> @okaycj could you explain why we need all of the changes in web/static/custom_bootstrap5.css, instead of just changing the default red value to match the CHS red (i.e. $red: #ff5f5c; in scss/custom-variables.scss). Is it just so that we could use a new custom class (text-red) rather than text-danger class to modify the icon color, or is there more to it than that? (It's fine if this is just to create the new text-red class, I just want to make sure I understand!)

Nevermind, I get it now per our conversation 👍

becky-gilbert (approved):

PR #1175, opened 2023-04-12
Remove panel class from response consent view

okaycj:
To complete a task on the BS5 style concerns issue, I've removed the last "panel-*" class from our project.  There are two html files that still have this class, but they're not being used.  At some point we'll have to remove unused html files. 

Below is the images from the one view altered for this PR:
Screenshot 2023-04-12 at 11 00 36 AM
Screenshot 2023-04-12 at 11 00 39 AM

### DIFF
--- a/studies/templates/studies/study_responses_consent_ruling.html
+++ b/studies/templates/studies/study_responses_consent_ruling.html
@@ -111,11 +111,9 @@
                         </li>
                     {% endfor %}
                 </ul>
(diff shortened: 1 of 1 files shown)


mekline (approved):
Adios panels!! 

(I am not able to perceive a difference between this & this current page; just checking that these were the intended screenshots. But also if I just can't tell the difference, then great job :) )

PR #1177, opened 2023-04-17
Replace template function with new css class

okaycj:
Replace template function for sass class that extends BS5 classes that were used to style empty text.

### DIFF
--- a/accounts/templates/accounts/participant_detail.html
+++ b/accounts/templates/accounts/participant_detail.html
@@ -50,8 +50,7 @@
                     </div>
                 </div>
             {% empty %}
-                {% trans "No children profiles registered!" as trans_no_children %}
-            {% empty_text trans_no_children %}
+                <div class="empty-text">{% trans "No children profiles registered!" %}</div>
             {% endfor %}
         </div>
         <div class="col-6">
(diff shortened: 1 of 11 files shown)


becky-gilbert (approved):
Thanks, this does seem like a better solution! 👍 

For my own understanding: the only thing you had to add to create this class was the class name and `@extend` line to `base.css`, and the change to `custom_bootstrap5.css` were auto-generated - right?

Also, does this mean our general approach should be to custom classes in this way whenever possible, and then move to custom template tags when necessary?

okaycj:
> For my own understanding: the only thing you had to add to create this class was the class name and `@extend` line to `base.css`, and the change to `custom_bootstrap5.css` were auto-generated - right?

You're correct.  Here's the sass documentation I pulled this from -> https://sass-lang.com/documentation/at-rules/extend.

okaycj:
> Also, does this mean our general approach should be to custom classes in this way whenever possible, and then move to custom template tags when necessary?

I would say this is a column A/B situation.  But yes this is far easier to understand and implement.  The template function becomes clunky fast.  We can see this "clunky"ness with the breadcrumbs function.

PR #1186, opened 2023-04-24
Remove invoke

okaycj:
Dependabot wants to update invoke, but we're no longer using invoke to setup local development environment.  Removing this package makes more sense.


becky-gilbert (approved):
A few questions/comments: 
- I noticed that `tasks.py` is deleted. I assume that's because everything in that file is related to invoke? (It appears that way but I just wanted to be sure).
- There's quite a lot removed from `contributing.md`. I know that some of the deleted stuff is related to invoke, but there's some other stuff that was removed that I think is still useful/relevant. But no problem if it's easier to just delete large sections now as we'll still have the old version in version history - maybe I can bring some of it back later.

okaycj:
> * There's quite a lot removed from `contributing.md`. I know that some of the deleted stuff is related to invoke, but there's some other stuff that was removed that I think is still useful/relevant. But no problem if it's easier to just delete large sections now as we'll still have the old version in version history - maybe I can bring some of it back later.

Some was invoke... other stuff was release and version stuff that we no longer do.  I've made this argument before, but most of that document is probably unneeded, inaccurate, or from when the project was new.

mekline:
I thought we'd merged those contribute changes already, so go ahead if it's inaccurate!

mekline (approved):

PR #1202, opened 2023-05-02
Add age filters

okaycj:
Here are the new age grouped filters for the study list view. 

Screenshot 2023-05-02 at 10 42 26 AM
Screenshot 2023-05-02 at 10 42 32 AM
Screenshot 2023-05-02 at 10 42 41 AM
Screenshot 2023-05-02 at 10 42 53 AM
Screenshot 2023-05-02 at 10 43 03 AM
Screenshot 2023-05-02 at 10 43 16 AM

### DIFF
--- a/web/urls.py
+++ b/web/urls.py
@@ -6,6 +6,11 @@ from web import views
 app_name = "web"
 
 urlpatterns = [
+    path("studies/babies/", views.StudiesListViewBabies.as_view()),
+    path("studies/toddlers/", views.StudiesListViewToddlers.as_view()),
+    path("studies/preschoolers/", views.StudiesListViewPreschoolers.as_view()),
+    path("studies/school-age-kids/", views.StudiesListViewSchoolAgeKids.as_view()),
+    path("studies/adults/", views.StudiesListViewAdults.as_view()),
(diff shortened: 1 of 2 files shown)


becky-gilbert:
@mekline @okaycj I know this is already merged, but I just noticed something about the age-group-filtered study pages. When the user IS NOT logged in then the age group filtering is indicated by the age group shown in the "Find studies for..." drop-down menu, as in the screenshots above. However, if the user IS logged in then there's no indication of the age-group filtering once they're on the page, other than the URL (which is easy to miss!). When the user is logged in, the drop-down options for "Find studies for..." change from the age groups to specific children. Here's what the 'babies' page looks like when logged in:

Screenshot 2023-05-03 at 10 08 35 AM

Do we want to add some kind of indication of the age-group filter on these pages when the user is logged in? Maybe we could add a page title somewhere, or put that specific age-group value back into the drop-down menu..?

mekline:
Good catch!  Can we have it so the drop-down menu starts on the
corresponding filter?  Would changing the filter actually resolve to
showing all studies on site, or do we actually need to redirect people back
to the main search page for that behavior?

PR #1217, opened 2023-05-09
Remove reference to CHS on Resources for merger

becky-gilbert:
As part of our upcoming merger with CHS, this PR removes the reference to the separate CHS site in the "Other ways to participate from home" section on the Resources page. This just leaves SciStarter as the only link in that section.

Screenshot 2023-05-09 at 3 37 15 PM


okaycj (approved):
💯

mekline (approved):

PR #1229, opened 2023-05-30
Fix issue with JS that is blocking state change.

okaycj:
PR to fix issue with state change for external studies. 

out


mekline (approved):
A test is failing but this update looks sensible!

- Melissa

PR #1230, opened 2023-05-30
Move bug fix into production

okaycj:
This PR is to get a bug fix into production.


becky-gilbert (approved):
LGTM! Sorry for missing this problem with the `commentsHelpText` in the initial review. 

And thanks for also catching/fixing the issue with `declarations[trigger]` being undefined for the collecting data checkbox when the trigger is not 'submit' 👍

PR #1233, opened 2023-06-02
Add S3 env vars into `run_docker_container` (new var names)

becky-gilbert:
This PR is another attempt at adding the S3 environment variables (for saving the EFP video recordings) to the `run_docker_container` method in the `EmberFrameplayerBuilder` class. This is a test to see if these variables will be become available to the lookit-api and EFP environments.

This is the same as my previous attempt here: #1189 except that the variable names have changed to match those used in my latest update in the orchestrator repo (see here: https://github.com/lookit/lookit-orchestrator/pull/41).


okaycj (approved):
LGTM!

PR #1239, opened 2023-07-06
Efp local build

okaycj:
This PR updates the EFP build so that it works in the local environment.


project/settings.py
```diff
@@ -291,8 +291,7 @@
     STATICFILES_LOCATION = "/static"
 
     MEDIAFILES_LOCATION = "/media"
-
-    EXPERIMENT_LOCATION = "/experiments"
+    EXPERIMENT_LOCATION = "experiments"
```
becky-gilbert, line 294: This could be moved out of the if/else statement since it's now the same value for cloud and local storage. Not a big deal though.

becky-gilbert (approved):
LGTM! I tried it locally and it works great 👍 
It doesn't look like you've done anything that would affect how things work in the remote environments, but we should still remember to test this after merging by building and running an experiment on staging (or is this covered by automated tests?).

PR #1240, opened 2023-07-18
fix test: sometimes fails due to list order mismatch

becky-gilbert:
This PR fixes the `test_correct_message_structure` test in `studies/tests.py` that can fail due to a mismatch in the order of elements between two lists. The order mismatch occurs because the `all` method on querysets does not necessarily return objects in the same order. This PR fixes the problem by comparing sets of objects rather than lists.

Also, the test result output contains logs and warnings that make it difficult to see clearly whether the tests are running as expected, or if there's anything that needs to be fixed. This PR does two other things to clean up the test result output. 
1. Disables logging for non-critical issues during the `CheckPipeProcessingTestCase` tests, since these tests always produce error logs (due to `Video` model [`check_and_parse_pipe_payload`](https://github.com/lookit/lookit-api/blob/f05dc63d2d1c3544e798f29f97d296e5151ce883/studies/models.py#L1301-L1316)). Note that disabling the logs here only prevents the messages from showing up in the test output - it does not prevent exceptions from being raised.
2. Add ordering to any querysets that are passed into `Paginator` to fix this warning: `UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list`.

Test results on `develop`:

Screenshot 2023-07-19 at 10 34 03 AM

Test results on `fix-test-issues`:

Screenshot 2023-07-19 at 10 29 28 AM

These last two changes are optional - I can remove them from this PR if there are any concerns.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -436,7 +436,11 @@ class FeedbackViewSet(FilterByUrlKwargsMixin, ConvertUuidToIdMixin, views.ModelV
             | (Q(study__id__in=studies_for_preview) & Q(is_preview=True))
         )
         response_ids = consented_responses.values_list("id", flat=True)
-        return qs.filter(
-            Q(response__id__in=response_ids)
-            | Q(response__child__user=self.request.user)
-        ).distinct()
+        return (
+            qs.filter(
+                Q(response__id__in=response_ids)
+                | Q(response__child__user=self.request.user)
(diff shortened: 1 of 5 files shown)


okaycj (approved):
LGTM!

mekline (approved):

Issue #1074, opened 2023-01-27
Replace pipe in lookit API


becky-gilbert:
This issue is to track and scope the remaining work around replacing Pipe with RecordRTC and AWS S3 in the lookit-api.  Please edit/add/comment with anything I got wrong or missed!

These are things we need to do on the lookit API side to get RecordRTC/S3 working:
- [x] 1. Update `studies/experiment_builder.py` with the new S3 environment variables needed for the docker container that builds the ember app. **Done on staging/`develop` [here](https://github.com/lookit/lookit-api/blob/f05dc63d2d1c3544e798f29f97d296e5151ce883/studies/experiment_builder.py#L235-L238).**
- [x] 2. Handle permissions for managing access to S3 videos. #1237 
- [x] 3. ~~Pipe webhook: At the moment, we're using Pipe's webhook that fires upon uploading to S3. The handler that renames the video file is in `/exp/views/video.py`. We may need to trigger this in another way (since I believe it is dependent on Pipe?) and change the way it is authenticated (since it currently uses a Pipe webhook key).~~ (No need to trigger the same view that's used for Pipe - instead we can just add to the existing API)
- [x] 4. Add API endpoint that takes new video data POSTed from S3 and creates new video object in database. This will use token authentication rather than user credentials. #1237 
- [x] 5. Modify the views that provide video recordings so that the recordings are fetched from the correct bucket, depending on whether the study used pipe or recordrtc. These views include the participant study history page, researcher consent rulings, and researcher response video download (any other views?). #1245 (this also required adding `recording_method` to data sent from EFP, added here: https://github.com/lookit/ember-lookit-frameplayer/commit/8285603260e649afd22579a44a5539412582672f)
- [x] 6. Write new video API tests in `api/tests/`. #1237 
- [x] 7. Write the AWS Lambda function that will be used for (a) saving a video object to the database after it has been saved to S3 and (b) logging the S3-related activity (using AWS CloudWatch). 

Some of this work has generated new issues:
- We also need a way of replicating this system (video upload to S3 -> Lambda -> lookit-api) in our local development environment, and deploying local changes to the cloud Lambda function. #1290 
- Ideally we would also provide the videos in mp4 format for researchers. #1255 

These are things we'll need to do once Pipe is actually deprecated:
- Possible database migration and related code query changes if we change column names such as `pipe_name`, `pipe_numeric_id`.
- Update `contributing.md` and other docs related to setting up local dev environment. (Added to #1290)

[...]

PR #1245, opened 2023-07-27
Replace pipe video access

becky-gilbert:
Related issue: #1074 (but merging this PR shouldn't automatically close the issue)

This PR contains some of the changes required to support our pipe-replacement webcam recordings. To go along with the other changes made to support non-Pipe webcam recordings in #1237, we need to fetch the recordings from the correct S3 bucket, depending on whether the video was recorded using Pipe or RecordRTC. 

The changes in this PR depend on those made in the `replace-pipe` branch (#1237) - I've just put them in a separate branch off of `replace-pipe` to make the PRs a little smaller and easier to review.

This PR does the following:            
- Adds an argument to the `get_download_url` attachment helper function indicating whether or not the video was recorded with Pipe, and generates the correct download URL accordingly.
- Adds this argument to the `get_download_url` calls in `studies/queries.py` and `studies/models.py`.
- Adds logic to the `delete_video_from_cloud` Celery task to determine the recording method and corresponding bucket before attempting to delete the video from S3.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -442,10 +442,15 @@ class FeedbackViewSet(FilterByUrlKwargsMixin, ConvertUuidToIdMixin, views.ModelV
             | (Q(study__id__in=studies_for_preview) & Q(is_preview=True))
         )
         response_ids = consented_responses.values_list("id", flat=True)
-        return qs.filter(
-            Q(response__id__in=response_ids)
-            | Q(response__child__user=self.request.user)
-        ).distinct()
+
+        return (
+            qs.filter(
+                Q(response__id__in=response_ids)
+                | Q(response__child__user=self.request.user)
+            )
+            .distinct()
+            .order_by("-id")
+        )
 
 
 class VideoViewSet(ConvertUuidToIdMixin, views.ModelViewSet):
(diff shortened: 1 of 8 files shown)


mekline (commented):
This logic generally makes sense to me, I'll ask you about testing strategy during standup!

becky-gilbert:
For some reason the file changes in this PR include some unrelated changes that I made and merged here: https://github.com/lookit/lookit-api/pull/1240 These changes have been made on both the incoming and base branches, but they were added to the base branch after the PR was created and the file diff isn't updating. I tried changing the base branch off of `replace-pipe` and then back again, hoping that the file diff would refresh, but no such luck. So please ignore the changes in:
- `api/views.py`
- `exp/tests/test_video_views.py`
- `exp/views/analytics.py`
- `exp/views/responses.py`
as these changes have already been merged. 

The relevant changes for review are those in:
- `attachment_helpers.py`
- `studies/models.py`
- `studies/queries.py`
- `studies/tasks.py`

okaycj (approved):
This all looks good.  Thank you.  A small question, you made a change to `exp/views/analytics.py` and I had assumed those view weren't being used.  This might be out of scope for this issue, but is it being used?

PR #1253, opened 2023-08-17
Custom query for populating participation fields

okaycj:
Change query to populate participation fields.  Because the form fields are custom, the save and initial values for the form had to also be custom.  

Before:
Screenshot 2023-08-17 at 1 26 25 PM

After:
Screenshot 2023-08-17 at 1 23 08 PM

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -235,26 +235,46 @@ class StudyUpdateView(
 
         """
         initial = super().get_initial()
+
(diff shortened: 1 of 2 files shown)


becky-gilbert (approved):
LGTM! 
Do we already have test coverage for the must (not) have participated fields populating and saving correctly? If not, that would be helpful to add.

PR #1258, opened 2023-08-24
Push to production

okaycj:
Here are the features we're pushing to production:
- Pipedeath begins!
- Developer development support.
- Performance improvements on study edit/create view.
- UI Changes to Study edit/create view.


mekline:
On the list of commits for this PR (https://github.com/lookit/lookit-api/pull/1258/commits) I see a few that have a red X instead of green check next to them.  I think of green check as referring to a review request having been approved in its final form, or maybe to some of our automatic checks failing? Are there other things that need re-reviewing from me? 

Assuming you're satisfied with the process for QA and prepping for the update, I have no objections; left comments on the copy (good catch) in the other PR).  Go forth!

becky-gilbert:
@mekline Thanks for asking about this. The red X means that one or more of the automated checks have failed for that particular commit - likely the CircleCI automated tests or the SonarCloud code analysis (which checks for things like possible security issues and code duplication). 
 
Importantly, the checks are all re-run after each push to a branch, so any green check marks that occur after a red X means that whatever check was failing in previous commits has now been fixed. So in other words, the most recent set of checks is the one that we care about.

bleonar5 (approved):

PR #1262, opened 2023-08-30
Jspsych studytype

okaycj:
More to come....


bleonar5 (approved):

okaycj:
> I'm a little stuck on getting a jspsych study to reroute back to the exitURL and record the response data. I see that I can use the on_finish function to redirect to the exitURL and I see that I have access to the responseUuid, but I'm not seeing that I have access to something like childUuid. Is there a more intuitive way to go about this? I feel like I'm probably missing something obvious

That is a feature that I didn't write into the issue and probably should have.  Nice catch.  I'll make the update.

bleonar5:
Okay, with the new commits, I now get successful redirects upon study completion, and the study response is successfully recorded and marked as completed (within the database). I'm not sure how to view the responses data within the app's UI, or whether some kind of jspsych based consent frame is necessary to have the responses be fully recognized by the app, but it may be that such concerns are outside of the scope of this PR. If so, I would approve the PR at this point.

okaycj:
@bleonar5 I accidentally blew away your change.  My apologies.  

> I'm not sure how to view the responses data within the app's UI, or whether some kind of jspsych based consent frame is necessary to have the responses be fully recognized by the app, but it may be that such concerns are outside of the scope of this PR. If so, I would approve the PR at this point.

This does seem out of scope (I'm already guilty of this for this issue).  It seems reasonable to make sure your concerns are encoded in a issue though.

bleonar5 (approved):

bleonar5 (approved):

becky-gilbert (changes requested):
This is part 1 of my review. I didn't have time to finish but wanted to share some initial comments. I'll do the rest tomorrow!

I tried out this jspsych-studytype branch locally and have a few questions/comments:

1. I'm getting 'icon does not exist' in places where we're inserting bootstrap icons (example below, but it happens on other pages too). Do you know why?
2. I tried running a Hello World study but I'm getting the error `opts is undefined` in `lookitInitJsPsych` (code and error below). Is this something that should be working on this branch, or not yet?

Icon does not exist example:

Screenshot 2023-11-29 at 4 38 31 PM

jsPsych test code:

```js
const jsPsych = initJsPsych();

const hello_trial = {
    type: jsPsychHtmlKeyboardResponse,
    stimulus: 'Hello world!'
};

jsPsych.run([hello_trial]);
```

Error:

Screenshot 2023-11-29 at 4 59 25 PM
Screenshot 2023-11-29 at 4 59 52 PM

exp/urls.py
```diff
@@ -253,10 +255,15 @@
         PreviewProxyView.as_view(),
         name="preview-proxy",
     ),
+    path(
+        "studies/preview/<int:pk>/<int:child_id>/jspsych/",
+        JsPsychPreviewView.as_view(),
+        name="preview-jspsych",
+    ),
```
becky-gilbert, line 262: I'm not sure why, but when I try to preview an external study, I'm being redirected to the jsPsych preview URL..?

Also, (less important), I noticed that the new jsPsych study preview URL uses the study and child PK IDs, and it appends jsPsych to the end of the URL. E.g. "/exp/studies/preview/751/525/jspsych/". Whereas the EFP and external preview URLs use the study/child UUIDs and don't include anything about the study type, e.g. "/exp/studies/1fa2b85d-2438-4814-aee3-21d060b32188/60d13465-68e0-4d7a-ad57-1bb0e2329565/preview/". 

Was this an intentional decision? If not, would it be better to keep the format the same? I do like the shorter URLs, but I'm wondering if there's a benefit to using UUIDs in that it makes the valid URLs less guessable and therefore less vulnerable to spam?

exp/views/study.py
```diff
@@ -873,9 +877,60 @@ def post(self, request, *args, **kwargs):
...
+        if study.study_type.is_ember_frame_player:
+            # EFP confuses "id" and "uuid"
+            kwargs["child_id"] = child.uuid
+            return HttpResponseRedirect(reverse("exp:preview-proxy", kwargs=kwargs))
+        else:
+            return redirect(
+                reverse(
+                    "exp:preview-jspsych", kwargs={"pk": study.id, "child_id": child.id}
+                )
+            )
```
becky-gilbert, line 892: Ah, this might answer my question about why external study previews are redirecting to a jsPsych URL. It looks like we need an `if study.study_type.is_external: ...`?

web/urls.py
```diff
@@ -60,6 +60,11 @@
         views.ExperimentProxyView.as_view(),
         name="experiment-proxy",
     ),
+    path(
+        "studies/jspsych/<int:pk>/<int:child_id>/",
+        views.JsPsychExperimentView.as_view(),
+        name="jspsych-experiment",
+    ),
```
becky-gilbert, line 67: Same question as with the preview URL structure. My preference would be to not include the experiment type in the study URL, which I think would mean we'd need an ExperimentRedirect view? I also have concerns about using the PKs in the URL since they are guessable and are probably still valid with typos.

web/static/js/lookit-jspsych.js
```diff
@@ -0,0 +1,121 @@
...
+    };
+}
+
+function lookitInitJsPsych(responseApiUrl, responseUuid, exitUrl) {
+    /**
+     * Function that returns a function to replace jsPsych's initJsPsych.
+     */
+    return function (opts) {
+        const jsPsych = origInitJsPsych({
+            ...opts,
```
becky-gilbert, line 108: I've tried a few test experiments and I'm still getting an error here when I try to initialize jsPsych without passing an argument (which should be possible since none of the experiment setting arguments are required, and this is how it's done in the [hello world tutorial](https://www.jspsych.org/7.3/tutorials/hello-world/#step-3-create-a-script-element-and-initialize-jspsych)).
```js
const jsPsych = initJsPsych();
```
The error goes away as soon as I pass an argument into `initJsPsych`. 
So I think the function returned by `lookitInitJsPsych` just needs to check whether `opts` exists, or maybe use an empty object as a default:
```js
return function(opts={}) {
```

studies/forms.py
```diff
@@ -539,3 +544,22 @@ class ExternalForm(ModelForm):
...
+        label="jsPsych Experiment Code",
+        widget=AceOverlayWidget(
+            mode="javascript",
+            wordwrap=True,
+            theme="textmate",
+            width="100%",
+            height="100%",
+            showprintmargin=False,
+        ),
+        help_text="Please enter jsPsych experiment code.",
```
becky-gilbert, line 560: This isn't important for now but I want to write it down so that we don't forget. We might want to make sure that users know that this box is just for the jsPsych JavaScript and not for HTML (since jsPsych users are probably used to putting their experiment in an HTML file with the actual code in script tags, and might try to do the same thing here). We'll also need to let them know what plugins are available. We could add a bit more help text here and/or add a link to our jsPsych docs.

becky-gilbert (commented):
Amazing work 💯 I left comments with questions/suggestions, but everything is looking great. I'm excited to test on staging!

exp/views/study.py
okaycj: This is preview detail view.  There hasn't been a need for a redirect to external preview.

web/static/js/lookit-jspsych.js
okaycj: This has been updated in a later commit.

studies/forms.py
becky-gilbert: How about something like: "Please enter jsPsych experiment code. See the \<a href="jspsych-docs-link">CHS jsPsych documentation\</a> for more information about how to add your code and what plugins are available."

becky-gilbert: I'll make a note in the looki-docs jsPsych issue about adding this link once the docs page exists!

web/urls.py
becky-gilbert: FYI, I added an issue about different URL formats based on experiment type: #1325

PR #1277, opened 2023-09-28
Update github api to pull from EFP's master branch

### LINKED ISSUE #1275: EFP commit SHAs come from `develop` instead of `master`
**Description**

When creating a new internal study, the experiment version is automatically populated with the latest commit on the EFP `develop` branch. Also, when editing an internal study, the 'Check for updates' button retrieves commits from the EFP `develop` branch. The form's help text and our documentation say that the commits should come from the EFP `master` branch.

Screenshot 2023-09-26 at 4 52 24 PM

Screenshot 2023-09-26 at 4 41 21 PM

**How to reproduce**

1. Create a new internal study. After initially filling in and saving the Study Ad and Recruitment form and then being directed to the `study-details/efp` page, the default commit SHA for the experiment runner comes from the `develop` branch.
2. Go to the Study Details page for an existing internal study and click the 'Check for updates' button. The commits shown will come from the `develop` branch.

For reference:
- Commits on EFP master: https://github.com/lookit/ember-lookit-frameplayer/commits/master
- Commits on EFP develop: https://github.com/lookit/ember-lookit-frameplayer/commits/develop

**Expected behavior**

We say in the help text that the commits should come from the EFP `master` branch:
> If you're using the default Ember Frame Player, you can see [the commits page](https://github.com/lookit/ember-lookit-frameplayer/commits/master) for other commit SHA options.

> Since the version you are using, there have been updates to the master branch of https://github.com/lookit/ember-lookit-frameplayer. 

We could just change the help text to say `develop` instead of `master`, but I do think we should be using the `master` branch for releases.

**Additional context**

It's not immediately clear to me why this is happening since there is a `EMBER_EXP_PLAYER_BRANCH` variable in the settings file that defaults to `master`, and that variable is used in the code that constructs the URL that is used to retrieve the commit SHAs. There could be a problem with that bit of code (I haven't looked into it), or it could be that the value of this variable is `develop` rather than `master` on production. Another possible issue is that `develop` is the default branch for EFP, which is why you get the `develop` commits on the repo's commit page with no branch specified: https://github.com/lookit/ember-lookit-frameplayer/commits

So far I've only tested this on the production site.

okaycj:
Closes #1275

Github's API uses by default the default branch of a repo.  For EFP this was develop.  Updated API calls to specifically use the master branch.

### DIFF
--- a/web/static/js/efp-runner.js
+++ b/web/static/js/efp-runner.js
@@ -1,3 +1,6 @@
+// Get global variables from html
+const DATA = document.currentScript.dataset
+
 // Show the generator function field only if use_generator is checked.
 function updateGeneratorDisplay() {
     const generator = document.querySelector('[for=id_generator]').parentNode;
@@ -98,7 +101,7 @@ function updateCommitUpdateInfo() {
     const currentCommitDate = document.querySelector('#commit-description .date').innerHTML;
     const playerRepoUrl = document.querySelector('#id_player_repo_url').value;
     if (playerRepoUrl && currentCommitDate) {
-        const githubApiUrl = `${playerRepoUrl}/commits?since=${currentCommitDate}&sha=master`.replace('github.com', 'api.github.com/repos')
+        const githubApiUrl = `${playerRepoUrl}/commits?since=${currentCommitDate}&sha=${DATA.branch}`.replace('github.com', 'api.github.com/repos')
         httpRequest.open("GET", githubApiUrl, true);
         httpRequest.send();
     }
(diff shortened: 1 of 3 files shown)


web/static/js/efp-runner.js
```diff
@@ -98,7 +98,7 @@ function updateCommitUpdateInfo() {
     const currentCommitDate = document.querySelector('#commit-description .date').innerHTML;
     const playerRepoUrl = document.querySelector('#id_player_repo_url').value;
     if (playerRepoUrl && currentCommitDate) {
-        const githubApiUrl = `${playerRepoUrl}/commits?since=${currentCommitDate}`.replace('github.com', 'api.github.com/repos')
+        const githubApiUrl = `${playerRepoUrl}/commits?since=${currentCommitDate}&sha=master`.replace('github.com', 'api.github.com/repos')
```
becky-gilbert, line 101: There is an EFP branch variable in the project settings, `EMBER_EXP_PLAYER_BRANCH`. Do we want to insert that here instead of hard coding?

okaycj: Sure.  I'll make that update now.

becky-gilbert (approved):
Looks great, thank you!!

PR #1279, opened 2023-10-12
Diable json validation when using generator

### LINKED ISSUE #1043: Study edit form: disable JSON validation when protocol generator box is ticked
**Description**

Not sure if this qualifies as a bug, it's a little confusing for users (researchers). On the study edit page, a warning is shown if the Protocol Configuration box does not contain valid JSON, _even if the Protocol Generator box is ticked_. So if the user deletes everything from the Protocol Configuration box (for instance because they want to cut/paste their study JSON into a Protocol Generator function), then they'll see an error on the page about invalid JSON and will not be able to save the study.

**How to reproduce**

On a study edit page:
1. Tick the "Use protocol generator" box.
2. Delete all text inside the Protocol Configuration box, or add a typo to invalidate the JSON string.

**Expected behavior**

I think the JSON validation for the Protocol Configuration box shouldn't run when the "Use protocol generator" box is ticked. The Protocol Configuration box could also be disabled. This should make it clearer to users that only one of the two methods should be used to define the study protocol - not both.

Screenshot 2022-12-01 at 1 53 26 PM

okaycj:
Closes #1043 

In the EFP config, it was confusing to have to validate JSON when the generator is used.  

To resolve this, we moved the "use_generator" checkbox to the top of the form.  When it's not checked, the JSON protocol field is visible and the generator field will NOT be validated.  The reverse is true when the "use_generator" checkbox is checked.  

out

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -424,8 +424,8 @@ class EFPForm(ModelForm):
     class Meta:
         model = Study
         fields = (
-            "structure",
             "use_generator",
+            "structure",
             "generator",
             "player_repo_url",
             "last_known_player_sha",
@@ -441,7 +441,12 @@ class EFPForm(ModelForm):
 
     def clean_structure(self):
         try:
-            structure = json.loads(self.cleaned_data["structure"])
+            # Validate structure if not using generator
+            if not self.cleaned_data["use_generator"]:
+                structure = json.loads(self.cleaned_data["structure"])
+            else:
+                structure = {}
+
             structure["exact_text"] = self.cleaned_data["structure"]
             return structure
         except json.JSONDecodeError:
@@ -456,7 +461,10 @@ class EFPForm(ModelForm):
             if not generator.strip():
                 generator = DEFAULT_GENERATOR
 
-            js2py.eval_js(generator)
+            # Validate generator only if using generator
+            if self.cleaned_data["use_generator"]:
+                js2py.eval_js(generator)
+
             return generator
         except js2py.internals.simplex.JsException as err:
             raise forms.ValidationError(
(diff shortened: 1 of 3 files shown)


mekline:
I like this change in general, but also yesterday we had a user report about weird behavior regarding which code gets used, generator or vanilla (did you see this on slack?)  Can you potentially incorporate testing for that bug into this fix?

studies/forms.py
```diff
@@ -441,7 +441,12 @@ class Meta:
 
     def clean_structure(self):
         try:
-            structure = json.loads(self.cleaned_data["structure"])
+            # Validate structure if not using generator
+            if not self.cleaned_data["use_generator"]:
+                structure = json.loads(self.cleaned_data["structure"])
+            else:
+                structure = {}
+
```
becky-gilbert, line 449: Does this mean that any JSON in the protocol box will be overwritten if the researcher selects 'use generator' and saves? Probably not a big deal if so, but just checking because I think the way it works now is that any JSON in the protocol box is still saved when the researcher chooses to use the protocol generator. 
Would it work to just remove the `else` here or put the next few lines inside the `if`, or would that break things?

becky-gilbert (approved):
LGTM! 
Just one question about whether or not we save (or could save) any existing JSON structure when the generator is used. 
And I'm now wondering whether we're consistent with how this works the other way around, i.e. do we save anything that was previously added to the protocol generator box if the researcher switches to using the protocol structure?

studies/forms.py
okaycj: The code the user enters is retained in the "exact_text" key of that dictionary.  This means there is no JSON for EFP to use, but if the user selects "use_generator" and saves, then the JSON is translated to dictionary, validated, and is available for EFP.  

We need the empty dictionary on 448 to hold the "exact_text" on 450.

becky-gilbert: Ok thanks!

Issue #1285, opened 2023-10-26
Child Birthdays aren't required


okaycj:
**Description**
While the form requires a birthday, the database table `accounts_child` is allowed to set this field to null.

**How to reproduce**
The model has `null=True` for the child's birthday.

**Expected behavior**
It's not breaking anything, but we should probably have this requirement at the database level.

becky-gilbert:
Closed via #1300

PR #1300, opened 2023-11-03
Add eligibility to response model - fixes #1292

### LINKED ISSUE #1292: Add redtext warning status to researcher session data
# TL;DR

As a researcher, I want to know if the child was eligible for the study *at the moment they started it*  Happily, this is identical to the logic performed to generate eligibility and resulting warnings on the search and study-detail pages.  Can this information be injected into the Lookit session info so it's saved and included in the resulting data tables? 

Desiderata - 

- Add a new column to the session overview data download, named "eligibility"
- Levels should be "Eligible", or which of the messages was shown, e.g. "Ineligible_TooYoung", "Ineligible_TooOld", "Ineligible_Other" - or whatever possible values that redtext display can take on. 

Acceptance criteria:
- [x] Add a new 'eligibility' field/column to the Response model/table
- [x] New response objects should get a value for the 'eligibility' field, which is a list with one or more of the 5 eligibility values (existing responses will not be changed - their values will be empty lists)
- [x] Add explanation of this value to the data documentation: https://github.com/lookit/lookit-docs/pull/367
- [x] Add tests

becky-gilbert:
This PR adds a new 'eligibility' field to the Response model. The value is determined when the Response object is first created, and it is a list of one or more strings:
- Eligible
- Ineligible_TooYoung
- Ineligible_TooOld
- Ineligible_Criteria (based on the study's criteria expression)
- Ineligible_Participation (based on the study's must have/not have participated requirements)

The eligibility value will only be set for newly-created Response objects. Existing responses will have an eligibility field that contains an empty list.

The 'eligibility' field value roughly corresponds to the 'red text' warnings displayed under the "Participate/Preview Now" button, which let the participant know that the selected child is too young/old or does not meet the study criteria. However, there are some minor differences between these:
- If a child is too old for a study, then they are only shown the 'too old' warning text (even if they are also ineligible based on criteria/participation).
- A child is shown the same warning when they are ineligible based on the study criteria expression and/or the prior participation requirements.

# Examples

## Eligible

Participation page: 

eligible_warning

Data:

eligible_data

Study Responses -> Individual Responses -> Response details:

eligible_details

## Ineligible: too young

Participation page: 

ineligible_young_warning

Data:

ineligible_young_data

Study Responses -> Individual Responses -> Response details:

ineligible_young_details

## Ineligible: criteria

Participation page: 

ineligible_criteria_warning

Data:

ineligible_criteria_data

Study Responses -> Individual Responses -> Response details:

ineligible_criteria_details

## Ineligible: multiple reasons

Participation page: 

ineligible_mulitple_warning

Data:

ineligible_multiple_data

Study Responses -> Individual Responses -> Response details:

ineligible_multiple_details

## Existing responses 

The 'eligibility' column will be present but will contain an empty list.

eligibility_existing


becky-gilbert:
Questions about this issue: what should the value of the `eligibility` field be in these cases?:
1. **Child's birthday is missing, which means we can't calculate age range eligibility**. At the moment, the `eligibility` value is blank for existing responses in the database. Does it also make sense to use a blank value when the child's birthday is missing? Or do we need another 'Ineligible'-type category?

2. **Study's criteria expression is invalid (parsing error).** Should we default to always treating the criteria check as eligible (since otherwise there will be no eligible children for the study)? Or always code this as 'Ineligible_Criteria'?

okaycj:
> * **Child's birthday is missing, which means we can't calculate age range eligibility**. At the moment, the `eligibility` value is blank for existing responses in the database. Does it also make sense to use a blank value when the child's birthday is missing? Or do we need another 'Ineligible'-type category?

There's an issue on this https://github.com/lookit/lookit-api/issues/1285.  If you look through the production database, all children have a birthday.  This require for a birthday was enforced at the form level and probably should be at the database level.  Assuming, we want to require a child to have a birthday.  

> * **Study's criteria expression is invalid (parsing error).** Should we default to always treating the criteria check as eligible (since otherwise there will be no eligible children for the study)? Or always code this as 'Ineligible_Criteria'?

It my preference to not have an invalid criteria.  Can we do a better job of validating the criteria on form submission?  For existing invalid criteria, could we signal the researcher in someway?  Maybe on the preview child selection view?

studies/helpers.py
```diff
@@ -94,6 +100,56 @@ def repl(match):
...
+class ResponseEligibility(models.TextChoices):
+    """Participant eligibility categories for Response model"""
+
+    # This must be determined at the point of the study participation/response, because child eligibility can change over time for a given study
+    ELIGIBLE = "Eligible"
+    INELIGIBLE_YOUNG = "Ineligible_TooYoung"
+    INELIGIBLE_OLD = "Ineligible_TooOld"
+    INELIGIBLE_CRITERIA = "Ineligible_CriteriaExpression"
+    INELIGIBLE_PARTICIPATION = "Ineligible_Participation"
+
```
okaycj, line 165: TextChoices!  I should've known about this, very nice.

okaycj (commented):
This is just a draft, but it seems to make sense.  I'll do another review when you're ready.

becky-gilbert:
It turns out that we do require a birthday on the Child create form, and we validate the criteria expression on the study create/edit form.

I've updated the Child model so that `blank` is false and `null` is false. This ensures that a Child instance cannot be saved to the database without a birthday, which means that we don't need to validate the data that's coming out of the database (i.e. checking that a Child has a birthday before trying to access that field).

Similarly, I've added a default blank (`''`) value to the Study model's `criteria_expression` field. This prevents our testing tools from generating random text strings in this field when we create a new Study instance where the `criteria_expression` is not specified.

mekline:
Couple of comments from Ian on typos, but I think they are in the PR text rather than the codebase: 

I found a few quick typos:
-"Eligibile" string should be "Eligible". Looking at the code examples in the comments, the actual code probably doesn't have this error and it's just in this documentation.
- "If a child is shown the same warning when they are ineligible based on the study criteria expression and/or the prior participation requirements." should remove the "If" at the beginning I believe.

becky-gilbert:
Thanks Melissa and Ian! I checked the codebase and these errors are just in the PR text. Fixed! 👍

bleonar5 (approved):

okaycj (approved):
LGTM!  That was a lot of work.

Issue #1083, opened 2023-02-01
Missing Security-Related HTTP Headers


okaycj:
**Description**
During the engagement, LMG identified that web service responses did not contain one or more of the following security-related HTTP header flags:
- Content-Security-Policy
- X-Permitted-Cross-Domain-Policies 
- Cross-Origin-Resource-Policy
- X-XSS-Protection

Web administrators can use Content-Security-Policy (CSP) as an added layer of security that helps the web server detect and mitigate certain types of attacks, including XSS and data injection attacks. There are currently three header options that define allowed content types and the ability to define sources of trusted JavaScript.

To ensure the website is calling the cross-domain file that the site’s root has defined, administrators should implement the X-Permitted-Cross-Domain-Policies header. This is similar to a crossdomain.xml policy file and restricts/allows locations where Flash content can be accessed.

The Cross-Origin Resource Policy (CORP) header can be set to allow websites and applications to block certain requests from other origins such as <script> and < img>. This can mitigate side-channel attacks, like Spectre and XSS attacks.

**Additional context**
LMG recommends that system administrators add the following HTTP headers as required. Typical options include:
- Content-Security-Policy
- X-Permitted-Cross-Domain-Policies: master-only
- Cross-Origin-Resource-Policy
- X-XSS-Protection

**LMG Document**
https://drive.google.com/file/d/18vNuBbHOTZCrTZYvgz7oAbVwzfAJwsdd/view?usp=share_link

PR #1303, opened 2023-11-07
Django CSP

okaycj:
This is for Issue #1083

From the security review, it was requested to add CSP to our site.  This for of protection will allow us to limit what is able to be loaded/executed on our site.  

The [Django-CSP](https://django-csp.readthedocs.io/en/latest/) has been added and implemented.  The following has yet to be completed:

- QA.  Each view has different complications with this security restriction.  Most often it'll be inline style or scripts. 
- Authorize some JavaScript libraries (most notably google's tag manager) using a [nonce](https://django-csp.readthedocs.io/en/latest/nonce.html?highlight=nonce).  
- Others.  CSP has enough of a learning curve that its not entirely certain what might come up.

It might be worth noting that Chrome seems to have a slightly less restrictive implementation of this feature then firefox.

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -65,7 +65,7 @@ class TOTPField(forms.CharField):
     def widget_attrs(self, widget):
         """Override - used to update widget attrs in Field initializer."""
         attrs = super().widget_attrs(widget)
-        return {**attrs, "placeholder": "123456", "style": "width: 50%;"}
+        return {**attrs, "placeholder": "123456", "class": "w-50"}
 
 
 class TOTPCheckForm(forms.Form):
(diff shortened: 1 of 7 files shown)


bleonar5:
You may be aware of this already, but here are at least three instances of inline javascript in the app:

studies/templates/studies/_image_display.html
web/templatetags/web_extras.py
studies/templates/studies/study_detail.html

okaycj:
> You may be aware of this already, but here are at least three instances of inline javascript in the app:
> 
> studies/templates/studies/_image_display.html web/templatetags/web_extras.py studies/templates/studies/study_detail.html

Yup!  I was aware of at least one of these, but need to put this down to start working on other things.

okaycj:
There's been no movement on this PR.  Closing.

PR #1305, opened 2023-11-09
Add country + state from demographic data to session data on consent ruling page (Closes #1291)

### LINKED ISSUE #1291: Add info to consent frame to assist with account detection
# TL;DR

When a researcher needs to request review of a participant, the necessary information should be easy for them to find. The earliest in the pipeline we can do this is at consent!  Add some fields to make this possible.

This is designed to make it very straightforward to fill out this form: https://docs.google.com/forms/d/e/1FAIpQLSd_g32qCLgwjmYZs8zrllJtNRVHYtjXk6CNliRtM41UeAyZqQ/viewform?usp=sf_link

# Narrative
As a researcher I want to be able to report everything I need to without hunting through multiple data tables

This is the current info made available at consent:

Screenshot 2023-10-30 at 12 44 04 PM

To the parent account line, add the country, *and state, for US location** that is listed in the most recent demographics form.

bleonar5:
This PR is in response to issue #1291 

Now, on the consent ruling page for a study, when a user clicks on a given consent to rule on, the user's location will be collated with the rest of the info under "Session Data"

The changes required for this were fairly minimal. The consent_ruling data query already grabs "demographic_snapshot_id" from the response row, so I just had to add "demographic_snapshot" when calling "select_related" for "child" and "child__user".

Screenshot 2023-11-09 at 3 43 45 PM

### DIFF
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -794,6 +794,9 @@ class StudyResponsesConsentManager(
                         "hashed_id": hash_participant_id(response),
                         "uuid": str(response.pop("child__user__uuid")),
                         "nickname": response.pop("child__user__nickname"),
+                        "country": response.pop("demographic_snapshot__country"),
+                        "state": response.pop("demographic_snapshot__state")
+
                     },
                     "child": {
                         "hashed_id": hash_child_id(response),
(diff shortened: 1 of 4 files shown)


becky-gilbert (changes requested):
LGTM! Two things:
1. Try running `make lint` to fix the linting errors. (This is actually a pre-commit hook for me, which is really useful. Not sure what you might have to do to set that up?)
2. I assume this works fine when the Country/State values are missing? I'm pretty sure they're blank and not None, but it might be worth checking if you haven't already.

bleonar5:
applied linting changes and confirmed that country + state are blank, not None, in the database model. Going ahead and merging

Issue #1306, opened 2023-11-09
Prior participation eligibility: must meet criteria for all studies in each list


becky-gilbert:
# TL;DR

If there are multiple studies in the 'must have participated' / 'must not have participated' lists for study eligibility, the child must meet the criteria for ALL of studies in the list.

# Narrative

As a researcher, if I add multiple studies to the 'must have participated' study criteria, I expect that children will only be eligible if they have participated in all (not just some) of the studies in the list. Similarly, if I add multiple studies to the 'must not have participated' list, I expect that children will only be eligible if they have not participated in any of the studies in the list.

# Acceptance Criteria

- [x] Participation eligibility checks will treat multiple studies with AND (rather than OR) logic, i.e. must have participated in all studies in the required list, must not have participated in any studies in the disallowed list
- [x] Tests
- [x] Make sure this is clear in documentation: https://github.com/lookit/lookit-docs/pull/365

Issue #1302, opened 2023-11-06
Prior participation eligibility: ignore empty responses, add tests


becky-gilbert:
# TL;DR

When checking the prior study participation criteria, when the study in question is internal, we should ignore empty responses (i.e. sequence = []). We should also add more tests around the eligibility checks/warnings.

# Narrative

We recently added the ability for researchers to add prior study participation criteria to their studies, and we included this in the set of checks that are used to produce the 'ineligible' warning that appears on the participation page when an ineligible child is selected. However, the 'ineligible' warning text appears when the child has an 'empty' study response (e.g. they click 'participate now', which generates an empty response object, but then they immediately end the study). This is confusing for families.

We also need to make sure that we're testing all edge cases (e.g. empty responses) and combinations of study eligibility criteria (age range, criteria expression, prior participation). These tests can be done specifically on the study eligibility logic, and/or incorporated into the existing tests for things that are affected by eligibility (e.g. recruitment emails, study list view with child filter, eligibility warning text).

# Acceptance Criteria

- [x] Study participation checks should ignore responses from internal studies that are totally empty
- [x] Tests: 
  - Empty responses for internal studies should not count as having 'participated'
  - Partial/incomplete and completed responses for internal studies do count as 'participated'
  - Responses to external studies are always counted as 'participated'
  - ~~Check that these things are true anywhere the study criteria checks are used: recruitment emails, study list view, study-details view (warnings)~~ (I decided not to add these tests because they use the same few functions that are now well tested)
- [x] Alternatives Considered: we considered using the response's `completed` field in the prior participation checks, but decided against it because it's not a reliable flag and there's no corresponding flag for external studies. Researchers can choose to implement their own checks using the protocol generator function (past sessions data).
- [x] Documentation: in the section about study create/edit fields, explain how prior participation is checked, including edge cases, and point researchers to alternatives (e.g. get past sessions from protocol generator): https://github.com/lookit/ember-lookit-frameplayer/pull/350, https://github.com/lookit/lookit-docs/pull/365

becky-gilbert:
Closed via #1307

PR #1307, opened 2023-11-14
Prior participation eligibility logic (#1302, #1306)

becky-gilbert:
This PR makes a few changes to the 'must have/not have participated' Study eligibility criteria:

- When the study has more than one required (must have participated) study, the child must have participated in ALL of these in order to be eligible (#1306)
- When checking the child's eligibility, we now ignore empty (sequence=[]) responses to internal studies (#1302)
- Adds tests for more complex eligibility cases (e.g. study has multiple required or disallowed prior participation criteria, study has both required and disallowed participation criteria, required/disallowed studies are external/internal and responses are empty/non-empty)

### DIFF
--- a/accounts/queries.py
+++ b/accounts/queries.py
@@ -127,9 +127,7 @@ def get_child_participation_eligibility(child, study) -> bool:
             i
             for i in study.must_have_participated.all()
             if child.responses.filter(study=i)
-            .exclude(
-                study__study_type_id=StudyType.get_ember_frame_player().id, sequence=[]
-            )
+            .exclude(study__study_type=StudyType.get_ember_frame_player(), sequence=[])
             .exists()
         ]
 
@@ -143,9 +141,7 @@ def get_child_participation_eligibility(child, study) -> bool:
             i
             for i in study.must_not_have_participated.all()
             if child.responses.filter(study=i)
-            .exclude(
-                study__study_type_id=StudyType.get_ember_frame_player().id, sequence=[]
-            )
+            .exclude(study__study_type=StudyType.get_ember_frame_player(), sequence=[])
             .exists()
         ]
         # only true if response does NOT exist for any one (or more) of the studies in the list
--- a/accounts/queries.py
+++ b/accounts/queries.py
@@ -118,14 +118,24 @@ def get_child_participation_eligibility(child, study) -> bool:
     must_not_have_participated = True
 
     if study.must_have_participated.exists():
-        must_have_participated = child.responses.filter(
-            study__in=study.must_have_participated.all()
-        ).exists()
+        must_have_participated_all = [
+            i
+            for i in study.must_have_participated.all()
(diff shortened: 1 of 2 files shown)


mekline:
Just a reminder (guessing already covered, but just to make sure) that the negative criteria are still correct and different from the positive direction (which is probably how this implementation happened in the first place!)  Whitelist studies should require all studies (AND) completed, Blacklist studies should block you if at least one study (OR) was completed

mekline (approved):
See comment on the PR but otherwise LGTM!

becky-gilbert:
> Whitelist studies should require all studies (AND) completed, Blacklist studies should block you if at least one study (OR) was completed

Thanks for checking! Yes that's how it works now.

accounts/queries.py
```diff
@@ -118,14 +118,24 @@ def get_child_participation_eligibility(child, study) -> bool:
...
 
     if study.must_have_participated.exists():
-        must_have_participated = child.responses.filter(
-            study__in=study.must_have_participated.all()
-        ).exists()
+        must_have_participated_all = [
+            i
+            for i in study.must_have_participated.all()
+            if child.responses.filter(study__exact=i).exists()
+        ]
```
okaycj: I am not going to hold up the PR for this, but `__exact` seems needless.  Are you finding the query is returning different result with and without it?

accounts/queries.py
```diff
@@ -114,15 +114,25 @@ def get_child_participation_eligibility(child, study) -> bool:
     Returns:
         bool: Return true if child is eligible based on their prior study participation
     """
+    # Need to import StudyType here due to circular import problems
+    from studies.models import StudyType
+
```
okaycj, line 119: Nice.

accounts/queries.py
```diff
@@ -114,15 +114,25 @@ def get_child_participation_eligibility(child, study) -> bool:
...
     if study.must_have_participated.exists():
+
         must_have_participated_all = [
             i
             for i in study.must_have_participated.all()
-            if child.responses.filter(study__exact=i).exists()
+            if child.responses.filter(study=i)
+            .exclude(
+                study__study_type_id=StudyType.get_ember_frame_player().id, sequence=[]
+            )
```
okaycj, line 132: This might be a style choice, but why not just use `study__study_type=StudyType.get_ember_frame_player()` instead?

okaycj (approved):
LGTM!  Just a couple of questions.

accounts/queries.py
becky-gilbert: You're right, `__exact` isn't needed. I think the code that you're commenting on here might be outdated, because I think I fixed all of these. I'll double check and remove any that I missed.

accounts/queries.py
becky-gilbert: Because I didn't think of it 😅 Your suggestion is better - I'll fix it!

PR #1308, opened 2023-11-14
Generator JavaScript validation 

okaycj:
The existing JavaScript validation is restricted to ES5.  This has proven to be an issue with our experimenters, as well our developers.  To resolve this, I've removed the js2py package and moved JS validation with client side.  

On form submission, JS will grab the current script, run it through JSHint, and report any errors it finds back to the user.  JSHint's default behavior appears to be fairly strict.  I updated the default JSHint config to be able use ES6 (which is the point of this PR), but if it turns out to be more restrictive then we'd like [the config can be updated](https://jshint.com/docs/options/).  

Screenshot 2023-11-14 at 11 40 33 AM


becky-gilbert (approved):
I tried this out and it's great! It definitely accepts ES6 features now (yay!) and it correctly flagged the test errors that I tried out. Thanks!

PR #1310, opened 2023-11-20
Move changes to production

becky-gilbert:
This PR moves a few changes from staging into production:

- Protocol generator validation: switch from server-side (js2py package) validation to client-side JSHint: #1308 
- Add country/state to session info displayed on consent ruling page: #1305
- Add new 'eligibility' field to Response model: #1300 
- Fixes to 'prior participation' eligibility criteria (must have participated in ALL required studies, must not have participated in ANY blacklist studies, ignore empty responses to internal studies): #1307 
- Add lookit as project author: #1304


becky-gilbert:
Hi all, I just checked all of these changes on staging and it's all looking good to me. 

One thing that might confuse researchers is that the new ES6-friendly JavaScript validator for the protocol generator function is a little stricter than what we had before, so if the researcher goes to their study details page and tries to save it, they may see some validation errors (even if they haven't touched the code!). The errors are output in a nice readable list with line numbers, so it should be easy to make the necessary changes. We can mention this in the release announcement, along with links to the new documentation about the response eligibility field and prior participation eligibility criteria.

Are we adding anything else to this production release, or is it ready to go?

mekline:
Good point about the errors; Becky can you make the announcement so you draft the details on this for researchers to be aware of?

mekline (approved):

Issue #1312, opened 2023-11-27
JS validator errors don't appear on study detail page


becky-gilbert:
# TL; DR

When there are JS validation errors on the study EFP edit page, they are not always showing up because the error list is appended to the breadcrumb DOM element, and the breadcrumbs are sometimes missing from the study pages.

# Description

We recently switched to a new JS validator (JS Hint) for the protocol generator function. When the researcher tries to submit the study details form, the validator runs on the client side. If there are any errors, the researcher stays on the page (i.e. form is not submitted, changes are not saved) and the errors should be written to the top of the page. 

However, the error list is added after the breadcrumb element, and there are problems with missing breadcrumbs on the researcher study pages. When the breadcrumb element is missing, the error list cannot be added to the page. This means that some researchers are unable to edit their protocol generator code because it is not validating, and they cannot see/fix the errors.

## Missing breadcrumbs on production and staging

![production_missing](https://github.com/lookit/lookit-api/assets/9041788/f8485d5f-cd52-4952-9f99-abb9fade06af)

![staging_missing](https://github.com/lookit/lookit-api/assets/9041788/1d9c2cbb-ae20-4af5-ae61-d5feb7ac7619)

## The problem is inconsistent 

Sometimes the breadcrumbs do appear on production/staging (or show up as plain text rather than links), even for the same study:

![staging_broken_1](https://github.com/lookit/lookit-api/assets/9041788/7a4a7830-ce71-4509-9f49-549d0eb6c009)
![staging_broken_2](https://github.com/lookit/lookit-api/assets/9041788/0679c9fb-a3a5-4907-a191-e407644e956f)
![staging_working](https://github.com/lookit/lookit-api/assets/9041788/734a6c7e-43a9-4a49-9cec-db7e1020a48c)

I've looked at a few studies on localhost and they breadcrumbs all seem to be working:

![localhost_working](https://github.com/lookit/lookit-api/assets/9041788/ad5de1c1-30c5-4507-aa99-0cc7bd4a192f)

**Environment (OS, browser, branch, etc.):**

Tested on production/staging/local and on Firefox. The breadcrumbs are sporadically broken/missing on staging and production. I think they work more reliably on localhost?

# Solutions

Temporary fix for researchers: 
1. Paste the protocol generator code into the JS Hint web app: https://jshint.com/
2. Fix anything listed under 'errors' or 'warnings'
3. Paste modified code from JSHint back into the CHS site protocol generator box and save

CHS team:
[...]

PR #1318, opened 2023-11-29
Fixed issue with js errors not showing in ui.

okaycj:
There's a bug in production where the breadcrumbs aren't being displayed.  This validation was using the breadcrumbs element as a reference to place the JS errors on the EFP runner edit form. This removes the need for the breadcrumbs reference.

### DIFF
--- a/studies/templates/studies/experiment_runner/base.html
+++ b/studies/templates/studies/experiment_runner/base.html
@@ -14,6 +14,8 @@
 {% endbreadcrumb %}
 {% endblock breadcrumb %}
 {% block content %}
+    {% block js_error %}
+    {% endblock js_error %}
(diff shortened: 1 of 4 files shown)


becky-gilbert (approved):
Thanks CJ, this looks great! I like this approach much better (using a dedicated error block to show the errors, instead of attaching them to something unrelated). 
As I mentioned in #1312 I would still like to fix the breadcrumb errors, but at least now the two things are separate and we can get this validation error fix in production ASAP!

okaycj:
> As I mentioned in #1312 I would still like to fix the breadcrumb errors, but at least now the two things are separate and we can get this validation error fix in production ASAP!

@becky-gilbert I put up a PR to fix the breadcrumb issue.

PR #1319, opened 2023-11-29
Breadcrumbs not displaying

okaycj:
The breadcrumb template function that we developed ended up being a bit problematic.  The template formatter didn't do a good job with the template tag, the syntax to implement the function was ugly, and it was prone to developer errors.  Removing it seemed to be the best solution.


becky-gilbert (approved):
I'm a little sad that our custom template tag is a failure and that we have duplicated code now (SonarCloud is sad about that too!). But yes, this looks like the best fix. 
APPROVE! 🚀

PR #1330, opened 2023-12-14
Fix data availability with survey-consent

### LINKED ISSUE #1317: Responses with consent via `survey-consent` frame are not available to researchers
**Description**

Responses to studies that use the `survey-consent` frame for consent are not becoming available to researchers, because they are not marked as having a consent frame (all are `completed_consent_frame: false`) and therefore they do not become available to researchers for a consent ruling. This is true despite the frame being correctly identified as consent (`frameType: "CONSENT"`). 

**Expected behavior**

Responses that include the `survey-consent` frame should become available to researchers.

@mekline questions for you about how consent approval should work with `survey-consent`: 
1. Should all responses with a `survey-consent` frame appear in the study's 'Review Consent' section (even though at that point the researcher has no way of seeing the response to the consent question)?
2. Should all responses with a `survey-consent` frame appear in the study's 'Study Responses'  section (regardless of how the participant responded to the consent question)?
3. Do we need to add in some code to automatically extract the response to `survey-consent` frames and either (a) present it to the researcher in the 'Review Consent' section or (b) have the lookit-api check the value and use that to automatically move consented responses to 'Study Responses' section? This option will take longer to implement, and (b) will take longer than (a).

**Implementation notes**

The Study model's `judgeable_responses` property just filters the study responses on `completed_consent_frame: true`, and then the responses that are available to researchers come from the subset of judgeable responses that have been approved. So I think we just need to get the responses with survey-consent frames to be marked as `completed_consent_frame: true`. There might also be an issue with filtering responses based on the presence of video for the purpose of detecting/judging consent.

becky-gilbert:
Fixes #1317

This PR flags any responses that use the survey-consent frame on the consent manager page. Specifically: 

- Adds a new boolean field called `survey_consent` to the Response model. The default value is `false` and it is set to `true` by the survey-consent frame in EFP (https://github.com/lookit/ember-lookit-frameplayer/pull/354).
- Marks responses in the list on the left with [Survey consent] (similar to how preview responses are marked)
- Adds some yellow warning text in the 'video information' area explaining that the response contains survey-consent, and some guidance for how the researcher should handle the consent verification. (Note that the survey-consent information is independent from the 'no video found' warning because responses with survey-consent may or may not also include consent/assent video).

I made a few other UI changes while working on the consent manager page:

- Put all the elements in the video section (middle column) in a card with a 'video' heading, to match the other columns
- Add a 'comments' label for the comments section (middle column, below the video area)
- Highlight the entire 'video information' box above the video, rather than just the text, when there is a warning
- In the response list on the left, remove the 'N/A' when there are no comments
- Add a hover style (slight background color change) to the responses in the list on the left, to make it clearer that they are clickable elements

Screenshots:

Initial page with no response selected:

consent-no-selection

Response selected that contains a survey-consent form and no video found:

consent-survey-consent-update

Response selected with no video found:

consent-no-video-update

Response with a comment (I removed the 'N/A' text when there's no comment, but actual comments still show up as italicized text, as before):

consent-comment-update

More screenshots and a video here: https://github.com/lookit/lookit-api/issues/1317#issuecomment-1853349186


web/static/js/study-responses-consent-ruling.js
```diff
@@ -56,7 +57,7 @@ $(document).ready(function () {
 
     function saveComments() {
         let currentText = $responseComments.val();
-        if (currentText) {
+        if (currentText && $currentlySelectedResponse) {
```
okaycj, line 60: Correct me if I am wrong, but I think you have call `val()` or check the `length` of a jQuery object to check it's truthiness.

web/static/custom_bootstrap5.css
```diff
@@ -24,6 +24,9 @@ tr.preview-row {
   -moz-column-count: 2;
   -webkit-column-count: 2; }
 
+.list-group-hover .list-group-item:hover:not(.active) {
+  background-color: #f5f5f5; }
+
```
okaycj, line 29: Is creating a custom color to onerous?

okaycj (approved):
LGTM!  Just a couple of questions that aren't worth holding up moving this to develop.

web/static/custom_bootstrap5.css
becky-gilbert: Great idea. Done!

web/static/js/study-responses-consent-ruling.js
becky-gilbert: That's probably true in some cases, but here I get `undefined` for `$currentlySelectedResponse` when no response is selected from the list on the left. So just check whether or not it's undefined is giving me the result I need here (only check for a comment to save if a response is selected), and trying to call `val()` or `length` on that variable produces an error when no response is selected.

PR #1341, opened 2024-01-18
Is spam admin

### LINKED ISSUE #1311: Add "is_spam" and admin freetext field to CTRL interface
For spam management, give superusers two new fields to assist with tracking & blocking spam.  Desiderata:

- A checkbox/boolean for "is_spam."  Accounts that are flagged as spam should *also* be set to inactive and to "no" for all four email settings, but it's okay if the person reviewing spam needs to uncheck those boxes manually (IE, please *don't* pursue adding those updates unless they are also just a matter of changing something in the config file for CTRL.)

- An open text field for  "admin_comments" that's available for read/write by admins with access to CTRL. 

Interface notes: 

On the search view (e.g. https://lookit.mit.edu/__CTRL__/accounts/user/?q=struhl), it would be great to remove Given Name and Family Name (which hardly any family accounts ever fills out) from the table and instead include the new is_spam feature along with the email address

On the details view (e.g. https://lookit.mit.edu/__CTRL__/accounts/user/10317/change/?_changelist_filters=q%3Dstruhl), no strong preference on the ordering of either of these fields, but it would make sense to have them at the bottoms since we scan the full record when making the determination.   Something like this would be great: 

-- all the other fields ---
Locale: _____________
[ ] Is researcher
[ ] Email next studies, etc.
Admin comments:  _____________
[ ] Is spam

okaycj:
Here's the spam update to the CTRL interface for User accounts.  

Closes #1311

Screenshot 2024-01-18 at 4 30 03 PM

This update adds the "is_spam" boolean and "admin_comments" text field to the user model.  This will give those who have access to CTRL the ability to mark a user as a spammer and provide comments regarding this action.  What won't happen is the unsetting of the "is_active" boolean or the email options automatically.  It was decided to have these fields be manually and as it was a bit outside the normal case for the Django Admin views.  

An alternative to the above manual setting of fields could be an [Admin Action](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/actions/).  This action would be triggered from the drop down in the Users list view and could provide a easy way to set the is_spam field and unset the is_active and email fields.  This would not be able to capture any admin_comments.  The admin would have to go into the User's detail view and enter their comment directly into the field. 

Here is a screen shot of the Users list view in the admin interface to better show where the action would be triggered from. 

Screenshot 2024-01-18 at 4 35 52 PM

You might be able to imagine there's another selection below "delete selected users" that could stay "spammer" and it would do the above suggested actions.  Adding this action is fairly trivial.  

### Now that we're doing the above action idea, here's a screen shot of what it looks like for approval. 

out


becky-gilbert (approved):
LGTM!

becky-gilbert:
Quick question: these fields won't be visible in the user's account info or registration forms, right? I had a quick look at accounts/forms but couldn't see where we include/exclude fields from the User model. 

Oh and I like the 'Admin actions' option you suggested, with a 'mark as spam' action in addition to 'delete selected users'.

okaycj:
> Quick question: these fields won't be visible in the user's account info or registration forms, right? I had a quick look at accounts/forms but couldn't see where we include/exclude fields from the User model.

Yup, not visible.  The only User fields, that I can find, that are available is "username" and "nickname" in the "my account" view.

ianchandlercampbell:
@okaycj I got locked out of Slack and its not accepting my password or allowing me to reset-- did you just try to update my permissions?

becky-gilbert (approved):
Very nice! It's a little complicated (to me) and I appreciate all of the comments you added this all works.

If someone wants to undo an 'is spam' decision, they could still do this via the admin pages right? (Not with an admin action ,but just by clicking on individual users and resetting the various fields).

ianchandlercampbell:
This all looks good to me and should improve our spam review process quite a bit!

PR #1343, opened 2024-01-23
Allow removal of english specific content for home

rhodricusack:
On the home page, there are some graphics and content that is not relevant to a non-English speaking audience - for example, graphics and text corresponding to "Project Garden"

This commit allows for this to be removed:

- adding to the Project Garden div in web/templates/web/home.html a CSS style parameter with a variable style="{{ style_for_english_specific }}
- add custom method web.views.HomeView
- in this, set style_for_english_specific to 'display: block' if URL language code starts with en, otherwise makes it invisible

### DIFF
--- a/web/templates/web/home.html
+++ b/web/templates/web/home.html
@@ -73,7 +73,7 @@
              alt="University logos"
              src="{% static 'images/university_logo_grid.png' %}" />
     </div>
-    <div class="clearfix px-md-5 pt-5 mb-5">
+    <div class="clearfix px-md-5 pt-5 mb-5" style="{{ style_for_english_specific }}">
         <a href="{% url 'web:garden-home' %}">
             <img class="img-fluid mb-3 col-md-5 float-md-start me-md-3 img-hover p-2"
                  alt="Garden project logo"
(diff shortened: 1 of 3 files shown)


okaycj (changes requested):
At this time, we're not going to conditionally hide content based on language.

rhodricusack:
Could you unpack the logic a bit here, @okaycj? The specific thing we want to get rid of are graphics. We can get rid of irrelevant text, by translating to a null string. But all the tags make it hard to include full HTML  in here. It is just a little messy to have a figure (e.g., Project Garden) with no accompanying text.

ianchandlercampbell:
@rhodricusack Tagging in here to provide you a little bit more context. This isn't a "never" but a "not at the moment" decision. Although we think that pursuing many different options for specific localizations (e.g., showing a home page for Japanese users that shows studies specifically for them) is a great goal, we currently only have the capacity to support translation of existing content on the homepage.  There are many other things that could be done to support localizations-- for example we could also eventually have a highlighted Japanese study partner with a special banner. Right now, we have the capacity to support efforts to make current content readable and accessible for different languages but not to start creating specific pages (removing or adding content) for localizations. This may change in the future with further discussions between CHS and your team, but would be a larger discussion for another time.

rhodricusack:
Thanks, Ian, for taking us through the logic. Let us know when further discussions would be helpful!

PR #1344, opened 2024-01-23
Add filter-by-language to study list

rhodricusack:
There are many English studies on lookit and at present only one Japanese one. We are worried that on the studies list page, the mass of English studies will be off-putting and confusing, and rarely useful. So, this commit:

- adds a language field to each study (studies/models.py), giving the default of 'en'
- on the study list page web/templates/web/studies-list.html we then add a new drop-down to filter by language, currently we three options: All, English or Japanese. For English this defaults to All but if the URL country code is ja, it defaults to Japanese (thus filtering for Japanese studies). Note this is stored in a browser cookie and so will only be picked up the first time you hit the page.


accounts/forms.py
```diff
@@ -471,6 +471,11 @@ class StudyLocation(FormChoiceEnum):
         lookit = ("1", _("here on the Lookit platform"))
         external = ("2", _("on other websites"))
 
+    class StudyLanguage(FormChoiceEnum):
+        all = ("*", _("All languages"))
+        english = ("en", _("English"))
+        japanese = ("ja", _("Japanese"))
+
```
okaycj, line 478: Wow!  This is great.  The filter drop-down would need to have the languages of all available studies.

okaycj: How to do that, I am unsure at this moment.

okaycj (changes requested):
This is such a nice add.  Just the one issue to resolve.

accounts/forms.py
rhodricusack: At present, we believe there are only ja and en studies. But did you mean you'd like to automatically populate this list by querying the available studies?

okaycj: Correct. If someone were to create a Spanish study, they wouldn't be able to use this filter.

ianchandlercampbell: @okaycj is correct, if we want to create a language filter for studies (which is a great idea!), we should go ahead and build the capacity to do it for any of the other languages we support.

okaycj:
Please let me know if we need to work on this issue. I am assuming this PR can be closed.

PR #1342, opened 2024-01-23
Change consent video based on language

rhodricusack:
In the FAQ it shows an example of the consent video, which was in English. This commit internationalises this by:
- adding a Japanese video web/static/videos/consent_ja.mp4
- modifies the template web/templates/web/faq.html to take variables video_consent_mp4, captions_label, captions_src. 
The Japanese video does not currently have captions as we didn't have the resources to create these.
- creating a custom web.views.FaqView, which set variables that default to the English video unless the language code is 'ja' when it switches to the Japanese consent video

PR #1345, opened 2024-01-23
internationalisation and Japanese translations

rhodricusack:
Further small pieces of internationalisation and Japanese translation.

### DIFF
--- a/web/templates/web/resources.html
+++ b/web/templates/web/resources.html
@@ -11,7 +11,7 @@
 {% endblock title %}
 {% block content %}
     {% page_title "Resources" %}
-    <h3 class="subheading">Find a developmental lab near you</h3>
+    {% trans '<h3 class="subheading">Find a developmental lab near you</h3>' %}
     <div class="m-4">
         <p>
             Interested in participating in research in person? Find a list of labs that study child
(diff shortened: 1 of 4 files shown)


rhodricusack:
@okaycj Is there anything I can do to help with this and the other three requests #1342 #1343 #1344?

okaycj (approved):
This looks good!

rhodricusack:
Thank you!

web/templates/web/resources.html
```diff
@@ -11,7 +11,7 @@
 {% endblock title %}
 {% block content %}
     {% page_title "Resources" %}
-    <h3 class="subheading">Find a developmental lab near you</h3>
+    {% trans '<h3 class="subheading">Find a developmental lab near you</h3>' %}
     <div class="m-4">
```
okaycj, line 15: Sorry, I was taking another look at this PR and I noticed that we have HTML in the translations block.  It might be easier for translation if there wasn't HTML in the translation files.

okaycj (changes requested):
Sorry for backing out of the approved review.  Should be just the one change.

web/templates/web/resources.html
rhodricusack: Sure, done that!

okaycj: Thank you.

okaycj (approved):
Thank you again for the update.

PR #1346, opened 2024-02-13
Jspsych helpers

### LINKED ISSUE #1294: jsPsych studies should receive child/session info
# TL;DR

Researchers running jsPsych experiments should have access to the same child and session information that they currently do when using EFP using a protocol generator function.

# Narrative

As a researcher, I'd like access to certain bits of information about the child and their previous session data in the study code. This allows me to create an experiment that can change based on the child's information and their participation/performance history (e.g. for multi-session/longitudinal designs and adaptive difficulty methods).

# Acceptance Criteria

- [ ] Tests
- [ ] Scoping
- [ ] Documentation
- [ ] Should include, at a minimum, the same information and format that we use in the EFP's [protocol generator function](https://lookit.readthedocs.io/projects/frameplayer/en/latest/utils/protocol-generator.html#what-information-it-has-access-to)

okaycj:
Closes #1294

### Summary

Now that the lookit-helpers package is fairly close to being released to NPM, we can implement it in the JSPsych template.  

#### Additional Work

1. The existing linter for researcher code was proving to be difficult as we moved away from ES6.  A modular version of Babel is used instead.  
2. Added endpoint "past-sessions" to fulfill the need for researchers to know the past sessions (or responses) a user has done on their experiment.

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -13,7 +13,7 @@ router.register(r"labs", api_views.LabViewSet)
 router.register(r"feedback", api_views.FeedbackViewSet)
 router.register(r"video", api_views.VideoViewSet)
 router.register(
-    r"past-sessions", api_views.PassSessionsViewSet, basename="past-sessions"
+    r"past-sessions", api_views.PastSessionsViewSet, basename="past-sessions"
 )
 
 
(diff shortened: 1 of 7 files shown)


bleonar5 (approved):
Should the viewset be called PastSessionsViewSet?

becky-gilbert (approved):

PR #1347, opened 2024-02-20
Release to production

okaycj:
This PR is to move features already in staging to production.  These features include...

1. Adding assistance for spam:
    - #1341
3. Stepping closer to supporting jsPysch:
    - #1331 
    - #1346 
    - #1339


becky-gilbert (approved):
LGTM!! 🚀 
Two comments:
1. I think the jsHint change (#1339) is now irrelevant because of the switch to Babel in this same release, right? (Not that it matters to include it).
2. WRT this comment from the PR with the initJsPsych unpkg link(#1331):
> This PR will have to be updated as the final touches to the package are uploaded.
I think what you meant is that we'll need to update the version in the link when there are new releases, but that doesn't actually matter until we're using jsPsych in production, right?

okaycj:
> I think the jsHint change (https://github.com/lookit/lookit-api/pull/1339) is now irrelevant because of the switch to Babel in this same release, right? (Not that it matters to include it).

Yes.

> I think what you meant is that we'll need to update the version in the link when there are new releases, but that doesn't actually matter until we're using jsPsych in production, right?

That comment should have been removed.  The link was updated in [3eaf7fd](https://github.com/lookit/lookit-api/pull/1331/commits/3eaf7fdef74e9b846aab607b0ff6d241302b551f).

PR #1349, opened 2024-02-23
How we added One-click unsubscribe to emails

### LINKED ISSUE #1350: One-Click Unsubscribe
# TL;DR
SendGrid needs us to add a one-click unsubscribe link in our emails. 

# Narrative
For subscribed messages, enable one-click unsubscribe and include a visible unsubscribe link in the message body. 

# Acceptance Criteria

- [ ] Tests for new feature
- [ ] QA on new feature

okaycj:
### Summary

This PR adds a view that will remove all email preferences from the logged-in user.

### Details

Per the request of SendGrid, we needed to add a link to our emails that provides our users with a "one-click" way to unsubscribe from our automated emails.  This is in addition to the email link that's already there that gives users an easy way to update their email preferences.  

### Implementation

~~The simplest solution was to create a view that just updates the logged-in user's email preferences on the User model and adds that URL to the body of emails.  If the user clicks the link from their email and is not logged in, they will be asked to authenticate and are redirected to the view.~~ **This is no longer how we solved this issue.  See below for the new implementation.**   

### Concerns

Two concerns really, I don't think either of these cases is that big of a deal. 

1. Users accidentally click the link and are quickly and without notice unsubscribed.  They would have to go into their account preferences and click the checkboxes to re-up. 
2. Bad actors "provide" the link to people and they unknowingly unsubscribe themselves.  The consequences of this are very low and user-fixable.  

### One-Click Unsubscribe Implementation

Here's a step-by-step description of what happens when a user clicks the unsubscribe link.

1. The user clicks the unsubscribe link in their email.  The URL contains the username and token.  The token expires in two days.  Here's an example of the URL: `/account/okaycj@mit.edu/1rf4fP:2PweoVk6k92sRLHzZTcoIgbAChC8VHA8J61WFfgkgco/`
2. The page loads via GET.  The view contains a message that indicates we're in the process of unsubscribing them.  
3. Once the page loads, it'll trigger a post for the view using a minimal HTML form.  The form contains only a name, method, and the required CSRF token.
4. On POST, the view updates the user's email preferences, sets an appropriate message depending on whether the update was successful, and redirects to the home view.  I decided on the home view because it doesn't require the user to be logged in.  There is a link to update their email preferences in the success message. 

Screenshot 2024-02-28 at 12 48 22 PM
Screenshot 2024-02-28 at 12 48 42 PM
Screenshot 2024-02-28 at 12 48 55 PM

 
Closes #1350


ianchandlercampbell:
@okaycj looks good overall! Because this is a big change that will impact all users (even if it's an easy and straightforward change), let's talk quickly on Monday to think through any other implications. In the meantime, if @becky-gilbert and/or @bleonar5 can review from the technical side that would be fantastic!

becky-gilbert:
@okaycj my understanding is that one-click unsubscribe has to happen without requiring a login. 

Specifically, I've read that we can't require the user to do anything or provide any info other than:
> ...the recipient’s electronic mail address and opt-out preferences, or take any other steps except sending a reply electronic mail message or visiting a *single* Internet Web page

And: 
> when a subscriber clicks on an unsubscribe link and is confronted with a log in page because they have to log into their account in order to unsubscribe, that violates the “visiting a single Internet Web page” in order to opt out” requirement of the law

Source: https://www.isipp.com/blog/the-one-click-unsubscribe-law/

But I don't know what changes would be needed for that to work, so I'm not sure this should necessarily hold things up. This solution is still an improvement on what we have! 

WRT the GET vs POST question - my first thought was that GET should be fine, but then I read the comment below from someone about GET causing problems because browsers 'preload links' - not sure if/when this actually happens?!

> There is nothing wrong with that landing page in response to a GET request, but it is not a good idea.  Browsers that preload links for performance will be unsubscribing users without their knowledge.
For one-click unsubscribe, read the RFC and make sure that the URL responds to a POST request without requiring confirmation.

I think this is the relevant RFC section: https://datatracker.ietf.org/doc/html/rfc8058#section-3.1 

> The URI in the List-Unsubscribe header MUST contain enough information to identify the mail recipient and the list from which the recipient is to be removed, so that the unsubscription process can complete automatically.  Since there is no provision for extra POST arguments, any information about the message or recipient is encoded in the URI.  In particular, one-click has no way to ask the user what address or from what list the user wishes to unsubscribe.
> The POST request MUST NOT include cookies, HTTP authorization, or any other context information.  The unsubscribe operation is logically unrelated to any previous web activity, and context information could inappropriately link the unsubscribe to previous activity.

So it does seem like the people who wrote the specs expected this to be a POST request, but I'm not sure if it has to be.

becky-gilbert:
@okaycj more useful info here about the Unsubscribe POST request: https://docs.sendgrid.com/ui/sending-email/list-unsubscribe#how-does-the-list-unsubscribe-post-header-work

> Inboxes such as Gmail and Yahoo! require all marketing email to contain a List-Unsubscribe-Post header with the value List-Unsubscribe=One-Click. This header communicates that the unsubscribe method is one click via a POST request. Additionally, anti spam software may issue GET requests to links in emails. To prevent accidental unsubscribes, add an additional step so that the GET request to that link does not unsubscribe someone. As an example, you could make the GET request to the link present a landing page to the user containing a link or button to confirm the unsubscribe. A POST request to this link, such as that generated through the one click unsubscribe in the email header, should result in an unsubscribe.

web/templates/web/participant-email-preferences-remove-all.html
```diff
@@ -0,0 +1,19 @@
...
+{% block content %}
+    <div class="row">
+        <div class="col-4">{% include "accounts/_account-navigation.html" with current_page="email-preferences" %}</div>
+        <div class="col-8">
+            <div class="card">
+                <div class="card-body">
+                    You've been unsubscribed from all emails.  If this was a mistake click
+                    <a href="{% url 'web:email-preferences' %}">here</a>
+                    to update your email preferences.
+                </div>
```
becky-gilbert, line 15: What if instead of having a _link_ to the email preferences form (which the left-hand navigation shows is the current page), we actually include the email preferences form here and insert the "You've been unsubscribed..." message above it? That way they get confirmation that they're unsubscribed from all, along with a reminder of the different email categories and a quick way to change them again.
(I guess one problem with doing it this way is that it would make things more complicated if we decide to remove the login requirement.)
Screenshot 2024-02-26 at 5 04 00 PM

becky-gilbert (approved):
LGTM with the caveats about requiring login and GET vs POST. But if we decide to stick with the approach you have here, then this looks great! 😄 
Thank you for moving the email unsubscribe text from individual templates into the base file. As usual, the codebase gets a little better every time you touch it! 💯

web/templates/web/participant-email-preferences-remove-all.html
okaycj: OH!  I like that.

web/views.py
```diff
@@ -299,6 +300,35 @@ def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
...
+        else:
+            email_pref_url = reverse("web:email-preferences")
+            messages.error(
+                request,
+                mark_safe(
+                    f'{username} could not be unsubscribed. Click <a href="{email_pref_url}">here</a> to update your email preferences.'
+                ),
+            )
+
+        return redirect("web:home")
```
becky-gilbert, line 329: Could we maybe present the unsubscribe success/failure message on the user's email preferences page, rather than the home page? Or does the token-based authentication not allow us to do that?
If it is possible, it might be nice to remind the user what the email categories are after they've clicked the unsubscribe-all link, in case they want to change it.

web/views.py
```diff
@@ -299,6 +300,35 @@ def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
         return context
 
 
+class ParticipantEmailUnsubscribeView(generic.View):
+    template_name = "web/unsubscribe.html"
+
+    def get(self, request, username, *args, **kwargs):
+        return render(request, self.template_name, {"username": username})
```
becky-gilbert, line 307: So with this GET request, the user would see this page if they go directly to the `account/<username>` link (with no token)? Or even just `account/`? And then the user sees the 'attempting to unsubscribe' page (web/templates/web/unsubscribe.html)? 
Does this use the session cookie for auth? 
What's the use case for this GET request? Are we handling the case when the user clicks the unsubscribe-all link in the email, but the token has expired, so we're falling back to login for authentication?

okaycj: I'll add a step-by-step rundown of what happens when a user clicks the unsubscribe link to the PR description above.

okaycj: What I could do instead, which might be cleaner, would be to simply redirect to POST in the view.

web/views.py
becky-gilbert: @okaycj NM, I see in your updated PR description that you chose to redirect to home because it doesn't require being logged in. Makes sense!

web/templates/web/participant-email-preferences-remove-all.html
becky-gilbert: I see now that my suggestion here isn't possible because the email preferences page/form requires a log in. Nevermind!

web/views.py
becky-gilbert: Thanks very much for explaining all of this here and in the PR description. All makes sense now!

becky-gilbert (commented):
@okaycj I'm confused about when users would see this template:
`web/templates/web/participant-email-preferences-remove-all.html`
because I don't see it referenced anywhere. Is that a mistake? Or am I (once again) misunderstanding something about how this works?

I do see the new URL pattern `accounts/<username>/<token>` which loads the `ParticipantEmailUnsubscribeView` view and the `web/unsubscribe.html` template.

okaycj:
> I'm confused about when users would see this template:
> `web/templates/web/participant-email-preferences-remove-all.html`
> because I don't see it referenced anywhere. Is that a mistake? Or am I (once again) misunderstanding something about how this works?

They don't.  It should've been removed.  I removed it in commit 2e1b7d9.  Nice looking out.

becky-gilbert:
@okaycj do you know if this is ready for a final review and merge, or is there more left to do? I'm wondering whether @ianchandlercampbell should test some things on staging now (redirection, emails from researchers) or wait until this has been merged and test the one-click unsubscribe options at the same time.

PR #1353, opened 2024-02-28
Change default exit URL to CHS domain

### LINKED ISSUE #1274: Change domain in the default value for a study's Exit URL
The researcher's study edit form (Study Ad and Recruitment page) automatically populates the 'Exit URL' field with 'https://lookit.mit.edu/studies/history'. This is true regardless of what domain the user is on when they create the study (either of the production domains, staging, or localhost).

https://github.com/lookit/lookit-api/blob/90c2e5ad1fe1c5480607a180a0f33d46addd4afa/studies/models.py#L314

Screenshot 2023-09-26 at 2 06 30 PM

If we are assuming/hoping that families will use the Children Helping Science domain, then we should consider changing this default URL to be that same Children Helping Science domain, so that the family isn't directed to a different domain after completing a study, which could be weird/confusing for them. Or a better solution might be to create the study's default Exit URL using the value of `SITE_DOMAIN` from the project settings.

becky-gilbert:
Fixes #1274 

This updates the default value for a study's exit URL from a hard-coded MIT-domain link to one that uses the `BASE_URL` value from the project settings. The `BASE_URL` value has been recently updated in orchestrator as part of the switch to making CHS the primary domain and redirecting from mit.edu (PR here: https://github.com/lookit/lookit-orchestrator/pull/49). That value will change on production the next time we build and deploy the production site. Then, the change in this PR will mean that the default study exit URL will be the participant's study history page on the CHS domain: `https://childrenhelpingscience.com/studies/history/`.

A nice side effect of using the `BASE_URL` instead of hard coding this link is that, when we are testing locally and on the development server, the default exit URL for new studies will use the same site that we're working on (e.g. localhost, babieshelpingscience.com).

Here's what it looks like when I create a new study on my local server:

Screenshot 2024-02-28 at 12 16 38 PM


ianchandlercampbell:
This is wonderful! Go ahead and merge this and then we can test it works correctly on the staging server. Would there be any problems with this changing for existing studies? I assume maybe a minor inconvenience of having to re-log-if they started on Lookit, but that should be eliminated when we have the full domain redirect, right?

studies/models.py
```diff
@@ -281,6 +281,10 @@ def default_study_structure():
     return {"frames": {}, "sequence": []}
 
 
+def default_exit_url():
+    return f"{settings.BASE_URL}/studies/history/"
+
```
okaycj, line 286: Instead of a static URL, could we use "reverse" to generate the path and concatenate it with the base?

okaycj (approved):
Just two quick style issue and we're good to go.  Thank you,

bleonar5 (approved):

Makefile
```diff
@@ -65,3 +65,6 @@ lint: poetry
 
 css: poetry 
 	poetry run ./manage.py custom_bootstrap5
+
+migrations:
+	poetry run ./manage.py makemigrations
```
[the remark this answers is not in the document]
becky-gilbert: True! How about 'makemigrations' (to match the actual Django command)? Or other suggestions?

studies/models.py
becky-gilbert: Yes! Great suggestion, thank you!

becky-gilbert:
> Would there be any problems with this changing for existing studies? I assume maybe a minor inconvenience of having to re-log-if they started on Lookit, but that should be eliminated when we have the full domain redirect, right?

@ianchandlercampbell @msheskin

Great question! Yes you're right that if someone started a study on the MIT domain and then we switch the default exit URL domain to CHS, then they would have to log in again. **BUT** this exit URL update should happen at the same time as the full domain redirect. This means it won't be possible for anyone to complete a study on the MIT domain (they can try to _start_ a study on the MIT domain but they would immediately be redirected to the same page but with the CHS domain, and would probably need to log in again). And since participants will have to start a study on the CHS domain, they will never have to log in when they are sent to the CHS-domain exit URL.

The other thing to keep in mind is that this change to the default exit URL will only affect studies that are created after the update - it will not affect existing studies. So after the domain redirect update, existing studies that still use the lookit.mit.edu domain in their exit URL will do this: 

participant completes study on the CHS domain -> redirect to a MIT-domain exit URL -> redirect to the same page on the CHS domain

This little extra redirection at the end of a study isn't ideal, but we should double check on staging that it doesn't actually cause any problems. The other thing we can do is put out some kind of announcement asking people to change the domain in their study's exit URL when the domain redirection goes live on production.

Let me know if you have questions/concerns/suggestions.

becky-gilbert:
@okaycj I'm going to merge this since the updated version using `reverse` for the study history page seems to be working:

Screenshot 2024-03-01 at 2 52 57 PM

But I will be sure to test on staging.

And I'm happy to change the Makefile target later if you're not a fan of `makemigrations`!

ianchandlercampbell:
Thanks for checking into that @becky-gilbert ! Will definitely be on the lookout for issues related to this when we test in staging, but this sounds reasonable.

PR #1357, opened 2024-03-05
Send all emails from same MIT domain and move lab email to reply-to

### LINKED ISSUE #1356: Fix 'from' headers from labs, and edit custom email template
# TL; DR

The use of lab email addresses as the "From" address in custom messages is causing deliverability problems. The domain used in the 'from' address in all emails we send needs to match the domain that we're actually sending from (currently 'mit.edu', though we could change this in the future).

# Narrative

Custom emails sent by researchers using BC or Berkeley domains for the lab email addresses are already being blocked (not delivered) by email clients (due to strict DMARC policies for those domains). And the new email deliverability requirements for Gmail say that we "must not impersonate From: headers", which means that emails sent from researchers using a Gmail address as their lab account will also be blocked.

We can probably fix this by sending out the custom emails from our mit.edu domain, hopefully using the lab's email as the reply-to address. We still need to test this solution to confirm, but SendGrid actually recommends this as a solution (see [here](https://support.sendgrid.com/hc/en-us/articles/360041356934-Troubleshooting-Email-Delivery-Failures-due-to-DMARC)), so I believe it will work.

From the family's perspective, this change will mean that, instead of seeing study-related emails sent from specific labs, they will see start seeing those emails as sent from lookit-bot@mit.edu, i.e. the same address that sends the automated study announcement emails. If this change will be confusing, we can add some automatic text to the email templates (text and HTML) for emails sent from researchers. Something like "This message was sent on behalf of [LAB/RESEARCHER] regarding [STUDY NAME]. If you'd like to reply, feel free to contact [LAB/RESEARCHER] or reply directly to this email." @ianchandlercampbell I'll let you specify exactly what this text should say and where it should go in the email content. 

Here are our existing templates for custom emails:
- html: https://github.com/lookit/lookit-api/blob/develop/studies/templates/emails/custom_email.html
- text: https://github.com/lookit/lookit-api/blob/develop/studies/templates/emails/custom_email.txt

Once the unsubscribe PR is merged, those templates will include two links at the bottom ("Update your email preferences" and "Unsubscribe from all emails", [see changes here](https://github.com/lookit/lookit-api/pull/1349/files#diff-66d11538a1941e93e85ae74a8d7b7293c08701cce2a571868b938efe0f8a27e8)).

I don't think anything should change from the researcher's perspective - @ianchandlercampbell please let me know if I'm wrong about that.

Relevant links:

https://sendgrid.com/en-us/blog/gmail-yahoo-sender-requirements#5-dont-impersonate-gmail-from-headers

https://sendgrid.com/en-us/blog/gmail-yahoo-sender-requirements#8-for-direct-mail-the-domain-in-the-senders-from-header-must-be-aligned-with-either-the-spf-domain-or-the-dkim-domain

becky-gilbert:
Fixes #1356

This PR does two key things:
1. Removes the optional 'from_email' argument from the `send_email` helper function, because we are now always sending email from the default MIT email address (specifically the `EMAIL_FROM_ADDRESS` in project settings).
2. In cases where we were sending email "from" a lab email address, this PR adds the lab email to the "reply-to" email header.

A few implementation notes:
- The reply-to must be a list or tuple, hence the square brackets around the lab email address in the reply-to argument.
- I didn't need to modify every `send_mail` call in the codebase because the 'from_email' argument was optional, and when omitted, it defaulted to sending from the project's default 'from' address. 
- The `send_mail` function is mocked in `exp/tests/test_contact_views.py`, so there we're just checking that the mocked function wasn't called with the "from_email" argument. 
- There's a new test in `studies/tests.py` called `test_no_custom_from_address` that I think does a better job of testing that any 'from_email' arguments that are passed to `send_mail` will be ignored, and mail will only be sent from the project's 'from' address.

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -677,7 +677,7 @@ class Message(models.Model):
             self.subject,
             to_email_list,
             bcc=bcc_email_list,
-            from_email=lab_email,
+            reply_to=[lab_email],
             **context,
         )
 
--- a/exp/tests/test_contact_views.py
+++ b/exp/tests/test_contact_views.py
@@ -263,8 +263,9 @@ class ContactViewTestCase(TestCase):
             [p.username for p in self.participants[0:3]],
         )
         self.assertEqual(
-            mock_send_mail.call_args.kwargs["from_email"], self.study.lab.contact_email
+            mock_send_mail.call_args.kwargs["reply_to"], [self.study.lab.contact_email]
         )
+        self.assertFalse("from_address" in mock_send_mail.call_args)
 
         # And that appropriate message object created
         self.assertTrue(Message.objects.filter(subject="test email").exists())
@@ -296,8 +297,9 @@ class ContactViewTestCase(TestCase):
         )
         self.assertEqual(mock_send_mail.call_args.kwargs["bcc"], [])
         self.assertEqual(
-            mock_send_mail.call_args.kwargs["from_email"], self.study.lab.contact_email
+            mock_send_mail.call_args.kwargs["reply_to"], [self.study.lab.contact_email]
         )
+        self.assertFalse("from_address" in mock_send_mail.call_args)
 
         # And that appropriate message object created
         self.assertTrue(Message.objects.filter(subject="test email").exists())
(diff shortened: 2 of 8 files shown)


okaycj (approved):
LGTM!

PR #1360, opened 2024-03-07
Updated jspsych urls

### LINKED ISSUE #1325: Study participate/preview URLs for different experiment runners
While working on adding jsPsych study types, we realized that it'll be tricky to use the exact same URL format for different study types. 

Right now, the format for internal (Lookit) studies is:

participate: `studies/STUDY-UUID/CHILD-UUID/`
preview: `studies/STUDY-UUID/CHILD-UUID/preview`

And for jsPsych  (in #1262), it's:

participate: `studies/jspsych/STUDY-UUID/CHILD-UUID/`
preview: `studies/jspsych/STUDY-UUID/CHILD-UUID/preview` 

Questions:
1. Do we want/need all study URLs to have the exact same format (i.e. not contain any information about the experiment runner)? If so, this will require some restructuring of the existing jsPsych work.
2. Or, if it's ok to vary the URL format based on the study type, does it matter whether we're using the study type name in the URL (e.g. '/jspsych/...') vs some other identifier (e.g. '/3/...)?

okaycj:
Closes  #1325

This issue was hoping to have the same URL for all experiment runners.  After some research, redirecting to the reverse proxy prevents us from having a single URL for all runners.  The solution we leaned into was to provide a unique URL for each experiment runner.

Here are the URLs for the two experiment runners we currently support. 
- EFP: studies/`study_uuid`/`child_uuid`/
- jsPsych: studies/j/`study_uuid`/`child_uuid`/

The single "j" in the URL provides us with the uniqueness to tell them apart without necessarily exposing additional data to the experiment participant. Additionally, this leaves the EFP URL as it was, so no legacy URLs need to be updated.

### DIFF
--- a/exp/urls.py
+++ b/exp/urls.py
@@ -14,6 +14,7 @@ Including another URLconf
     2. Import the include() function: from django.conf.urls import url, include
     3. Add a URL to urlpatterns:  url(r'^blog/', include(blog_urls))
 """
+
 from django.urls import path
 from django.views.decorators.csrf import csrf_exempt
 
@@ -256,7 +257,7 @@ urlpatterns = [
         name="preview-proxy",
     ),
     path(
-        "studies/jspsych/<uuid:uuid>/<uuid:child_id>/preview/",
+        "studies/j/<uuid:uuid>/<uuid:child_id>/preview/",
         JsPsychPreviewView.as_view(),
         name="preview-jspsych",
     ),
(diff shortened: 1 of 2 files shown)


becky-gilbert (approved):
Thanks CJ, LGTM! 🚀 
And thanks for all the work you did looking into better solutions for this. It's a shame that Django/revproxy won't let us use the same path structure for both experiment runners 👎 But I agree that it's more important to preserve the EFP links to make sure that existing study links continue to work.

PR #1363, opened 2024-03-12
This is the implementation of "Add email template text to Contact Participants page"

### LINKED ISSUE #1361: Add email template text to Contact Participants page
## TL;DR

On a study's 'Contact Participants' page, we should show researchers the actual text that will emailed to participants, including anything we add to the email.

## Description

As part of the changes to the ways email from researchers to participants are sent, we decided to add automated text about the relevant study/lab and contact info to the researcher message template (see #1359). It would be helpful for researchers if they could see the full email text on the 'Contact Participants' page.

## Implementation options

There are two general approaches to this:

1. We can add some automated text to all researcher-sent emails, and just display that text to the researcher somewhere on the page. Researchers would see this text but would not be able to edit it.
2. We can pre-fill the message text box with the default text, which would encourage researchers to include it, but they would be able to remove it if they want to.

I believe we're going with 1, but we could switch to 2 if there's a good reason to do so.

One thing to consider about implementing this is that it would probably be more difficult to display the _actual_ text that is sent to each individual participant, since information such as the child name will change when a message recipient is added/changed on the page. Also, it's possible for researchers to add multiple recipients to the same message, so we need to account for that case as well. 

Thus the easiest solution might be to just display the template text with variables such as [CHILD_NAME]. We could either use similar static variable names for the study-specific information ([STUDY_NAME, LAB_EMAIL]). Or, since that information should be the same for all messages on a given study's Contact Participants page, it might not be too difficult to replace that information in the template text that is displayed on the page. Something like this:

  > **Each message sent to participants will include the following text above/below your message:**

  > "This is an email for [CHILD_NAME] from Fake Lab for "Awesome fun study!". You may reply to this email to message directly with the researchers at Fake Lab or email them at fakelab@email.com if you have other questions.
  >
  > (YOUR MESSAGE HERE) 
  >
  > <a href="">Update your CHS email preferences</a>
  > <a href="">Unsubscribe from all CHS emails</a>
  > <a href="">Questions or feedback for Children Helping Science?</a>"

## Screenshots

Here is a screenshot of the existing Contact Participants page, along with some suggestions for where we could add info about the additional default text that will be added to each email (or has been added to sent emails). 

Screenshot 2024-03-07 at 1 47 11 PM

Question: How does the 'Previous Messages' table work when a message is sent to multiple recipients? Does it appear as a single row, or one row per recipient?

okaycj:
Closes #1361

### Summary
Add text to emails to give the receiver context on which study they're being contacted about. 

### Implementation

Created a single template that is used in both types of email (text or HTML) and the participant contact view.  Since all three places share the same source, then they'll all have the same text.  

### Images
Text as seen from the participant view:
Screenshot 2024-03-12 at 11 34 14 AM
Emails as seen in the participant email inbox:
Screenshot 2024-03-12 at 11 35 03 AM

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -673,12 +673,15 @@ class Message(models.Model):
         return announcement_message
 
     def send_as_email(self):
+        lab_email = self.related_study.lab.contact_email
+
         context = {
             "base_url": settings.BASE_URL,
             "custom_message": mark_safe(self.body),
+            "lab_name": self.related_study.lab.name,
+            "lab_email": lab_email,
+            "study_name": self.related_study.name,
         }
-
-        lab_email = self.related_study.lab.contact_email
 
         recipient_email_list = list(self.recipients.values_list("username", flat=True))
 
--- a/studies/templates/emails/custom_email.html
+++ b/studies/templates/emails/custom_email.html
@@ -1,5 +1,6 @@
 {% extends "emails/base.html" %}
 {% block content %}
+    {% include "emails/custom_email_header.html" %}
     {% autoescape off %}
         {{ custom_message }}
     {% endautoescape %}
--- a/studies/templates/emails/custom_email.txt
+++ b/studies/templates/emails/custom_email.txt
@@ -1,4 +1,6 @@
 {% extends "emails/base.txt" %}
+
 {% block content %}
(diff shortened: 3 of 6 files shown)


okaycj:
There's been no movement on this issue.  Closing.

PR #1372, opened 2024-03-20
move email footer links to new lines

becky-gilbert:
This PR puts line breaks between the three links at the bottom of participant emails, so they don't appear all on the same line (see screenshot). This issue only affects the HTML version of the base template.

Screenshot 2024-03-19 at 2 52 56 PM


okaycj (approved):
LGTM

PR #1377, opened 2024-04-09
Parameterize and update translations

### LINKED ISSUE #1373: Fix static word order in translated text
# TL;DR

We need to fix instances where our text strings assume a static order between translated text segments and variables.

# Summary

There is at least one text string on the site that restricts the word order of the sentence because it concatenates a portion of the sentence with text that comes from a variable, and doesn't allow the translation file to move the variable elsewhere in the sentence. 

> trans "This study is conducted by " + study contact info

The above text string is not compatible with Hungarian (and possibly other languages) because the study contact info word order is not the last part of the sentence. 

Thanks to @NevenaKlo for reporting this. We should work with MBAH to identify any other text strings that could be causing the same problem (whether in Hungarian or another language) and fix these at the same time.

# Other relevant issues

See https://github.com/lookit/lookit-api/issues/840

becky-gilbert:
Fixes #1373

This PR does the following:
- Updates all translation .po files. These are out-of-date (all last updated in 2021?) and therefore are missing some text translations and include some translations that are no longer used. 
- Parameterizes the contact information in this case: 
  > trans "This study is conducted by " + contact
  
  so that the `contact` text is passed into the translation string as an argument. Now the translation is written as:
  > "This study is conducted by %(contact)s."
  
  This allows the translator to move the 'contact' text to a different part of the sentence, which is a requirement for the Hungarian translation.
- Updates all translation text for the "This study is conducted by" string to include the new 'contact' argument in the translation. In all cases, the 'contact' argument was added to the end of the existing translation (to match the translation prior to parameterization). E.g. `"Este estudio es realizado por"` is now `"Este estudio es realizado por %(contact)s."`
- Adds some comments for translators to help them handle translations that contain arguments.

**Developer workflow notes**

Updating all of the .po files on `develop` and `master` should not cause problems for the site. Translations of the site will be missing some translations regardless of whether or not we update these files - when this happens, the English text is used. And having the most up-to-date versions on our `develop` branch should make it easier for outside contributors and translators to grab the most recent .po files, check for missing translations, and submit their updates via PRs. But I'm open to other suggestions if anyone sees a problem with this workflow.

### DIFF
--- a/web/templates/web/study-detail.html
+++ b/web/templates/web/study-detail.html
@@ -135,7 +135,10 @@
                             {% trans "Your child is older than the recommended age range for this study. You're welcome to try the study anyway, but researchers may not be able to compensate you or use the collected data in their research." %}
                         </p>
                         <p class="text-danger mb-0 mt-3 d-none" id="criteria-not-met">
-                            {% blocktranslate with contact=study.contact_info %} Your child does not meet the eligibility criteria listed for this study. You're welcome to try the study anyway, but researchers may not be able to compensate you or use your data. Please review the “Who can participate” information for this study and contact the study researchers ({{ contact }}) if you feel this is in error. {% endblocktranslate %}
+                            {# Translators: Please include the untranslated "%(contact)s" string to mark where the contact information should be inserted. #}
+                            {% blocktranslate trimmed with contact=study.contact_info %}
+                                Your child does not meet the eligibility criteria listed for this study. You're welcome to try the study anyway, but researchers may not be able to compensate you or use your data. Please review the “Who can participate” information for this study and contact the study researchers ({{ contact }}) if you feel this is in error.
+                            {% endblocktranslate %}
                         </p>
                         {% if preview_mode and object.needs_to_be_built %}
                             {% bootstrap_button build_runner href=build_runner_url button_class=btn_secondary_classes id="participate-button" %}
--- a/web/templates/web/study-detail.html
+++ b/web/templates/web/study-detail.html
@@ -83,7 +83,12 @@
                     {% endif %}
                 </table>
                 <p class="pb-0">
-                    <em class="mt-3 mb-0 text-center text-muted">{% trans "This study is conducted by" %} {{ object.contact_info }}</em>
+                    <em class="mt-3 mb-0 text-center text-muted">
+                        {# Translators: Please include the untranslated "%(contact)s" string to mark where the contact information should be inserted. #}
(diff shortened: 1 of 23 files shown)


bleonar5 (approved):
Reviewed, haven't tested

ianchandlercampbell (approved):
That workflow seems very sensible to me! If you can put a few notes in a Translations section in the handbook as you finish this, that would be helpful!

web/templates/web/study-detail.html
```diff
@@ -83,7 +83,12 @@ <h4>{% trans "Compensation" %}</h4>
                     {% endif %}
                 </table>
                 <p class="pb-0">
-                    <em class="mt-3 mb-0 text-center text-muted">{% trans "This study is conducted by" %} {{ object.contact_info }}</em>
+                    <em class="mt-3 mb-0 text-center text-muted">
+                        {# Translators: Please include the untranslated "%(contact)s" string to mark where the contact information should be inserted. #}
+                        {% blocktranslate trimmed with contact=study.contact_info %}
+                            This study is conducted by {{ contact }}.
+                        {% endblocktranslate %}
```
okaycj, line 90: The note to the translators is nice, but don't they just see the PO files?

okaycj (commented):
💯 LGTM!  Just a quick unnecessary question.

okaycj (approved):

web/templates/web/study-detail.html
becky-gilbert: Yes, they probably won't be looking at comments here. But this is a trick I learned from [this section](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#comments-for-translators-in-templates) in the Django docs. If you have a comment that starts with the "Translators" keyword right before the `trans`/`blocktrans` template tag, it gets added to the corresponding msgid in .po file, like this:
> #. Translators: Please include the untranslated "%(contact)s" string to mark where the contact information should be inserted.
#: web/templates/web/study-detail.html:88
msgid "This study is conducted by %(contact)s."
msgstr "Diese Studie wird durchgeführt von %(contact)s."

(You can also leave [translator comments](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#comments-for-translators) for text strings that are set in the python code rather than a template.)

For the most part I don't think we need to leave translator comments. But I think they are useful when we're incorporating arguments that need to be added with untranslated variable names and specific syntax, because this seems to be a common source of confusion for translators (moreso for EFP).

PR #1379, opened 2024-04-10
Fix nav button translations

becky-gilbert:
# TL;DR

This PR fixes a problem with the navigation bar button labels not being translated. The problem was related to how we were generating the navigation bar buttons, which was causing Django to not pick up the button text as needing translation.

# Description

While testing the translation updates, I noticed that some of the navigation bar button labels weren't being added to the translation (.po) files. A while ago we changed the way that we create the nav buttons in the templates (with a custom `nav_link` template tag) but it turns out that translating the text within that template tag meant that the text wasn't being picked up by Django and added to the .po file. This fixes the problem by translating the text in the template first, before passing the translation as an argument into the `nav_link` tag. I also re-generated all of the .po files so that they include the navigation bar button labels - the updated file versions are included in this PR.

# Screenshots

Before this change, most of the navigation bar button labels were not being picked up for translation. Here is a screenshot from the `develop` branch in Japanese, which uses our most up-to-date translation file:

Screenshot 2024-04-10 at 3 42 25 PM

The text for the untranslated buttons (CHS Home, FAQ, The Scientists, My Account, My Past Studies) does not exist in the .po file (vs existing in the file but not containing a translation value).

After the changes in this PR, the navigation button labels are being picked up for translation and added to the .po files. (For this screenshot, I was able to set some actual Japanese translations using fuzzy/suggested values, and then I just put in some "TEST" placeholder values for the others that were empty.)

Screenshot 2024-04-10 at 3 40 00 PM

# Implementation notes:

I realize this is less elegant than translating the text argument inside the template tag, because now we have to have a bunch of separate lines in the templates to store each translation string. I did try a few different ways of translating the text inside the `nav_link` code but couldn't get that approach to work.

### DIFF
--- a/locale/de/LC_MESSAGES/django.po
+++ b/locale/de/LC_MESSAGES/django.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: MBAH_studiespage\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-04-09 19:34-0400\n"
+"POT-Creation-Date: 2024-04-10 18:01-0400\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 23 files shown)


okaycj (approved):
💯 LGTM!

becky-gilbert:
This PR has picked up the linting errors that were merged in from develop, so I'll wait for those fixes to be merged in, and then I'll update this PR to sync with develop again.

PR #1381, opened 2024-04-11
Support current jsPsych changes 

okaycj:
### Summary

Here's a bit of cleanup that will be required from changes in our jsPysch plugins mono repo.  

Also, there were a few linting errors introduced by merging in MBAH changes.


becky-gilbert (approved):
Looks good!

PR #1392, opened 2024-04-15
Update djlint

okaycj:
### Summary

`djlint` remains one of the worst performing linter/formaters we use.  This PR is another attempt to help with that.


becky-gilbert (approved):
Thanks for handling this!

PR #1391, opened 2024-04-15
Move changes into Production

okaycj:
### Summary

These are the changes on staging that we're going to move to production.

PR #1397, opened 2024-04-17
Add ruff linter

okaycj:
### Summary 

Changes from dependabot weren't caught by testing or existing linters on project.  This PR replaces black, autoflake, and isort and adds ruff as formatter and linter.

### Note

The import issue found in https://github.com/lookit/lookit-api/pull/1391 is resolved by this PR.


api/tests/test_demographics.py
```diff
@@ -96,18 +96,6 @@ def testGetDemographicsInactiveUsers(self):
...
-        self.client.force_authenticate(user=self.participant_without_consented_response)
-        api_response = self.client.get(
-            self.url, content_type="application/vnd.api+json"
-        )
-        self.assertEqual(api_response.status_code, status.HTTP_200_OK)
-        self.assertEqual(
-            api_response.data["results"][0]["additional_comments"],
-            "Additional comments 2",
-        )
-
```
becky-gilbert, line 110: Do we know why the linter removed this whole test?

exp/views/study.py
```diff
@@ -9,12 +9,7 @@
...
 from django.db.models.functions import Lower
 from django.forms.models import BaseModelForm
-from django.http import (
-    HttpRequest,
-    HttpResponse,
-    HttpResponseForbidden,
-    HttpResponseRedirect,
-)
+from django.http import HttpRequest, HttpResponseForbidden, HttpResponseRedirect
 from django.http.response import HttpResponse
```
becky-gilbert, line 13: I think `HttpResponse` was removed here because it's also imported below from `django.http.response`. But it looks to me like it should be imported from `django.http`, not `django.http.response` (see [docs here](https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpResponse)). 
Should we change this to:
```python
from django.http import HttpRequest, HttpResponse, HttpResponseForbidden, HttpResponseRedirect
```
And remove this line:
```python
from django.http.response import HttpResponse
```

studies/models.py
```diff
@@ -26,7 +26,6 @@
 
 from accounts.models import Child, DemographicData, User
 from attachment_helpers import get_download_url
-from project import settings
```
becky-gilbert, line 29: Was this line removed automatically by the linter or by you @okaycj?
This file needs access to `settings`, so I'm guessing this was removed because settings is imported earlier:
```python
from django.conf import settings
```
Do we know which of these imports is correct?

project/sendgrid_backend.py
```diff
@@ -14,16 +15,6 @@
...
-try:
-    from urllib.error import HTTPError  # pragma: no cover
-except ImportError:  # pragma: no cover
-    from urllib2 import HTTPError  # pragma: no cover
-
-try:
-    import rfc822
-except ImportError:
-    import email.utils as rfc822
-
```
becky-gilbert, line 18: @okaycj did you remove this or was it removed by the linter? If it's the latter case - are we ok with this?

becky-gilbert (commented):
Looks great @okaycj, thanks! I've reviewed all of the code changes but not tested.

I left some comments to double check a few of the changes, because I wasn't sure whether they were things that you changed or if they were done automatically by the linter.

api/tests/test_demographics.py
okaycj: It's duplicated!  There's another method with the same name and the same code.

okaycj: Oh!  I removed it.

studies/models.py
okaycj: They're probably both correct, but `django.conf` is correct per docs.  https://docs.djangoproject.com/en/5.0/topics/settings/#using-settings-in-python-code.

project/sendgrid_backend.py
okaycj: I removed it.  I am okay with this as these types of imports are to support varying versions of python.  Especially from python2 to python3.

exp/views/study.py
okaycj: Per the docs, `django.http` is the correct import, but both probably work. https://docs.djangoproject.com/en/5.0/ref/request-response/#usage

okaycj: Fixed in commit https://github.com/lookit/lookit-api/pull/1397/commits/38b12e47771da372ad8ec5f16e5c5a864b924185.

project/sendgrid_backend.py
okaycj: Updated import in commit https://github.com/lookit/lookit-api/pull/1397/commits/e59417ad2913041224a550a720ad6ee83dc3d3bb.

becky-gilbert (approved):
Thanks for answering all of my questions 🙂

PR #1408, opened 2024-05-24
Handle large single video downloads

### LINKED ISSUE #1404: Check/handle large webcam recording files
## Summary

Recently a researcher attempted to download a very large recording file, which the site was not able to handle and caused a short period of downtime. We have addressed this on the experiment runner side by capping the recording duration. But a more comprehensive solution would be to also check the file size in the lookit-api's download views and handle the case when a single video file is too large.

## Implementation notes

Here is the relevant bit of code for [downloading single video files](https://github.com/lookit/lookit-api/blob/be61058ca92d03f83f7d671b4520e42196a819bd/exp/views/responses.py#L654-L667). It doesn't look like we do any file size check here, so this is one possible place where we could address this.

~~The simplest option is just to block the download with an error message saying that the file size is too big, and with an email address to contact.~~ This scenario shouldn't happen often, so it's probably fine for researchers to have to go through the CHS to access large single recording files. This would also alert us to the frequency of this issue.

Another option is to handle this in the same way we do with the ['download all videos' option](https://github.com/lookit/lookit-api/blob/be61058ca92d03f83f7d671b4520e42196a819bd/exp/views/responses.py#L1416-L1427), which uses the [`build_zipfile_of_videos` task](https://github.com/lookit/lookit-api/blob/be61058ca92d03f83f7d671b4520e42196a819bd/studies/tasks.py#L327). So maybe we could re-use this method for single-file downloads that exceed some size limit.

## Implementation update

@okaycj has pointed out that there is a much better solution here, which is to allow users to download the video directly from the S3 link, instead of getting the whole file from S3 and returning it as our own response. The main benefit of this approach is that we don't need to load the file into the site's memory (thus we don't need to do an arbitrary file size check or risk further out of memory errors).

To implement this, I changed the video attachment view so that it redirects to the video download URL. But one problem with this approach is that the file download only works off-the-shelf for same-origin links. Because we're using a different origin, we also need to set the "Content-Disposition" header to "attachment" (see [here](https://macarthur.me/posts/trigger-cross-origin-download/)). This has to be done when we generate the presigned URL, which is in `attachment_helpers.py`. Without that header value, the file will not download and will instead open in the browser (i.e. the same as our 'view' option).

Another option is to switch out the URLs in the video "download" and "view" buttons (`a` tags) inside the templates so that they use the S3 links. That way, instead of first directing users to our `StudyResponseVideoAttachment` view (with a query parameter for 'download' or 'view'), the download/view buttons would use the S3 links directly. However, this option would bypass the `StudyResponseVideoAttachment` view entirely, which means that there would be no permissions checks at the point when the user clicks the download/view buttons.

## Open question

Users need to pass various permissions checks to reach the two response page views that contain the video download/view buttons and to see those buttons on the page. 
1. Do we need the additional permission checks at the point when the user clicks the video download/view button?
2. Does it matter if the pre-signed, time-limited S3 links are available on the page when it is rendered on the client side, vs more obfuscated (but still available to the client due to the redirection)? 

Depending on the answers to these questions, we could potentially get rid of the `StudyResponseVideoAttachment` view entirely and put the S3 links directly into the template (other option mentioned above).

becky-gilbert:
Fixes #1404

# UPDATE

I've re-implemented this with an entirely different solution - see the "**Implementation update**" section in #1404.

## Summary

This PR fixes the problem with attempted downloads of very large single videos causing out-of-memory errors. Instead of downloading the whole file onto the server and then delivering it as our response to the client, we now just redirect the client to the S3 download URL. These pre-signed time-limited URLs are already available to researchers. 

## Implementation notes

To get the video file to download automatically (vs open in a browser tab), we need to set a header value at the point when we request the pre-signed URL, which is in the `get_download_url` function in `attachment_helpers.py`. Previously this function always returned a single URL, so I've added a parameter that allows it to either return a download URL or a view URL. I've also changed the name to `get_url` to reflect this, and therefore had to update the code in a few places where it is used.

Also, we had a video property that stored the download URL value that was retrieved from the `get_download_url` function. So I've changed that to only store the actual download URL, and added a new property that gets the view URL. I've also therefore had to change the code in a few places where we were referencing the video's `download_url` property but where we actually need the `view_url`.

## Testing

This has been tested locally. The following actions need testing on staging/production, on Chrome and Firefox:
- On the 'individual responses' page, click a "view" button to view a video. The video should open in another browser tab.
- On the 'individual responses' page, click a "download" button to download a video. The file should download.
- On the 'videos' (attachments) page, click a "download" button to download a video. The file should download.
- On the 'videos' (attachments) page, click the "Download all videos" button. You should receive an email with the zip containing the videos.
- On the 'videos' (attachments) page, click the "Download all consent videos" button. You should receive an email with the zip containing the consent videos.
- On the studies history page, the videos should appear on the page.

# OLD VERSION - for reference

## Summary

This PR fixes the problem with attempted downloads of very large single videos causing out-of-memory errors. In the video download request, I added a check for the file's content length. If the content length is absent or too large, then it keeps the user on the Individual Responses page and returns a banner error message:

> The file size for {video.filename} could not be verified. Please contact the CHS team on Slack or at childrenhelpingscience<span>@</span>gmail.com.

> The file {video.filename} is too large to download. Please contact the CHS team on Slack or at childrenhelpingscience<span>@</span>gmail.com.

If neither of those conditions are true, then the file is downloaded.

## Questions

1. What should the file size limit be?
2. Instead of just presenting an error message, do we want to try to handle the downloading of a large file with an offline celery process to email a zip to the researcher (similar to how the 'Download all videos' option works)? If so, this will take longer to implement.

## Implementation notes

I first tried to get the content length with a HEAD request (`requests.head(url)`), but AWS doesn't return the content-length in response to a HEAD request. So instead I used `requests.get(url, stream=TRUE)`, which gets the headers for a GET request and doesn't actually download the content until it is specifically requested. But using `stream=TRUE` means that the connection stays open until it is closed, which is why it's now in a `with` statement.

You can test the file size limit locally by reducing the `MAX_FILESIZE_BYTES` value and trying to download a video file that you know is larger. You will probably want to test this with existing response/video data (e.g. downloaded off the staging site) or you can run through a local study using an older Pipe version of the experiment runner, because the newer RecordRTC system does not create new Video objects in a local database (yet).

To test the absence of the content-length info, you can change `with requests.get(download_url, stream=True) as r:` to `with requests.head(download_url) as r:`.

### DIFF
--- a/attachment_helpers.py
+++ b/attachment_helpers.py
@@ -19,7 +19,7 @@ def get_all_study_attachments(study_uuid):
     return bucket.objects.filter(Prefix=f"videoStream_{study_uuid}")
 
 
-def get_download_url(video_key, recording_method_is_pipe):
+def get_url(video_key, recording_method_is_pipe, set_attachment_header):
     """
     Generate a presigned url for the video that expires in 10 minutes.
     """
@@ -32,11 +32,22 @@ def get_download_url(video_key, recording_method_is_pipe):
         bucket = settings.S3_BUCKET_NAME
 
     try:
-        url = S3_CLIENT.generate_presigned_url(
-            "get_object",
-            Params={"Bucket": bucket, "Key": video_key},
-            ExpiresIn=600,
-        )
+        if set_attachment_header:
+            url = S3_CLIENT.generate_presigned_url(
+                "get_object",
+                Params={
+                    "Bucket": bucket,
+                    "Key": video_key,
+                    "ResponseContentDisposition": "attachment",
+                },
+                ExpiresIn=600,
+            )
+        else:
+            url = S3_CLIENT.generate_presigned_url(
+                "get_object",
+                Params={"Bucket": bucket, "Key": video_key},
+                ExpiresIn=600,
+            )
     except ClientError as e:
         logger.warning(f"Video {video_key} not found in bucket. {e}")
         return None
--- a/attachment_helpers.py
+++ b/attachment_helpers.py
@@ -31,23 +31,16 @@ def get_url(video_key, recording_method_is_pipe, set_attachment_header):
         # RecordRTC bucket
         bucket = settings.S3_BUCKET_NAME
 
+    params = {"Bucket": bucket, "Key": video_key}
+    if set_attachment_header:
+        params["ResponseContentDisposition"] = "attachment"
+
     try:
-        if set_attachment_header:
-            url = S3_CLIENT.generate_presigned_url(
-                "get_object",
-                Params={
-                    "Bucket": bucket,
-                    "Key": video_key,
-                    "ResponseContentDisposition": "attachment",
-                },
-                ExpiresIn=600,
-            )
-        else:
-            url = S3_CLIENT.generate_presigned_url(
-                "get_object",
-                Params={"Bucket": bucket, "Key": video_key},
-                ExpiresIn=600,
-            )
+        url = S3_CLIENT.generate_presigned_url(
+            "get_object",
+            Params=params,
+            ExpiresIn=600,
+        )
     except ClientError as e:
         logger.warning(f"Video {video_key} not found in bucket. {e}")
         return None
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -529,14 +529,12 @@ class StudyResponsesList(ResponseDownloadMixin, generic.ListView):
                 if col.id in columns_included_in_summary
             ]
             this_resp_data["videos"] = resp.videos.values("pk", "full_name")
-            download_urls = [v.download_url for v in resp.videos.all()]
-            for i, v in enumerate(this_resp_data["videos"]):
+            for v in this_resp_data["videos"]:
                 v["display_name"] = (
                     v["full_name"]
                     .replace("videoStream_{}_".format(study.uuid), "...")
                     .replace("_{}_".format(resp.uuid), "...")
                 )
-                v["download_url"] = download_urls[i]
             response_data.append(this_resp_data)
         context["response_data"] = response_data
         context["data_options"] = [col for col in RESPONSE_COLUMNS if col.optional]
@@ -654,7 +652,12 @@ class StudyResponseVideoAttachment(
 
     def get(self, request, *args, **kwargs):
         view_url = self.video.view_url
-        return redirect(view_url)
+        download_url = self.video.download_url
+
+        if self.request.GET.get("mode") == "download":
+            return redirect(download_url)
+        else:
+            return redirect(view_url)
 
 
 class StudyResponseSubmitFeedback(StudyLookupMixin, UserPassesTestMixin, View):
@@ -1394,12 +1397,6 @@ class StudyAttachments(CanViewStudyResponsesMixin, generic.ListView):
         context["sort"] = self.request.GET.get("sort", "")
         context["study"] = self.study
 
-        paginated_videos = context["object_list"]
-        download_urls = [v.download_url for v in paginated_videos.all()]
-        for i, v in enumerate(paginated_videos.values()):
-            v["download_url"] = download_urls[i]
-
-        context["videos"] = paginated_videos
         return context
 
     def post(self, request, *args, **kwargs):
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -4,11 +4,9 @@ import zipfile
 from functools import cached_property
 from typing import Dict, KeysView, List, NamedTuple, Set, Text, Union
 
-import requests
 from django.contrib import messages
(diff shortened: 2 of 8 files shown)


becky-gilbert:
@ianchandlercampbell @okaycj this is a draft because I'm still writing tests, but in the meantime it might be useful to get your feedback.

ianchandlercampbell:
@becky-gilbert This seems like good progress! I think the copy for the warning messages makes sense. My question for now is: Do we have limitations whether we can verify file size on RecordRTC? I think we can talk more about this next week as well and I'm happy to do some user-level testing then as well.

becky-gilbert:
> My question for now is: Do we have limitations whether we can verify file size on RecordRTC? I think we can talk more about this next week as well and I'm happy to do some user-level testing then as well.

Thanks Ian, good question. There's no need to do any separate testing with RecordRTC because the videos are stored the same way and download process works the same for Pipe and RecordRTC. I only wrote the note about that because, in order to test this code locally you need to have a video object in your database (which you then try to download), and you can't add new video objects with a local RecordRTC study. So the RecordRTC vs Pipe issue here is just about generating the local data that's needed to test this, rather than any differences that affect the way this download code works. I hope that makes sense. But yes of course we can talk next week. And user testing on the staging site would be helpful!

attachment_helpers.py
```diff
@@ -32,11 +32,22 @@ def get_download_url(video_key, recording_method_is_pipe):
...
+        if set_attachment_header:
+            url = S3_CLIENT.generate_presigned_url(
+                "get_object",
+                Params={
+                    "Bucket": bucket,
+                    "Key": video_key,
+                    "ResponseContentDisposition": "attachment",
+                },
+                ExpiresIn=600,
+            )
```
okaycj, line 44: 600 seconds? Or minutes?

okaycj: Also, it looks like the only difference between the if and else is the params.  It might be cleaner if you pulled out the params dictionary and had only it in the conditional.  This is a code style note, feel free to ignore.

exp/views/responses.py
```diff
@@ -652,19 +649,13 @@ def can_view_this_video(self):
...
-                File.open(io.BytesIO(r.content)),
-                filename=video.filename,
-                as_attachment=True,
-            )
-            return response
-
-        return redirect(download_url)
+            return redirect(download_url)
+        else:
+            return redirect(view_url)
```
okaycj, line 658: Love this!

studies/queries.py
```diff
@@ -145,7 +145,7 @@ def get_responses_with_current_rulings_and_videos(study_id, preview_only):
         ).recording_method_is_pipe
         videos_per_response[video["response_id"]].append(
             {
-                "aws_url": get_download_url(video["full_name"], recording_is_pipe),
+                "aws_url": get_url(video["full_name"], recording_is_pipe, False),
```
okaycj, line 148: The comment preceding this change might need to go.

okaycj (approved):
As always, I am impressed.  Just a few comments.

attachment_helpers.py
becky-gilbert: > 600 seconds? Or minutes?

@okaycj 600 seconds, i.e. 10 minutes. I kept this value the same. Do you think it should change?

studies/queries.py
becky-gilbert: @okaycj you mean I should remove this comment?

> \# Not prefetching videos above because the inability to use values() and
    # prefetch_related in tandem without combinatorial explosion of result set is precluding us
    # from relying on django's join machinery. Instead, we need to manually join here.
    # See: https://code.djangoproject.com/ticket/26565n
    
Does that comment no longer apply? 

I don't think I did anything to change this. This is the query that gets all of the consent videos for the consent manager view, so I just updated the function name and used `False` for the attachment header param, since this view plays the videos in the browser. So it should work the same as before.

becky-gilbert:
@okaycj thanks for the review! 

I responded to your comments and modified the code style for the S3 request here: https://github.com/lookit/lookit-api/pull/1408/commits/93a626e945487806ab57a3f081963ff3c5abcab0 I agree this is much cleaner, and it clarifies the only difference between the two conditions - thanks!!

studies/queries.py
okaycj: I don't really care, If you feel more comfortable leaving it, we can.  It's just that I don't think it applies to the code it's referencing and the Django ticket link doesn't work.

okaycj (approved):
Thanks for making those changes and responding to my comments.

bleonar5 (approved):
Looks great, and all tests listed worked as expected. Great work!

PR #1411, opened 2024-06-05
Sentry: AttributeError  /2fa-login/

okaycj:
# Summary
[Sentry Issue](https://massachusetts-institute-of-technology.sentry.io/issues/5080130387/events/e7fdbc7ab1a94a69acfe53220c32b894/)

This PR adds a check before verifying OTP.  I am not entirely sure why this is happening, but it's simple enough to check if the OTP object not None before the verify.

### DIFF
--- a/accounts/forms.py
+++ b/accounts/forms.py
@@ -91,7 +91,7 @@ class TOTPCheckForm(forms.Form):
     def clean_otp_code(self):
         """Validation check on OTP code."""
         otp_code = self.cleaned_data["otp_code"]
-        if self.otp.verify(otp_code):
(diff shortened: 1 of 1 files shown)


becky-gilbert:
I'm guessing we don't know how to reproduce the problem? I wish we knew why `self.otp` can be `None` to make sure that this is the right way to handle it. 

The check makes sense, but I wonder if this is just going to cause a different downstream Sentry error since `clean_otp_code` will return nothing in this case. Should we throw an error?

becky-gilbert (approved):

PR #1412, opened 2024-06-05
Sentry: Video.DoesNotExist studies.tasks.delete_video_from_cloud

okaycj:
# Summary 

[Sentry Issue](https://massachusetts-institute-of-technology.sentry.io/issues/4457053316/events/4a55e982852e4364b302a2f6b5c44b9c/)

Django will throw `DoesNoExist` when we attempt to `get` something that isn't in the DB.  This PR adds an exists check before attempting to ask S3 to remove it.  

## Open Question

Do we need to handle the state where the video is on S3 but not in our DB?

### DIFF
--- a/studies/tasks.py
+++ b/studies/tasks.py
@@ -483,12 +483,14 @@ def delete_video_from_cloud(task, s3_video_name):
 
     Meant to have a delay of about 7 days.
     """
-    video_in_pipe_bucket = studies.models.Video.objects.get(
-        full_name=s3_video_name
-    ).recording_method_is_pipe
-    if video_in_pipe_bucket:
-        # delete from Pipe bucket
-        S3_RESOURCE.Object(settings.BUCKET_NAME, s3_video_name).delete()
-    else:
-        # delete from RecordRTC bucket
-        S3_RESOURCE.Object(settings.S3_BUCKET_NAME, s3_video_name).delete()
+    video = studies.models.Video.objects.filter(full_name=s3_video_name)
+
+    if video.exists():
+        video_in_pipe_bucket = video[0].recording_method_is_pipe
+
+        if video_in_pipe_bucket:
+            # delete from Pipe bucket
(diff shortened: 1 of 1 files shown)


becky-gilbert:
> Do we need to handle the state where the video is on S3 but not in our DB?

I'm not sure. `delete_video_on_s3` is a pre-delete hook in the Video class, so this `delete_video_from_cloud` task should only run when a Video object is deleted from the DB. In order for a Video object to be deleted, it has to exist in the DB in the first place, right? Or is it possible for a delete request to be made for a video that doesn't exist in the DB, and for that request to get as far as this pre-delete hook without causing a 'does not exist' error? I guess we could try it if we wanted to know for sure...

becky-gilbert:
Would it be hard to add a log for this case? Something like "Error deleting video <FILENAME> from cloud: it was not found in the database." That way we could search for this message in the future, so that we can see if/when it's happening and handle it if necessary.

becky-gilbert:
> Do we need to handle the state where the video is on S3 but not in our DB?

After investigating this a bit more (see here: #1423) I think the answer is yes. We'll need to: 
1. Fix the code so that the S3 deletion hook can successfully delete the video (or at least get the filename) _before_ the video object is deleted from the DB. Even if this PR fixes the Sentry errors, the videos will continue to not be deleted from S3.
2. Fix the backlog of S3 videos that should've been deleted by going through the list of Sentry errors, or compare the S3 objects list to the DB list and remove any from S3 that do not exist in the DB.

becky-gilbert (commented):
Unfortunately this PR will not fix the actual problem. See #1423. I'm inclined to not merge this and just wait to make further changes that will fix this video deletion problem, because as it is now, at least with the Sentry errors we have a record of the video file names that _should_ be deleted from S3 but haven't been. If we merge this PR, then those cases will just fail silently.

becky-gilbert:
Closing this as it was fixed in #1433

PR #1440, opened 2024-07-18
Add Celery task to handle incomplete video uploads

### LINKED ISSUE #1438: Celery job to handle incomplete video uploads in S3
## Summary

The AWS S3 multi-part upload system (used in EFP and jsPsych) can result in incomplete video uploads (e.g. if the complete request fails or the participant leaves the study early - see https://github.com/lookit/ember-lookit-frameplayer/issues/395), and researchers are requesting access to those partial recordings. We need an automated task for completing partial/incomplete video uploads.

## Description

We are trying to prevent video recording uploads from remaining in an incomplete state (see https://github.com/lookit/ember-lookit-frameplayer/issues/395), but that is not always possible, so we still need a way of handling the video uploads that remain incomplete in S3. We discussed simply deleting the incomplete uploads after some duration, but we have been receiving feedback from researchers that videos from incomplete sessions are still helpful (for verifying that a child participated for compensation, and sometimes obtaining useable data, e.g. when the video recording is mostly complete). Because of the frequency of requests we're getting for partial/incomplete video recordings, it makes sense to create a Celery job to automatically try to complete any incomplete video uploads so that they become available to the researcher.

It is not always possible to complete an incomplete upload. Sometimes the upload was created but is empty, corrupted, or missing parts. In these cases, we should probably just delete the partial upload in order to clear our S3 storage space.

## Acceptance criteria

- [x] Decide on a minimum age for the upload creation before attempting to complete it, to ensure that we're not automatically completing the uploads from in-progress sessions. One day (24 hrs) is probably sufficient. 
- [ ] Communicate to researchers that they may need to wait a certain amount of time (e.g. 1 day) to see incomplete videos appear in participants' responses.
- [x] Decide what to do with incomplete uploads that cannot be completed via the Celery task. ~~Delete immediately~~, or allow the AWS rule to delete it after some period (30 days)? ~~Log their presence/deletion somewhere?~~ **Decided to leave them and allow the AWS lifecycle rule to delete after 30 days. Deleting via the lookit-api seems unnecessarily risking, and leaving them for 30 days also serves as a temporary "log" - we can check the in-progress MPU list if a researcher wants to know why an incomplete video hasn't shown up after ~1 day.**

becky-gilbert:
Fixes #1438

## Celery task

This PR adds a new Celery cleanup task to handle incomplete video recording uploads in S3. The task does the following:
1. Grabs all of the incomplete uploads in the RecordRTC S3 bucket.
2. Filters the list to get only the uploads that were created at least 24 hours ago.
3. If the filtered list of videos is not empty, then for each video in the list:
   1. Get the list of parts that were uploaded for that video.
   2. If the list of parts is not empty, then try to complete the upload using those parts.

There will be some cases when it is not possible to complete a video upload, e.g. there are no uploaded parts or some parts are missing/corrupted. In those cases, the video upload will just remain in an incomplete state and grabbed again the next time the task runs. These cases will accumulate over time. Therefore we also need an S3 [lifecycle rule to abort incomplete multi-part uploads after a certain duration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html#lc-expire-mpu), so that the task size does not continue to grow, and to control our S3 storage costs. ~~I would like to put this lifecycle rule in place after the Celery task runs at least once, so that the old incomplete videos can be completed before they're automatically deleted with the age-based rule.~~ (I already ran the Celery task on the production bucket, and then set up a 30-day lifecycle rule for incomplete multi-part uploads).

## Video card help text on Individual Responses page

This PR also adds some text to the Individual Responses page informing researchers that it may take 24 hours for incomplete recordings to become available for download. This text now always appears under the Video card title (rather than just appearing when the video list is empty). The reason for showing this text at all times is because it's now possible for some of the recordings from a session to become available immediately, while others may become available after a delay, and this is likely to be especially confusing for researchers.

I'd appreciate feedback on the help text!

Video card text:

Screenshot 2024-07-23 at 1 25 54 PM

This text is still shown when there are videos available for download:

Screenshot 2024-07-23 at 1 27 02 PM

Full page view:

Screenshot 2024-07-23 at 1 25 34 PM


ianchandlercampbell:
Some small tweak suggestions (feel free to go with whatever you think is best!)

"Video recordings usually become available for download after a few minutes. In some rare cases, it can take up to 24 hours for videos to become available if they were not completed successfully (e.g. because the participant left the study early, internet connection problems). If a video does not show up after 24 hours, it may not be recoverable."

mekline:
Thanks Ian! Final text combining your suggestions:

"Video recordings usually become available for download after a few minutes, but may take as long as 24 hours to appear. 

Note: In rare cases, video uploads can be damaged or lost, e.g. because the participant left the study early, there was an internet connection problem, etc. We automatically attempt to restore any incomplete video files that reach the servers. If session video files do not appear here after 24 hours, they unfortunately can't be recovered."

(@becky-gilbert still accurate?)

mekline (approved):
(see comment!)

becky-gilbert:
Got it, thanks @mekline and @ianchandlercampbell!

becky-gilbert:
Updated screenshots with the modified video help text:

Screenshot 2024-07-26 at 11 37 34 AM

Screenshot 2024-07-26 at 11 40 05 AM

bleonar5 (approved):

okaycj (approved):
This is good.  Thanks.

becky-gilbert:
@okaycj thanks for your questions about when/how the task is actually scheduled! I thought that it was automatically scheduled as part of the 'cleanup' queue, but that's not the case - each task and associated cron schedule need to go in the database in the `django_celery_beat_periodictask` and `django_celery_beat_crontabschedule` tables. 

I've added a migration file that adds the task and schedules it for daily at 5am. (See the screenshots for our other jobs and schedules). I've tested the forward and reverse migration functions several times and am happy that they're working. 

Tables before migration:

Screenshot 2024-08-01 at 1 49 52 PM

Screenshot 2024-08-01 at 12 56 47 PM

Tables after migration:
(I assume the non-consecutive ID for the new cron entry is because I migrated and reverted the migration multiple times to test)

Screenshot 2024-08-01 at 1 47 22 PM

Screenshot 2024-08-01 at 1 47 40 PM

okaycj (approved):
Yup.  Very good.

PR #1444, opened 2024-07-29
Update child participation eligibility

okaycj:
# Summary

Antidotally, the participation eligibility function was running slowly.  This PR combines some queries within this function so that there are fewer calls to the database.

### DIFF
--- a/accounts/queries.py
+++ b/accounts/queries.py
@@ -144,7 +144,7 @@ def get_child_participation_eligibility(child, study) -> bool:
         )
         must_not = must_not_have_count == 0
 
-    return must_have and must_not
+    return must_not
 
 
 def _child_in_age_range_for_study(child, study):
--- a/accounts/queries.py
+++ b/accounts/queries.py
@@ -114,39 +114,37 @@ def get_child_participation_eligibility(child, study) -> bool:
     Returns:
         bool: Return true if child is eligible based on their prior study participation
     """
-    # Need to import StudyType here due to circular import problems
-    from studies.models import StudyType
 
-    must_have_participated = True
-    must_not_have_participated = True
+    ember_frame_player_id = 1
+    must_have = True
+    must_not = True
 
     # for both must have and must not have participated, ignore responses from internal studies that are empty
-    if study.must_have_participated.exists():
-        must_have_participated_all = [
-            i
-            for i in study.must_have_participated.all()
-            if child.responses.filter(study=i)
-            .exclude(study__study_type=StudyType.get_ember_frame_player(), sequence=[])
-            .exists()
-        ]
 
-        # only true if response exists for each of the studies in the list
-        must_have_participated = set(must_have_participated_all) == set(
-            study.must_have_participated.all()
+    if study.must_have_participated.exists():
+        must_have_count = (
+            child.responses.filter(study__in=study.must_have_participated.all())
+            .exclude(study__study_type_id=ember_frame_player_id, sequence=[])
+            .distinct()
+            .values_list("study")
+            .count()
         )
+        must_have = must_have_count == study.must_have_participated.count()
+
+        if not must_have:
+            return False
 
     if study.must_not_have_participated.exists():
-        must_not_have_participated_all = [
-            i
-            for i in study.must_not_have_participated.all()
-            if child.responses.filter(study=i)
-            .exclude(study__study_type=StudyType.get_ember_frame_player(), sequence=[])
-            .exists()
-        ]
-        # only true if response does NOT exist for any one (or more) of the studies in the list
-        must_not_have_participated = len(must_not_have_participated_all) == 0
+        must_not_have_count = (
+            child.responses.filter(study__in=study.must_not_have_participated.all())
+            .exclude(study__study_type_id=ember_frame_player_id, sequence=[])
+            .distinct()
+            .values_list("study")
+            .count()
+        )
+        must_not = must_not_have_count == 0
 
-    return must_have_participated and must_not_have_participated
+    return must_have and must_not
 
 
 def _child_in_age_range_for_study(child, study):


accounts/queries.py
```diff
@@ -114,39 +114,37 @@ def get_child_participation_eligibility(child, study) -> bool:
...
-
-        # only true if response exists for each of the studies in the list
-        must_have_participated = set(must_have_participated_all) == set(
-            study.must_have_participated.all()
+        must_have_count = (
+            child.responses.filter(study__in=study.must_have_participated.all())
+            .exclude(study__study_type_id=ember_frame_player_id, sequence=[])
+            .distinct()
+            .values_list("study")
+            .count()
```
becky-gilbert, line 130: Are we totally sure that this works? It looks to me like it is doing this:
1) for this child, get all associated responses for all studies in the 'must have participated' list
2) exclude any responses that are from EFP studies and have an empty sequence
3) call `distinct` to remove any duplicate response rows
4) get the list of study values in this set of distinct response rows
5) count the number of studies in the list

I could be wrong, but it seems to me that the order of `distinct` and `values_list` calls should be reversed, so that we're first grabbing just the study IDs from the filtered responses, and then looking for the distinct values in that set. If `distinct` is called on the queryset with the response rows, then it will return a queryset that can contain repeated responses for the same study (as long as the response rows differ in some other way). Then we're getting the study values from that queryset, which means the list it returns can contain duplicate entries for the same study. Right?

Happy to help test if necessary, I'm just not totally sure how to set up a test case.

mekline: Good catch, I think @becky-gilbert is right about this unless I miss something about how .distinct() is calculated!

accounts/queries.py
```diff
@@ -114,39 +114,37 @@ def get_child_participation_eligibility(child, study) -> bool:
...
         )
+        must_have = must_have_count == study.must_have_participated.count()
+
```
mekline, line 133: Along the same lines, I'm wondering just in case if there are other reasons I'm not thinking of why the number of studies might not match counts with the number of exclusions?  How much slower would it be to sort and ask if the two lists are identical, `study.must_have_participated.count()`  and the list generated in the 2nd to last line of that query?

accounts/queries.py
```diff
@@ -114,39 +114,37 @@ def get_child_participation_eligibility(child, study) -> bool:
...
-            .exclude(study__study_type=StudyType.get_ember_frame_player(), sequence=[])
-            .exists()
-        ]
-        # only true if response does NOT exist for any one (or more) of the studies in the list
-        must_not_have_participated = len(must_not_have_participated_all) == 0
-
-    return must_have_participated and must_not_have_participated
+        must_not_have_count = (
+            child.responses.filter(study__in=study.must_not_have_participated.all())
+            .exclude(study__study_type_id=ember_frame_player_id, sequence=[])
```
mekline, line 140: Now that I'm thinking about query speed, is this `exclude()` part slow? Can you fetch the child responses 1x, exclude empty sequences 1x, and *then* calculate the must have/must have nots? My certainty on this is extremely low, ignore if wrong!

accounts/queries.py
becky-gilbert: Thank you for explaining this offline @okaycj! I now see that the order of statements here is misleading/confusing, and the SQL query that's generated from this _IS_ selecting the studies first, before calling distinct. 

(For the record, this is totally confusing and counterintuitive to me! But I guess this is part of the behind-the-scenes ORM magic. Thanks for showing me how to use the queryset's `query` attribute to see the actual SQL query that the code generates, which I'm sure will help me figure this out in the future 👍 )

becky-gilbert (approved):

accounts/queries.py
okaycj: The functions provided by the Django ORM construct the SQL query and run that query when it comes to a  QuerySet function such as `count()`.  I like to think of each function adjusting the SQL query and running it when we need to.  The first two functions add to the "WHERE" statement which filters the output.  The `distinct()` and `values_list()` update the `SELECT` statement limiting the number of columns returned.  The necessary `JOINS` are generated from the double underscores in the filter/exclude functions.  This is how they convert the imperative Python to the declarative SQL.

accounts/queries.py
okaycj: We can, it's just unnecessary.  

The responses for this query will be from a study in the `must_have_participated` list.  This is from the filter `study__in=study.must_have_participated.all()`.  

Here's SQL generated by `child.responses.filter(study__in=study.must_have_participated.all())`:

```SQL
select
	"studies_response"."id",
	"studies_response"."uuid",
	"studies_response"."study_id",
	"studies_response"."completed",
	"studies_response"."completed_consent_frame",
	"studies_response"."survey_consent",
	"studies_response"."exp_data",
	"studies_response"."conditions",
	"studies_response"."sequence",
	"studies_response"."date_modified",
	"studies_response"."date_created",
	"studies_response"."global_event_timings",
	"studies_response"."child_id",
	"studies_response"."is_preview",
	"studies_response"."demographic_snapshot_id",
	"studies_response"."study_type_id",
	"studies_response"."recording_method",
	"studies_response"."eligibility"
from
	"studies_response"
where
	("studies_response"."child_id" = 10176
		and "studies_response"."study_id" in (
		select
			U0."id"
		from
			"studies_study" U0
		inner join "studies_study_must_have_participated" U1 on
			(U0."id" = U1."to_study_id")
		where
			U1."from_study_id" = 3129))
```

accounts/queries.py
okaycj: I don't think it'll make the query slower, because it is only run after the `count()` function and more restrictive filters usually speed up the query's response.  More restrictions equals less data to return.

becky-gilbert:
In case anyone else sees the Sentry error above - I _think_ this was caused by doing a production release while the `send_announcement_emails` task was still running, hence the "closed the connection unexpectedly" error. So I think this is nothing to worry about. We will know for sure tomorrow after the announcement emails task has had the chance to run again.

PR #1450, opened 2024-08-07
Further Japanese translations

rhodricusack:
(1) Updates at lookit had left de-internationalised text. Specifically, replacing links with templatetags like this does not add the string "Log in to participate" to the .po translation files:
`<div class="my-3">{% nav_login request text="Log in to participate" button=True %}</div>`

from https://github.com/manybabies/mbah-lookit-api/blob/44fdfa06215dc10cfe72487ee0f2c7487fd04d22/web/templates/web/study-detail.html#L98

I have therefore change two links to this format. Would you please be able to use it in future?

```
  {% trans "Log in to participate" as log_in_to_participate %} 
  <div class="my-3">{% nav_login request text=log_in_to_participate button=True %}</div>
```

(2) The format for .po files that our translation software (POEditor) exported had no line breaks, but django's makemessages breaks up long blocks onto multiple lines. This would change all of the translations to the extend that git merge would fail to elegantly map, and would just report thousands of lines of translation as changed. We have now changed our procedures to use a text editor instead of POEditor. So, the .po file has been formatted in a way output by makemessages.

(3) A few new strings have been translated.


shotsuji:
@becky-gilbert we have updated translations, you can go ahead with the pull request.

PR #1454, opened 2024-08-28
Jspsych response helpers

okaycj:
# Summary

For the experiment data coming out of jsPsych, we needed to update how the data was parsed.  This included adding a couple of conditionals and a shim until we can complete other issues.  

 - Added parsing logic to exit frame property function
 - For the time being, we'll not get consented responses.  This will be reverted when the consent view has been updated.
 - Handle jsPsych videos on save

# Notes 
This isn't the full solution to the researcher's data view.  Instead, this was a good place to check in before doing more work.

### DIFF
--- a/exp/tests/test_response_postsave.py
+++ b/exp/tests/test_response_postsave.py
@@ -215,6 +215,7 @@ class ResponseSaveHandlingTestCase(TestCase):
 
 class JspsychResponseSaveHandlingTestCase(TestCase):
     def setUp(self):
+        settings.CELERY_TASK_ALWAYS_EAGER = True
         jspsych = StudyType.get_jspsych()
         user = G(User)
         child = G(Child, birthday=datetime.date(1911, 11, 11), user=user)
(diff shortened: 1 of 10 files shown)


exp/tests/test_response_postsave.py
```diff
@@ -208,3 +211,52 @@ def test_responses_per_study_type(self):
...
+        G(
+            Video,
+            response=self.response,
+            study=study,
+            s3_timestamp=timezone.now(),
+            frame_id=frame_id,
+        )
+
+    def test_exit_withdraw_videos(self):
+        settings.CELERY_TASK_ALWAYS_EAGER = True
```
becky-gilbert, line 239: Does this need to be set back to `False` at the end of the test or in a `tearDown` function?

becky-gilbert (approved):
LGTM! Just one question.

exp/tests/test_response_postsave.py
okaycj: It didn't change the outcome of the tests, but I did it anyway.

becky-gilbert: Thanks!

PR #1458, opened 2024-09-18
Fix issue when attempting to review jspsych study

okaycj:
# Summary

A quick PR to fix an issue found with reviewing JsPsych studies.  

# Explanation

When we start any experiment, we also create a response.  If you make a response and don't set the study type, it will default to E.F.P. even though you have provided a study.  When creating JsPsych studies, I created the response and stated the study type but forgot to update when a researcher would review a study.  These are two different views.  This PR places the creation of a response for a JsPsych study into a single function.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,9 @@ collectstatic:
 	docker compose run --rm web poetry run ./manage.py collectstatic --clear --noinput
 
 poetry:
-	poetry check && poetry install --sync --no-root
+	poetry self update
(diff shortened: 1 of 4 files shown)


becky-gilbert (approved):
Thanks for the explanation! I think I get it now. It sounds like the critical change was adding this `study_type=study.study_type` to the jsPsych preview response. 
Looks great, thanks for refactoring!

PR #1461, opened 2024-10-04
Update ja translations

becky-gilbert:
This PR contains the changes to Japanese translations that were made by @rhodricusack and @shotsuji: #1450 
It also contains the following changes:

## Fix for compile messages error

I moved their changes into this branch so that I could fix a `compilemessages` error that was caused by two "\n" strings that was present in `msgid`s but absent in the corresponding `msgstr`s:

Screenshot 2024-10-04 at 2 42 52 PM

After this change, `compilemessages` is running without errors:

Screenshot 2024-10-04 at 3 21 49 PM

## Update all other .po files

I also updated the other .po files so that anyone else who wants to update our other translations will have the most up-to-date translation files to use. The changes to these other files should _not_ affect the actual translated text on the site for these languages - it just affects the line number mappings and `msgid` values that might've changed since these .po files were last created, making it easier for translators to find and update incorrect/missing translations.

## Check for other non-translated text

As @rhodricusack mentions in #1450, the "Log in to participate" and "Create a new account" buttons on the study details pages for a non-logged-in user were not being translated, because we'd used other template tags to create the buttons and passed hard-coded text in as a parameter. I searched the codebase for other similar instances and couldn't find any. But if anyone else spots non-translated text that's not being added as `msgid` values in the .po files, we should know about it so that we can fix it!

## Screenshots

The "Log in to participate" and "Create a new account" buttons on the study details page are now being translated to Japanese:

Screenshot 2024-10-04 at 3 09 20 PM


okaycj (approved):
💯

PR #1465, opened 2024-10-15
jsPsych video/data access

becky-gilbert:
This PR adds the necessary changes for accessing the response data and videos from jsPsych studies.

- Adds the jsPsych AWS S3 environment variables that lookit-jspsych needs for uploading videos during jsPsych studies.
- Adds conditionals for accessing jsPsych study video files from the jsPsych bucket. This was done by adding an additional parameter to `get_url` and `delete_video_from_cloud`.
- Fixes problem with the jsPsych response object `recording_method` being "pipe" because it was the default value for that field. I've removed the default value and allowed NULL.
- Updates the `recording_method_is_pipe` and `recording_method_is_recordrtc` properties to handle the possible NULL `recording_method` value.
- Updates the experimenter 'responses' pages so that they show videos and experiment data for jsPsych study responses.
- Updates the queries for the study lists and study history pages so that jsPsych studies are also shown. They are categorized the same way as EFP studies (i.e. as "Studies happening here on the Lookit platform" and "Studies happening right now" on the study list page, and as "Lookit studies" on the study history page).
- Updates tests for the jsPsych study type.

## Screenshots:

jsPsych studies are shown on the Studies page (as Lookit/unscheduled):

Screenshot 2024-10-15 at 4 22 42 PM

jsPsych studies are shown on the study history page (as Lookit) and associated videos use the correct AWS bucket in the URL:

Screenshot 2024-10-15 at 4 25 02 PM

jsPsych response pages show the videos associated with responses, and links use the correct AWS bucket:

Screenshot 2024-10-15 at 4 26 24 PM

When a new Response object is created for a jsPsych study, the `recording_method` is NULL.

Screenshot 2024-10-15 at 4 27 34 PM


studies/tasks.py
```diff
@@ -479,17 +479,24 @@ def build_framedata_dict(filename, study_uuid, requesting_user_uuid):
...
-        # delete from Pipe bucket
-        S3_RESOURCE.Object(settings.BUCKET_NAME, s3_video_name).delete()
+    if study_type_is_jspsych:
+        # delete from lookit-jspsych bucket
+        S3_RESOURCE.Object(settings.JSPSYCH_S3_BUCKET, s3_video_name).delete()
     else:
-        # delete from RecordRTC bucket
-        S3_RESOURCE.Object(settings.S3_BUCKET_NAME, s3_video_name).delete()
+        # EFP
+        if recording_method_is_pipe:
```
okaycj, line 494: This is the most unnecessary comment, but you could use an `elif` here instead.

okaycj (approved):
Nice!

studies/tasks.py
becky-gilbert: Not an unnecessary comment - thanks for noticing and flagging this!

I wrote this with `elif` the first time, then re-wrote it this way to make the logic clearer. The properties `recording_method_is_pipe` and `recording_method_is_recordrtc` both return `False` (not null/undefined) for jsPsych studies, so I was (overly) worried that this could lead to bugs if we ever write the conditionals in a different way - e.g. trying to use `not recording_method_is_pipe/recordrtc` to differentiate EFP types, which would also catch jsPsych studies. But I'm now realizing that that would just be a super weird way of doing things and it's not something I should worry about 🤦‍♀️ Anyway, I'm going to re-write it.

PR #1472, opened 2024-10-28
AWS video upload credentials for jsPsych studies

becky-gilbert:
This PR generates temporary AWS credentials and passes them into the jsPsych study template, where they can be retrieved and used by the lookit-jspsych `data` package. We're using the maximum duration for these credentials, which is 36 hours.

**Screenshots**

The temporary AWS credentials have to be requested each time a user tries to start a jsPsych study. If for some reason there's an error obtaining those credentials, the user is redirected back to the study/preview details page, and they'll see an error message.

Screenshot 2024-10-28 at 11 51 13 AM

### DIFF
--- a/exp/views/study.py
+++ b/exp/views/study.py
@@ -918,17 +918,26 @@ class JsPsychPreviewView(
 
     test_func = can_preview
 
-    def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
-        context = super().get_context_data(**kwargs)
-        response = get_jspsych_response(context, is_preview=True)
+    def get(self, request, *args, **kwargs):
+        # Need to check for AWS variables here instead of get_context_data, so that we can redirect if there's an error.
         aws_vars = get_jspsych_aws_values()
         if aws_vars is None:
             messages.error(
                 self.request,
                 "There was an error starting this study. Please contact lookit@mit.edu.",
             )
+            return redirect(
+                reverse("exp:preview-detail", kwargs={"uuid": self.get_object().uuid})
+            )
+
+        self.aws_vars = aws_vars
+        return super().get(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
+        context = super().get_context_data(**kwargs)
+        response = get_jspsych_response(context, is_preview=True)
         context.update(response=response)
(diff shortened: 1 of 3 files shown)


web/views.py
```diff
@@ -850,8 +896,14 @@ def user_can_participate(self):
     def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
         response = get_jspsych_response(context)
+        aws_vars = get_jspsych_aws_values()
+        if aws_vars is None:
+            messages.error(
+                self.request,
+                "There was an error starting this study. Please contact lookit@mit.edu.",
+            )
```
becky-gilbert, line 904: @okaycj for this error handling, I think I need to make changes to keep the user on the study detail page, rather redirecting to the jsPsych experiment. Do you know whether I need to add a `render` or `redirect` in this case?

becky-gilbert: I think `redirect` is the right approach here but I can't redirect from within `get_context_data`, so my solution was to override the GET method and get the AWS credentials there. That way if there's an error I can still redirect back to the study details (with an error message), otherwise pass the credentials into the context.

exp/views/study.py
```diff
@@ -913,11 +918,26 @@ def can_preview(self):
 
     test_func = can_preview
 
+    def get(self, request, *args, **kwargs):
+        # Need to check for AWS variables here instead of get_context_data, so that we can redirect if there's an error.
+        aws_vars = get_jspsych_aws_values()
+        if aws_vars is None:
+            messages.error(
```
okaycj, line 925: You can just save the `aws_vars` to `self.aws_vars`.  It is probably not great to allocate memory for this variable in two spaces.

web/views.py
okaycj: Overriding GET will get the job done.  You could also try [get_success_url()](https://docs.djangoproject.com/en/5.1/ref/class-based-views/mixins-editing/#django.views.generic.edit.FormMixin.get_success_url).

okaycj (approved):
💯

exp/views/study.py
becky-gilbert: Doh! Yep. Thanks for that - fixed!

PR #1479, opened 2024-11-05
Update lookit-jspsych `surveys` package version

becky-gilbert:
This PR just bumps the lookit-jspsych `surveys` package version, which contains a bug fix.


okaycj (approved):
Looks good.

PR #1481, opened 2024-11-07
Update jspsych deps

okaycj:
# Summary

This PR updates our jsPsych plugins.


Makefile
```diff
@@ -14,7 +14,7 @@ migrate:
 superuser:
 	docker compose run --rm web poetry run ./manage.py createsuperuser
 
-site:
+site: migrate
```
becky-gilbert, line 17: Not sure what this is doing. Just checking that it was intentional 😄

becky-gilbert (approved):
Looks great, thanks!

Makefile
okaycj: This was.  I am always trying to improve local environment.

PR #1482, opened 2024-11-12
Response recording method

okaycj:
# Summary

It was found that some videos wouldn't load in the consent manager view.  This was due to misidentifying the recording method (e.g., pipe vs. "recordrtc"), as there are different S3 buckets for each type.  

# Solution
We changed the default value for the recording methods field from "pipe" to NULL. Researchers using EFP, which still uses the pipe service will not be able to see recent videos.  To fix this issue, I've put together a migration that sets the default value for this field back to "pipe."  This should resolve the problem from now on.  Additionally, the migration will check if the `recording_method` field is null and, if so, attempt to detect what the recording method could be and set the field.


becky-gilbert (approved):
Looks good to me! Thanks!

studies/migrations/0100_alter_response_recording_method.py
```diff
@@ -0,0 +1,44 @@
...
+JSPSYCH = 3
+
+
+def populate_recording_method(apps, schema_editor):
+    responses = apps.get_model("studies", "Response").objects.filter(
+        recording_method=None
+    )
+    for response in responses:
+        if response.study_type.id == EXTERNAL:
+            response.recording_method = None
```
becky-gilbert, line 16: I'm fine with using None/NULL for responses from external studies here. But I just realized that the external responses will go back to being "pipe" after this migration, because of the default. So this will produce an inconsistency. Does that matter? Should we explicitly set the recording method to NULL for external study responses?

okaycj: Yes, we should.  I make that change.

becky-gilbert (changes requested):
Ok, I like it. So just to be clear about what we're doing:
- A response's `recording_method` can be null, and "pipe" is the default
- Responses from EFP Pipe studies will now be "pipe" again, because that field will be unset and it's the default
- Responses from EFP RecordRTC studies will still be "recordrtc", because it's set explicitly by EFP
- Responses from external studies will have a `recording_method` of null, because we're setting that explicitly

I think the only thing we're missing is setting the `recording_method` value explicitly in jsPsych study responses, otherwise it will default to "pipe". Right? I think we just need to set it here:

```python
def get_jspsych_response(context, is_preview=False):
    study = context["study"]
    child_uuid = context["view"].kwargs["child_id"]
    child = Child.objects.get(uuid=child_uuid)
    demo = child.user.latest_demographics
    return Response.objects.create(
        study=study,
        child=child,
        demographic_snapshot=demo,
        exp_data=[],
        is_preview=is_preview,
        study_type=study.study_type,
    )
```

okaycj:
@becky-gilbert 
> I think the only thing we're missing is setting the `recording_method` value explicitly in jsPsych study responses, otherwise it will default to "pipe". Right?

I agree that we should set the `recording_method` in this function, but where did we set this value during development?

becky-gilbert:
It was unset for jsPsych studies because the default was null. But now that the default is "pipe" it needs to be set explicitly.

web/views.py
```diff
@@ -110,6 +111,7 @@ def get_jspsych_response(context, is_preview=False):
         exp_data=[],
         is_preview=is_preview,
         study_type=study.study_type,
+        recording_method="jspsych",
```
becky-gilbert, line 114: @okaycj so far we've been using `null`/`None` for the jsPsych recording method. Just double checking that we do want to intentionally switch to "jspsych". (FWIW, I don't think it matters!)

Maybe you've already covered this in the tests, but we just need to make sure that whatever logic we're using around `recording_method` still makes sense with jsPsych values. For instance, if we use "jspsych" as a `recording_method` value, then maybe we should add a new `recording_method_is_jspsych` property. That way we could fix this logic that checks the study type first, then the recording method:

```python
def get_url(
    video_key, recording_method_is_pipe, study_type_is_jspsych, set_attachment_header
):
    """
    Generate a presigned url for the video that expires in 10 minutes.
    """
    url = None
    if study_type_is_jspsych:
        # jsPsych bucket
        bucket = settings.JSPSYCH_S3_BUCKET
    else:
        # EFP
        if recording_method_is_pipe:
            # Pipe bucket
            bucket = settings.BUCKET_NAME
        else:
            # RecordRTC bucket
            bucket = settings.S3_BUCKET_NAME
```

I'm open to whatever make sense to you here.

becky-gilbert (approved):

PR #1483, opened 2024-11-13
Move S3 bucket fix into production

okaycj:
# Summary

This PR contains the fix for the recent issue with selecting the correct bucket when attempting to watch videos.


becky-gilbert (approved):
Looks good to me! 
FYI I've tested this on staging by running through the following study types:
- EFP Pipe version
- EFP RecordRTC version
- jsPsych
- External
I can confirm that, for all internal studies, the videos are saving to and being retrieved from the correct buckets. I checked this in the Consent Manager and Response views.
I also checked the `recording_method` values in the database and they are all as expected.

PR #1485, opened 2024-11-19
Tests

okaycj:
# Summary
This PR contains the test cases for the fix in  #1482.

### DIFF
--- a/studies/test_recording_method.py
+++ b/studies/test_recording_method.py
@@ -1,15 +1,15 @@
 import datetime
 import uuid
-from unittest import skip
-from urllib.parse import parse_qs, urlparse
+from datetime import date
+from unittest.mock import patch
 
 from django.conf import settings
 from django.core.files.uploadedfile import SimpleUploadedFile
-from django.test import Client, TestCase
+from django.test import Client, TestCase, override_settings
 from django.urls import reverse
 
 from accounts.models import Child, User
-from attachment_helpers import get_url
+from attachment_helpers import S3_CLIENT, get_url
 from studies.models import Lab, Response, Study, StudyType
 from web.views import create_external_response, get_jspsych_response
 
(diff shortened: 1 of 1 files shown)


studies/test_recording_method.py
```diff
@@ -135,44 +151,63 @@ def test_external_researcher_preview(self):
...
-@skip("Need AWS credentials.")
 class RecordingMethodEFPTestCase(TestCase):
     def test_efp(self):
         study_type = StudyType.get_ember_frame_player()
-        user, study, child = get_user(study_type)
+        _, study, child = get_user(study_type)
         response = Response.objects.create(
             study_type=study_type, child=child, study=study
         )
         self.assertEqual(response.recording_method, "pipe")
```
becky-gilbert, line 162: For the EFP test case, I was going to suggest that we add tests for the web/external view and researcher preview view (like you've done with external and jsPsych study types. But then I realized that the EFP recording method comes from the response POST, so I assume that's why we don't need a test for it here?

becky-gilbert (approved):
Fantastic, thank you!

studies/test_recording_method.py
```diff
@@ -1,15 +1,15 @@
 import datetime
 import uuid
-from unittest import skip
-from urllib.parse import parse_qs, urlparse
+from datetime import date
```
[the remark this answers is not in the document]
okaycj: No reason.  I copied it from Python documentation. They're equal.

studies/test_recording_method.py
okaycj: "Need" is a strong word. I did this to be complete, and the only way I could confirm that EFP was setting the recording method correctly.  The inference is that if EFP uses the API to generate the Response and  EFP relies on the default of the `recording_method` field, then this makes sure that the default setting remains "pipe."

okaycj: Testing for how EFP creates and uses the response should be in the EFP test cases.

PR #1491, opened 2025-01-06
Disable zip video download

okaycj:
# Summary
This PR will disable the ability to zip an experiment's video files.  This is in response to the issues with our site going down randomly.  Once we've redesigned the zipping videos celery task, these buttons should be re-enabled.


becky-gilbert (approved):
Tested locally and works, thanks!

PR #1496, opened 2025-01-23
Key error in incomplete videos task

### LINKED ISSUE #1495: Cleanup incomplete videos task causes errors
## Summary

The Celery task that cleans up incomplete video files on S3 is producing a set of errors on the builder pod every time it runs. The builder pod is currently causing memory-related problems that might be taking down the site, so handling any errors on this pod is helpful for narrowing down the problem.

## Description

The cleanup. incomplete videos task is causing these errors:

- ClientError: An error occurred (AccessDenied) when calling the ListMultipartUploads operation. 
   - **Update: this has been fixed via AWS policy settings.**
- KeyError: 'Uploads' (in `get_all_incomplete_video_files`). This will need to be addressed in the task code.
   - PR: #1496

becky-gilbert:
Fixes #1495

This PR fixes a key error in the "cleanup incomplete videos" task, which was caused by the "Uploads" key not existing in the AWS API response. 

First, this changes the conditional statement to prevent the error from being thrown when "Uploads" does not exist in the response. 

Second, this adds a log statement with the response when it doesn't contain the "Uploads" key, so that I can figure out when/why this happens. Once I have that information, I should be able to remove the log message and handle this case another way.

### DIFF
--- a/studies/tasks.py
+++ b/studies/tasks.py
@@ -522,10 +522,10 @@ def get_all_incomplete_video_files():
             Bucket=settings.S3_BUCKET_NAME
         )
     except ClientError as error:
-        logger.error(f"Failed to list multipart uploads: {error}")
+        logger.error(f"Failed to list multipart uploads due to a ClientError: {error}")
         raise error
     except ParamValidationError as error:
-        logger.error("Failed to list multipart uploads")
+        logger.error("Failed to list multipart uploads due to a ParamValidationError")
         raise ValueError(f"The parameters you provided are incorrect: {error}")
 
     if uploads_response is not None and "Uploads" in uploads_response:
--- a/studies/tasks.py
+++ b/studies/tasks.py
@@ -538,6 +538,10 @@ def get_all_incomplete_video_files():
                 > datetime.timedelta(hours=24)
             )
         ]
+    else:
+        logger.debug(
+            f"Uploads key not found in list of multipart uploads for bucket {settings.S3_BUCKET_NAME}: {uploads_response}"
+        )
 
     return incomplete_uploads
 
--- a/studies/tasks.py
+++ b/studies/tasks.py
@@ -602,16 +602,23 @@ def complete_multipart_upload(filename, id, parts):
             MultipartUpload={"Parts": parts},
             UploadId=id,
         )
-        if resp["ResponseMetadata"]["HTTPStatusCode"] == 200:
-            logger.debug(f"Completed file {filename}")
+        if (
+            resp is not None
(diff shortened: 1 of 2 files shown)


mekline (approved):

bleonar5 (approved):

studies/tasks.py
```diff
@@ -526,7 +526,7 @@ def get_all_incomplete_video_files():
     except ParamValidationError as error:
         raise ValueError(f"The parameters you provided are incorrect: {error}")
 
-    if uploads_response["Uploads"]:
+    if "Uploads" in uploads_response:
         # Filter out incomplete uploads that might still be actively recording - started in last 24 hours.
```
okaycj, line 530: This assumes that `uploads_response["Uploads"]' is always assigned.  Are we certain it will be?

okaycj (changes requested):
This looks good otherwise.  Thank you.

studies/tasks.py
becky-gilbert: I think this assumes that `uploads_response` exists and tests if "Uploads" is a key that exists in the dictionary. 
But this is a great point - I don't think I've done enough here to prevent other/similar errors. I've updated this PR with a few more changes to this logic and tests.

becky-gilbert: Oooh, I see, do you mean that I need to check for something like `uploads_response["Uploads"]` exists but is null/None?

okaycj: After the API call, I would check that `uploads_response` is not None.  Then I would check if there's a key for 'Uploads' and if the value at that key is not "None": `"Uploads" in uploads_response and uploads_response["Uploads"].`

okaycj: It's basically what you had before combined with what you have now.

becky-gilbert: Oh sorry I forgot to point out in my comment that I'd also updated this code to
```
if uploads_response is not None and "Uploads" in uploads_response:
```

becky-gilbert: @okaycj Can you have a look at this one more time? I moved the bit that handles the multipart uploads response into try/except, and did the same with the list parts response handling (same problem there - the "Parts" key is missing from the response when there are no parts). 

I also added some extra validation logic that I missed before, and more tests.

And I did test this on our S3 buckets and verified that it works. (At least, it handles the cases that I was able to test with our buckets, which are (1) incomplete uploads that can be completed, (2) uploads that were created but could not be completed because they do not have any associated parts).

okaycj (approved):
👍

PR #1503, opened 2025-02-04
Completed exit frame

### LINKED ISSUE #779: Change study field "completed" to "completed_exit_frame"
**TL;DR**: This field's name is confusing, changing it would make life easier. 

**Narrative**
This field should be named completed_exit_frame as that is more inline with its actual function.  Additionally, we should change the researcher-facing labelling as well.

**Acceptance Criteria**

- [ ] Tests
- [ ] Scoping
- [ ] Alternatives Considered
- [ ] Rico sign-off

**Implementation Notes**
This should be straight forward migration.

okaycj:
# Summary

Change the Response field "completed"  to "completed_exit_frame".

Closes #779


mekline:
Naive question but does the testing tell us if this is succeeding for both EFP and JsPsych style responses?

okaycj:
> Naive question but does the testing tell us if this is succeeding for both EFP and JsPsych style responses?

Great question.  I am looking into versioning the API so as not to break existing experiences.

becky-gilbert:
@okaycj this all looks good so far! I know you still need to do the API versioning so I'll wait until then to test and review.

PR #1504, opened 2025-02-06
change filename of psychdsignore in response download

bleonar5:
This is a one-line change to the psych-DS response download function, with the name of the generated ignore file being changed from .psychdsignore to .psychds-ignore


becky-gilbert (approved):

okaycj (approved):
💯

PR #1512, opened 2025-02-20
remove initial backslashes from Psych-DS download zip

bleonar5:
In the Psych-DS download response view, a portion of the files added to the zip did not include a "/" at the beginning of the filepath. For mac users, everything was being included in the zip file, but for windows users, only the files without the initial "/" were included. This commit removes the initial "/" from all relevant filepaths


mekline (approved):
nice!

becky-gilbert (approved):
Nice catch!

PR #1513, opened 2025-02-24
Add individual responses table features

### LINKED ISSUE #1506: IR page: Add features to responses table
See https://github.com/lookit/lookit-api/issues/1497#issuecomment-2660041924 for table contents/layout reference.

This issue is just for adding the following functionality to the responses table:

- [x] A "Show N entries" drop down option
- [x] A global search bar
- [x] Ability to filter/sort by column value
- [x] Pagination that includes "Showing X to Y of N entries" and buttons for moving to previous/next pages and specific page numbers
- [ ] Show/hide column options

Other than show/hide, all of these features are available on the "Previous Messages" table (Contact Participants page), so I suggest modeling this table on that one, which would mean using the DataTables package. However, the current "Responses" table is paginated on the backend, which means that only one page of data is loaded at a time, and a new page makes a new request to the server. To use DataTables, we would need to load all of the responses each time the page is loaded. This would mean changing the way that GET request works, and it would make the initial page load slower (but the searching would be faster). Alternatively, we can keep this as-is and do the searching/filtering on the backend (which would make the initial page load faster but searching/filtering would be slower).

Here's a mock up, but ignore the table headers/content as that will be changed separately.

Image

becky-gilbert:
Fixes #1506 (except for show/hide columns - that will go in a separate PR)

This PR adds search/filter features to the Individual Responses table. In doing so, we decided to use datatables and remove the backend pagination (to more easily search/filter all responses on the frontend). Here are the changes:

- removed backend pagination on the study responses list view
- removed ordering (sorting) via URL parameters, as this is now done in JS, and always return the full set of responses sorted by date (descending order)
- update the IR page template: replace the old table with a datatables version, and remove the BS icons, URLs and query transforms that are no longer needed
- update the IR page JavaScript file: add init/config for the table
- put JS for Date column filtering from Contact Participants page into a separate file for reusability, and use for Individual Responses page as well
- minor changes to Date filtering: remove initial filtering when page loads and unfilter when (1) user clicks "Clear" in date range picker UI, (2) the Date filter textbox value is empty, or (3) the date range is invalid.

## Screenshots/Videos

When page loads, the first row is selected and highlighted. The results are sorted by descending date, which is indicated by the active sort arrow on the Date column:

page_load

Date range filter
Note that the selected response can be filtered out of the results that are displayed, but it remains selected.

https://github.com/user-attachments/assets/183457cd-d6a3-4788-babb-30bcb1921c20

Global search and filter works for partial strings, including date strings:

child_id_filter

global_search

global_search_date

You can sort the results by other column values, such as the Preview status:

preview_sort


becky-gilbert:
And this is to show you that the date filtering on the "Previous Messages" table (Contact Participants page) works the same way as on the Individual Responses table, since I've refactored that bit of code so that it's used for both tables.

https://github.com/user-attachments/assets/f88891dc-77a8-4ccb-886b-d72c8ceeb4de

becky-gilbert:
@okaycj this is ready for review, unless we want to merge #1514 first, and then I can fix the merge conflicts and add that here for review too.

okaycj (approved):
💯

becky-gilbert:
Update: Here what this page looks like now, after merging the changes from #1514 and fixing conflicts in the template:

Screenshot 2025-02-27 at 9 37 33 AM
Screenshot 2025-02-27 at 9 37 54 AM

okaycj (approved):
Nice work!  The only thing I found was that the column filter for the Date acts odd if you accidentally click it.  If you click off the date picker, it'll just fill in a date range.

becky-gilbert:
@okaycj I updated this and am happy that it works, but it's not pretty. I agree with your suggestion to move from `daterangepicker`to something else, if/when one of us has time to do that. In the meantime, is it ok to merge this?


(Ignore the extra columns in this table, they're not included in this PR. I was just on the wrong branch).

https://github.com/user-attachments/assets/f5c817cc-39e6-44f3-9b9f-5df6342c8b6f


https://github.com/user-attachments/assets/43ef5054-7716-41c7-afcd-6e7337b78b05

okaycj (approved):
That works pretty good!  Thanks for the extra change.

PR #1516, opened 2025-03-03
Individual response info box

okaycj:
# Summary

The grey info box at the top right side of the individual response view is now updated with selections from the table.  Additionally, the link to contact participants' view will auto-populate the recipients field if their filter allows. 

## Screenshots

Screenshot 2025-03-04 at 1 01 49 PM

https://github.com/user-attachments/assets/43ee6a73-7abe-40dc-a92b-385f58de138a

https://github.com/user-attachments/assets/fed09f4c-a207-4804-a879-49995196a8a7

### DIFF
--- a/exp/tests/test_contact_views.py
+++ b/exp/tests/test_contact_views.py
@@ -1,4 +1,5 @@
 import datetime
+import uuid
 from unittest.mock import patch
 
 from django.contrib.sites.models import Site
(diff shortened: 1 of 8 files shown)


mekline:
Re: the final line ("DHFS-kaitlin") - check the table on the Message Participants table, which uses this convention, but I *believe* it's (user/participant hashid)-(user/participant nickname field).

Issue #1507, opened 2025-02-18
IR page: New table content


becky-gilbert:
Make the following changes to the "Individual Responses" table.

Changes to existing content:
- [x] Change "Child ID" value from just hashed ID to hashed ID + first name, e.g. ABC123-becky (same convention used in the "Previous Messages" table). #1517
- [x] Change "Response UUID" column name to "Response ID" and use ID rather than UUID as the value. #1517
- [x] Move "Date" column so that it comes after "Response ID". #1517
- [x] Add a "Time Elapsed" column with the number of days since the Response was created. - @okaycj  #1518 
- [x] Change "Status" column name to "Exit Frame Status". #1517

New content - needs to be added to the database:
- [x] Add column "Payment Status" where the value is a drop-down box with options: (blank), "Needs review", "To pay", "Don't pay", "Paid". Default value is (blank). -- **@mekline CONFIRMED** -- #1517
- [x] Add column "Session Status" where the value is a drop-down box with options: (blank), "To schedule", "Scheduled", "Session attended", "Session complete", "Follow up", "Communication complete", "Withdrawn or closed".  Default value is (blank). -- **@mekline CONFIRMED** -- #1517
- [x] Add column "Star" with value that is a boolean checkbox (perhaps a clickable star ⭐). Default is false (unchecked). #1517

New features:
- [x] Changing the researcher-defined response column values should update these values in the database. #1524
- [x] Add UI cues while database update is pending, and catch the case where the update fails (e.g. retrieve the previous value, either in JS or by reloading the page) #1524
- [x] Prevent researchers from changing these values if they do not have the correct permissions: table input elements should be disabled - @okaycj #1531
- [x] Prevent researchers from changing these values if they do not have the correct permissions: POST request should fail (#1524)
- [x] Add drop-down filter for the new columns ("Payment Status", "Session Status", "Star") and fix filter for "Exit frame Status" (currently text but should be drop-down) - @okaycj  #1527 #1529
- [x] Add Show/Hide options for all columns - @okaycj <--- Wait on this until we've seen the full table, I think we should punt on this unless crowding is a real problem (- @mekline) #1520

Other to do:
- [x] Improve table styling: #1521 #1523
- [x] Fix time formatting and timezone inconsistencies: #1522
- [x] Add tests #1530
- [x] Update documentation (IR screenshots, new info/buttons, column sorting/filtering, table searching) https://github.com/lookit/lookit-docs/pull/478

**Implementation notes**

[...]

PR #1517, opened 2025-03-05
Individual Responses table content updates

becky-gilbert:
This resolves some of #1507 (but should not close that issue).

Changes:
- Change "Child ID" value from just hashed ID to hashed ID + first name
- Change "Response UUID" column to "Response ID" 
- Change "Status" (complete/incomplete) column name to "Exit Frame Status"
- Reorder columns
- Add column "Payment Status" where the value is a drop-down box with options: (blank), "Needs review", "To pay", "Don't pay", "Paid". Default value is (blank). 
- Add column "Session Status" where the value is a drop-down box with options: (blank), "To schedule", "Scheduled", "Session attended", "Session complete", "Follow up", "Communication complete", "Withdrawn or closed". Default value is (blank).
- Add column "Star" with value that is a boolean checkbox (no fill = false, fill = true). Default is false.

Screenshot 2025-03-06 at 11 29 21 AM


okaycj (approved):
This looks good!  Thank you.

PR #1522, opened 2025-03-06
Fixed date field to show correct time based on tz

okaycj:
# Summary

The date column was showing in the default time zone (EST).  Should not be showing local time.

### DIFF
--- a/studies/templates/studies/study_responses.html
+++ b/studies/templates/studies/study_responses.html
@@ -2,6 +2,7 @@
 {% load django_bootstrap5 %}
 {% load exp_extras %}
 {% load web_extras %}
+{% load tz %}
 {% load static %}
 {% block title %}
     Responses | {{ study.name }}
@@ -109,7 +110,7 @@
                                                 Incomplete
                                             {% endif %}
                                         </td>
-                                        <td>{{ response.response__date_created|date:"n/j/Y g:i A"|default:"N/A" }}</td>
+                                        <td>{{ response.response__date_created|localtime|date:"n/j/Y g:i A"|default:"N/A" }}</td>
                                         <td>{{ response.response__date_created|timesince }}</td>
                                     </tr>
                                 {% endfor %}
--- a/web/static/js/study-responses.js
+++ b/web/static/js/study-responses.js
@@ -122,7 +122,7 @@ function updateInfoBox(index) {
     // Create date for response with formatted date
     document.querySelector('.response-date').textContent = moment(
         rows[2].children[1].textContent
-    ).format('M/D/YYYY h:m A');
+    ).format('M/D/YYYY h:mm A');
 
     // Parent Feedback
     document.querySelector('.parent-feedback').textContent =


studies/templates/studies/study_responses.html
```diff
@@ -109,7 +110,14 @@
...
                                         </td>
-                                        <td>{{ response.response__date_created|date:"n/j/Y g:i A"|default:"N/A" }}</td>
+                                        <td>
+                                            {# djlint:off #}
+                                            <!-- This turns off the converstion from UTC to the project TIME_ZONE (America/NY). -->
+                                            {% localtime off %}
+                                                {{ response.response__date_created|date:"n/j/Y g:i A"|default:"N/A" }}
+                                            {% endlocaltime %}
+                                            {# djlint:on #}
+                                        </td>
```
becky-gilbert, line 120: I had to turn off djlint here because it kept automatically indenting this incorrectly, which caused the rest of the file to also indent incorrectly ☹️

studies/templates/studies/study_responses.html
```diff
@@ -299,7 +307,9 @@ <h5 class="card-title">Response details</h5>
                                     {% for pair in response.summary %}
                                         <tr class="{% if pair.name == 'Video privacy level' %} {% if pair.value == 'public' %} list-group-item-success {% elif pair.value == 'scientific' %} list-group-item-warning {% else %} list-group-item-danger {% endif %} {% endif %}">
                                             <td tabindex="0" title="{{ pair.description }}">{{ pair.name }}</td>
-                                            <td>{{ pair.value }}</td>
+                                            {# djlint:off #}
+                                        <td>{{ pair.value }}{% if pair.name == 'Date created' %} UTC{% endif %}</td>
+                                            {# djlint:on #}
```
becky-gilbert, line 312: I also turned off djlint here because it kept putting the `pair.value` and `if` template tags on separate lines, which introduced lots of empty space in all of the `td` elements, which in turn broke the string manipulation used to grab the parent name/IDs from this table.

studies/templates/studies/study_responses.html
okaycj: Does the "localtime" not work?

studies/templates/studies/study_responses.html
okaycj: I'll remove djlint.

studies/templates/studies/study_responses.html
becky-gilbert: No, the `localtime` filter just converts it to the `TIME_ZONE` value from the project's settings , which is Eastern.
> The localtime filter in Django templates converts a timezone-aware datetime object to the local timezone, as defined by the TIME_ZONE setting in your Django project.

But this conversion already happens automatically in our project because we have `USE_TZ = True`, so the `localtime` filter doesn't do anything in our case. The `{% localtime off %}` tag passes the date to the template in UTC rather than Eastern by overriding the project's `USE_TZ = True` setting. 

The use of an EST time here might be fine if it was used consistently, but we also give timestamps in UTC (in the Response Details box and in data file downloads). So I think the use of Eastern time in the table is confusing for people who are not on Eastern time.

becky-gilbert:
@okaycj Thanks for removing djlint! Is this ok to merge now?

Also do you think we need to replace djlint with something else? If so I can write up an issue.

PR #1523, opened 2025-03-06
Add class to style dropdowns

okaycj:
# Summary

Add class to drop-downs to match Bootstrap style. 

# Screenshots

Screenshot 2025-03-06 at 5 32 13 PM

### DIFF
--- a/studies/templates/studies/study_responses.html
+++ b/studies/templates/studies/study_responses.html
@@ -136,7 +136,7 @@
                                         <td>
                                             <select name="payment-status"
                                                     id="payment-status-{{ forloop.counter }}"
-                                                    class="researcher-editable"
+                                                    class="researcher-editable form-select"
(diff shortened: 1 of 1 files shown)


becky-gilbert (approved):

becky-gilbert:
Looks great!

I like the fact that the select elements aren't as wide as the widest choice, which reduces those column widths. However this also means that the selected option labels are cut-off, which makes them harder to read after they've been selected. Not a big deal, except that once the page/table gets narrow enough, you can't see the difference between "Session attended" and "Session complete" selections (screenshot below).

If this is a problem, we could solve it either by adding a min-width to that column, or relabelling the choices so that they don't start with the same word (e.g. "Attended session" and "Completed session").

Screenshot 2025-03-06 at 9 52 56 PM

PR #1524, opened 2025-03-07
IR Table: Update database when values change

becky-gilbert:
This PR adds data saving to the researcher-editable fields in the Individual Responses table.
- Adds JavaScript to update the Response object in the database when the researcher-editable column values changes.
- Adds an endpoint that processes the POST request sent from the IR page. The `StudyResponseSetResearcherFields` view checks the following:
  - User is a researcher with `EDIT_STUDY_FEEDBACK` permissions for this study
  - Response object exists and is a response to this study
  - The field that was changed is one of the researcher-editable Response fields (session_status, payment_status, star)
  - The value is valid for that field - one of the allowed status options (for payment/session status) or a boolean (star).
- Adds a CSRF token and the endpoint URL to the IR page template.
- Adds CSS for the input element that was changed while the POST request is still pending. The input element will become disabled with a 'not allowed' cursor type until a success/error response is received.

Note: I'll put in a separate PR with tests.

## Screenshots

(Now updated with latest formatting changes from develop)

https://github.com/user-attachments/assets/a3c90776-82fe-4a1a-bca4-245ed2925cfd

### DIFF
--- a/web/static/js/study-responses.js
+++ b/web/static/js/study-responses.js
@@ -135,3 +135,57 @@ function updateInfoBox(index) {
     document.querySelector('.parent-id').textContent = parentId;
     document.querySelector('.contact-family').href = url.toString();
 }
+
+// For updating the researcher-editable input elements in response table
+function getCsrfTokenAndUrl() {
+    const csrfTokenEl = document.querySelector('#csrftokenurl');
+    const token = csrfTokenEl.value;
+    const url = csrfTokenEl.dataset.updateUrl;
+    return { token, url };
+}
+
+// Send the researcher-editable field/value to the server when the user changes those input elements
+$('.researcher-editable').change(
+    function (event) {
+        const target = event.target;
+        target.disabled = true;
+        const currentResponseId = target.closest('tr').dataset.responseId;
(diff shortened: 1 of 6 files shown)


web/static/js/study-responses.js
```diff
@@ -148,3 +148,58 @@ function updateInfoBox(index) {
...
     document.querySelector('.contact-family').href = url.toString();
 }
+
+// For updating the researcher-editable input elements in response table
+function getCsrfTokenAndUrl() {
+    const csrfTokenEl = document.querySelector('#csrftokenurl');
+    const token = csrfTokenEl.value;
+    const url = csrfTokenEl.dataset.updateUrl;
+    return { token, url };
+}
```
okaycj, line 158: This works just fine.  Don't change a thing.  I wanted to point out you can get this value from the session cookie. https://docs.djangoproject.com/en/5.1/howto/csrf/#using-csrf-protection-with-ajax

okaycj (approved):
This looks like almost exactly what we talked about.  Nice work.

mekline (approved):

PR #1527, opened 2025-03-11
Ir table col filters

okaycj:
# Summary

This adds column filter/sort (and in some cases general filters) for Frame Exit, Payment, and Session Status columns.  

## Screenshot

https://github.com/user-attachments/assets/676100d6-b350-49c4-bbb9-7deccd7dc408

### DIFF
--- a/web/static/js/study-responses.js
+++ b/web/static/js/study-responses.js
@@ -70,9 +70,7 @@ function filterText(table) {
             // Select input element for this column and apply search
             $('input', this.footer()).on('keyup change', function () {
                 if (column.search() !== this.value) {
-                    column
-                        .search(this.value)
-                        .draw();
+                    column.search(this.value).draw();
                 }
             });
         });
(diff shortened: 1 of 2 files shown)


web/static/js/study-responses.js
```diff
@@ -47,7 +87,7 @@ function showHideColumns() {
     const hide_show_elements = show_hide_columns.map(el => {
         return `<a href class="toggle-vis" data-column="${el.col}">${el.name}</a>`
     })
-    document.querySelector('.show-hide-cols').innerHTML = `<div class="text-center py-sm-2">Show/Hide: ${hide_show_elements.join(" | ")}</div>`
+    document.querySelector('.show-hide-cols').innerHTML = `Show/Hide: ${hide_show_elements.join(" | ")}`
```
becky-gilbert, line 90: Thanks for taking out the inline HTML/CSS here 👍

becky-gilbert (changes requested):
I tried this out and found that the drop-down filters aren't returning all of the rows that match the selection. For some reason, rows that are farther down the table (e.g. past row 10) and have a matching selection are not being returned. I haven't figured out the cause yet but will update when I do.

okaycj:
> I tried this out and found that the drop-down filters aren't returning all of the rows that match the selection. For some reason, rows that are farther down the table (e.g. past row 10) and have a matching selection are not being returned. 

It looks like adding the change event isn't being applied to rows off screen.

web/static/js/study-responses.js
```diff
@@ -110,41 +172,14 @@ $('.star-checkbox').change(function () {
...
-        rows[15].children[1].textContent;
-
-    // Create date for response with formatted date
-    document.querySelector('.response-date').textContent = moment.utc(
-        rows[2].children[1].textContent
-    ).format('M/D/YYYY h:mm A z');
+// On dropdown cell change, update their td's sort and filter values.
+resp_table.on('change', '.dropdown-cell', function () {
+    const td = this.parentElement;
+    const text = this.options[this.selectedIndex].text
```
becky-gilbert, line 178: @okaycj we decided that this change to dataset values needs to be integrated with the ["change" event callback that triggers the POST request](https://github.com/lookit/lookit-api/blob/3acdb72bfeb57ebd41c2e77e72ea7a6c73a922d2/web/static/js/study-responses.js#L161-L205), right? If so I will wait to review this.

okaycj: Thanks!  Updated in commit [954fa55](https://github.com/lookit/lookit-api/pull/1527/commits/954fa55bfac3983c350c530ffffa43d5bc1c32ca).

becky-gilbert:
I fixed a problem with this PR, only to realize that it is actually fixed (in a better way) in #1529, so I reverted the commit.

becky-gilbert (approved):

PR #1530, opened 2025-03-11
Add tests for new researcher-editable response fields

becky-gilbert:
This PR adds tests for the new exp `StudyResponseSetResearcherFields` view that is used to update the researcher-editable fields for responses. This view accepts a POST request sent from the IR page when a researcher changes the value of one of the editable fields in the table.

- Researchers without 'edit study feedback' permissions for the study cannot change any of the researcher-editable response values (POST request fails with 403 Forbidden).
- Researchers with 'edit study feedback' permissions for the study can change any of the researcher-editable response values (POST request succeeds with 200 ok and response field/value is changed in database).
- The new researcher-editable fields are included in the CSV and JSON response data downloads.
- The POST request fails for the following reasons (with appropriate error message):
  - missing data
  - invalid value for field
  - invalid (un-editable) field
  - response does not exist
  - response exists but is not associated with this study
- The POST request succeeds with valid blank (empty string) and `False` values

Note: this PR is failing the code duplication check (by .4% :weary:). I need to figure out how to ignore test files in this check, or refactor the code.

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,2 +1,2 @@
 sonar.python.version=3.9
-sonar.cpd.exclusions=./**/tests/**/*,./studies/test*
+sonar.cpd.exclusions = **/tests/**/*
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,1 +1,2 @@
 sonar.python.version=3.9
+sonar.cpd.exclusions=./**/tests/**/*,./studies/test*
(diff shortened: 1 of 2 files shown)


.sonarcloud.properties
becky-gilbert, line 1: @okaycj I tried to exclude our test files from sonarcloud code duplication checks here but it didn't work. Maybe I did it wrong (docs: https://docs.sonarsource.com/sonarqube-server/latest/project-administration/analysis-scope/) Ideas?

okaycj: I would look at [this part of the docs](https://docs.sonarsource.com/sonarqube-server/latest/project-administration/analysis-scope/#project-level).  Maybe add something like `sonar.exclusions = **/tests/**/*` to the Sonarcloud properties file.

okaycj: Or maybe try `sonar.cpd.exclusions = **/tests/**/*`. Which is closer to what you where trying in this PR.

becky-gilbert: This worked, thanks!

okaycj (approved):
💯  Thank you for getting these test done.

PR #1534, opened 2025-03-13
IR page updates

becky-gilbert:
This production release updates the Individual Responses page with the changes outlined here: #1507

It also releases some other bug fixes and minor changes:
- Fix key error in incomplete videos cleanup task #1496
- Change name of PsychDS ignore file #1504
- Fix PsychDS zip path issue #1512


becky-gilbert:
Becky QA testing:
- Info about pending/rejected consents is correct
- Zeros in the pending/rejected counts were showing up as empty strings, now fixed: https://github.com/lookit/lookit-api/pull/1535
- Link to consent manager works 
- Link to send message to family with correct recipient filter (question about their response)
- Link to consent manager fills in family ID when they allow that type of email, and displays info message when they do not
- Column show/hide works
- Table search works on Child ID, Response ID, Date columns (not others - ok or create issue?)
- Page with no consented responses displays correctly
- Date column filter works with daterangepicker range options and text input changes
- Date column was sorting as string, now fixed: https://github.com/lookit/lookit-api/pull/1536
- Data saving works for researcher-editable columns
- Input fields are disabled while data saving is pending
- Drop-down filters work for Payment/Session status, including rows not displayed
- Sorting works for Star column, including for rows not displayed
- Rows displayed with an active sort/filter are updated after a value is changed and the value change has been updated successfully in database

okaycj (approved):
:shipit:

PR #1535, opened 2025-03-13
Display a 0 instead of empty string

okaycj:
# Summary
In the info row on the IR view, the statistics data displayed an empty string when there was no data for a category.  We'll instead have a 0 when there's no data.


becky-gilbert (approved):
Looks great. I tested this on studies with 0 rejected and 0 pending, and it works. Thanks!

PR #1536, opened 2025-03-13
IR table: fix bug in date column sorting

becky-gilbert:
This PR fixes a bug in the Individual Responses table, where the Date column was being sorted as a string rather than date.


okaycj (approved):
💯

PR #1538, opened 2025-03-17
Add short hash to the end of static file names

okaycj:
# Issue

After releases there are some who find the site broken.  This is due to their browser having cached files that need to be updated.  

# Solution

Add a short md5 hash to the the end of static files.  

## Note

This _should_ work, but testing on staging will be needed.

### DIFF
--- a/project/storages.py
+++ b/project/storages.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.staticfiles.storage import ManifestFilesMixin
 from storages.backends.gcloud import GoogleCloudStorage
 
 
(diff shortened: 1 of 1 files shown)


becky-gilbert (approved):
I tried and failed to make this file name hashing work locally. I think you're right that we'll need to test this on staging, since there a number of differences between local and remote. I did check that we have the other required settings in place and it looks like we do.

mekline:
(I'm confused why this is happening now as opposed to other updates - any insight? No need to go on a goose chase if this is just a one-off, but does this for instance need to be part of the usual deployment checklist? Does the hash addition in fact risk breaking anything else?)

PR #1539, opened 2025-03-17
Remove poetry

okaycj:
# Issue

[Poetry](https://python-poetry.org/)  has been our package manager for a bit now and it's been... fine.  Recently, they've incremented their project version 2 with breaking changes.  To resolve this issue, we'd have to re-construct our `pyproject.toml` file using their their CLI.  This would mean we'd have to enter each package and its version into poetry which is the same work as changing to another package manager.  

# Solution

I'd like to pitch using [UV](https://github.com/astral-sh/uv) as the new package manager.  UV is a new but seemingly well liked project.  It's fast, easy to use, and using Github projects pinned to a commit was simple.  I have yet to find a major downside.


becky-gilbert (approved):
Looks great in principle! I wanted to test it locally today but ran out of time. Maybe another dev could try this out? 
Or is it the case that (1) we'll know that it works after merging into develop and staging comes back up, and (2) if there is a problem, then undoing this change will be straightforward? If so then I say 👍

becky-gilbert:
@okaycj I wasn't able to build the `develop` branch locally without updating gevent and greenlet. Here's my diff:

```diff
-    "gevent==21.12.0",
+    "gevent==23.9.1",
     "google-cloud-storage==2.1.0",
-    "greenlet==1.1.2",
+    "greenlet==2.0.0",
```

But I suppose these packages might get updated anyway with the subsequent Django update PRs, so I'll wait to see if this is still an issue after I get through those.

PR #1545, opened 2025-03-21
IR page: replace bootstrap star icons with fontawesome

becky-gilbert:
This PR replaces the bootstrap icon template tags with fontawesome icons as the stars in the "Star" column.

For some reason, the `bs_icon` template tags we were using to generate the filled/unfilled stars for the "Star" column were causing significant page load delays. I looked into options for replacing them, and found that loading fontawesome icons from a kit link is faster. 

## Timing results

I tested the loading times by finding a study in my local DB that has a large number of responses (997). I loaded the Individual Response page for this study on two different branches: develop and this one (ir-page-timeout-icons). Here are the loading times on the two branches, with browser caching disabled:

| branch  | load time (min) | load time (sec) |
| ------------- | ------------- | ------------- |
| develop | 1.87 | 112 |
| icons | 0.55 | 33 |

There are some studies on production with even more responses (the most is around ~2000), so it's possible that some studies on production would still produce a 504 timeout even with these improvements.

### Develop

develop_study_94_load_time
develop_study_94_debug_time
develop_study_94_debug_profiling

### This branch

icon_study_94_load_time
icon_study_94_debug_time
icon_study_94_debug_profiling

### DIFF
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -839,10 +839,13 @@ class StudyResponsesList(CanViewStudyResponsesMixin, generic.ListView):
     model = Response
 
     def get_queryset(self):
+        video_queryset = Video.objects.only("full_name")
+
         study = self.study
         return (
             study.responses_for_researcher(self.request.user)
             .order_by("-date_created")
+            .select_related("study_type", "child", "demographic_snapshot")
             .prefetch_related(
                 "consent_rulings__arbiter",
                 Prefetch(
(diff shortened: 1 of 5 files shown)


becky-gilbert:
This is ready for review, but I do need to double check that changes to the star field are still working (clicking a star should toggle its gray/false and yellow/true states, and send a POST request to update the value in the correct response). I will update here when I've done those tests.

becky-gilbert:
Update: I'm happy that the data saving still works. Here is a demo of changing/saving the star status, and screenshots of the new icon styling.

https://github.com/user-attachments/assets/7406a1ff-55ab-44c2-b700-a024c9d033d4

star_filled_selected

star_unfilled_selected

okaycj (approved):
Whoa!  I was not expecting that.

PR #1546, opened 2025-03-22
IR page: improve queryset

becky-gilbert:
This PR updates the queryset for the Individual Responses page view to reduce the number of hits to the database. A number of other object lookups related to the study and response are required for this view: study type, child, participant/user, demographic snapshot, videos. These relationships were causing the [N+1 query problem](https://dev.to/herchila/how-to-avoid-n1-queries-in-django-tips-and-solutions-2ajo) because they were being retrieved separately for each response in a loop.

## To do

I used `select_related` to get the related objects that we need for the relevant set of Response objects, but this could probably still be improved.

For instance, we could try just grabbing the specific columns/values we need from each of these foreign key objects. But this is tricky because the info we need about each response comes from `RESPONSE_COLUMNS`, which is not a list of DB columns but rather a set of response fields that contains an extractor function. The extractor function sometimes just returns a column value, but sometimes it is used to calculate a value based on other values (e.g. child age in days comes from the response `date_created` and the child `birthday` fields). So to do this we would need to gather all of the DB columns that are used by extractor functions in the relevant set of `RESPONSE_COLUMNS`.

I used `prefetch_related` to get the Video objects related to each response, and limited this to just the column that we need ("full_name"). Not sure if this one could still be improved?

## Query results

I compared the number of queries and their total duration when loading the IR page in (1) develop and the branch with fontawesome icons (should be the same), and (2) this branch that tries to improve the SQL query efficiency.

| branch  | N queries | query duration (ms) |
| ------------- | ------------- | ------------- |
| develop/icons | 2005 | 835 / 782 respectively |
| this branch | 13 | 452 |

## Load time results

There was an improvement in page load time with these query changes, although it was not as drastic as with the change in icons.

| branch  | load time (min) | load time (sec) |
| ------------- | ------------- | ------------- |
| develop | 1.87 | 112 |
| icons | 0.55 | 33 |
| this branch | 0.30 | 18 |

### develop/icons

develop_study_94_sql_queries

icon_study_94_sql_queries

### This branch

queryset_study_94_sql_queries

queryset_study_94_load_time

### DIFF
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -845,6 +845,7 @@ class StudyResponsesList(CanViewStudyResponsesMixin, generic.ListView):
         return (
             study.responses_for_researcher(self.request.user)
             .order_by("-date_created")
+            .select_related("study_type", "child", "demographic_snapshot")
             .prefetch_related(
                 "consent_rulings__arbiter",
                 Prefetch(
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -839,6 +839,8 @@ class StudyResponsesList(CanViewStudyResponsesMixin, generic.ListView):
     model = Response
 
     def get_queryset(self):
+        video_queryset = Video.objects.only("full_name")
+
         study = self.study
         return (
             study.responses_for_researcher(self.request.user)
@@ -850,6 +852,11 @@ class StudyResponsesList(CanViewStudyResponsesMixin, generic.ListView):
                     queryset=Feedback.objects.select_related("researcher").order_by(
                         "-id"
                     ),
+                ),
+                Prefetch(
+                    "videos",
+                    queryset=video_queryset,
+                    to_attr="prefetched_videos",
                 ),
             )
         )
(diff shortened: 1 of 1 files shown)


okaycj (approved):
Great work.  Two thoughts, consider them optional:
1. There are two `order_by` in this query, are they necessary?
2. The `get_consent_statistics` function was borrowed from another view, should this be simplified for our use on this view?

💯

mekline (approved):

becky-gilbert:
> There are two order_by in this query, are they necessary?

Great question. I _think_ the first `order_by("-date_created") relates to the responses themselves, and the order_by here:

```python
Prefetch(
    "feedback",
    queryset=Feedback.objects.select_related("researcher").order_by(
        "-id"
    ),
),
```

orders the feedback objects that relate to each response. But I could be wrong - I will double check!

> The get_consent_statistics function was borrowed from another view, should this be simplified for our use on this view?

Oooh, another great point. Yes you're right, this `get_consent_statistics` function also tallies the number of unique children (total and for each consent ruling type), which we don't need here. And this view is already super complex, so it is probably worth simplifying it however we can. If I'm not able to fix this quickly then I may put this in a separate issue to fix later. Thanks!

becky-gilbert:
> There are two order_by in this query, are they necessary?

Ok, I see what you mean about the response ordering via `order_by` in the query as well as in the JavaScript. I think you're right that `order_by` isn't necessary, but I'd like to leave it for now and put it in another issue. 

(Long answer: When the page first loads, the JS updates the summary, info box, and attachments using whatever response is in the first row, so I'm worried that this info could be incorrect if datatables then re-orders the rows. I'm not sure if this is even a problem, and if it is then solution might be just to move the initial info updating calls into datatables `initComplete`. But for now I'd like to get these updates out without having to do more testing or risk breaking things.)

PR #1548, opened 2025-03-27
IR table: sort date using Response ID

becky-gilbert:
This PR fixes a bug that was causing the date column not to sort properly. I believe the problem was caused by the date column losing access to the full timestamp. After multiple failed attempts at fixing this through datatables datetime formatting, I decided to sort the Date column on the Response ID value (which increases incrementally with the creation of each new response object).

@okaycj I'm not entirely sure why, but this change seems to fix the row re-sorting bug that was triggered by `updateAJAXCellData`, but please feel free to check too. I can also test again tomorrow and on staging.

### DIFF
--- a/web/static/js/study-responses.js
+++ b/web/static/js/study-responses.js
@@ -207,11 +207,11 @@ const resp_table = $("#individualResponsesTable").DataTable({
             { features: [{ div: { className: 'show-hide-cols text-center mx-3' } }] },
             'search'],
     },
-    order: [[3, 'desc']], // Sort on "Date" column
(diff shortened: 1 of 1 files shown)


mekline (changes requested):
Minuscule request to please include a comment to the effect of "we sort by responseID here even though that's weird because for some reason sorting by date doesn't work"

becky-gilbert:
> Minuscule request to please include a comment to the effect of "we sort by responseID here even though that's weird because for some reason sorting by date doesn't work"

Gah, yes of course, sorry for forgetting to do that! Will add it in now.

PR #1557, opened 2025-04-01
Individual Responses page: improve video prefetch queryset

becky-gilbert:
This PR modifies the Video Prefetch queryset to include all of the fields that are needed by the Individual Responses page. I added this Video Prefetch recently but somehow missed the fact that not all necessary fields were included, so the view was producing one additional query per Video to get the non-prefetched fields. 

Tested on Study 374 from staging. 

**Before (develop branch)**

Page load time: 3.61 seconds
Number of queries: 342, including 330 similar (these are the separate repeated queries for each video because some fields were left out of the prefetch)

IR_page_load_video_queryset_before

IR_page_sql_video_queryset_before

**After (this branch)**

Page load time: 1.02 seconds
Number of queries: 12, including 2 duplicates

IR_page_load_video_queryset_after

IR_page_sql_video_queryset_after

Note that there are still two duplicate queries happening with this view. It's not immediately clear to me where these are coming from. So this view might benefit from more optimization, but this change is at least a step in the right direction.

### DIFF
--- a/exp/views/responses.py
+++ b/exp/views/responses.py
@@ -908,7 +908,7 @@ class StudyResponsesList(CanViewStudyResponsesMixin, generic.ListView):
                 if col.id in columns_included_in_summary
             ]
 
-            video_info = [
+            this_resp_data["videos"] = [
                 {
                     "pk": v.id,
                     "display_name": (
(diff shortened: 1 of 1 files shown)


okaycj (approved):
💯 Love the iterative improvements.

mekline (approved):
Great!

becky-gilbert:
I've now tested this on staging with several studies that have high video counts, and have not encountered any problems. We will need to test again on production because of higher video numbers per study/response in that database.

PR #1562, opened 2025-04-15
Possible fix for `uv` build error

becky-gilbert:
### Description

This PR makes a few changes that fix some errors that I get on `develop` when I run `uv sync` (errors pasted below). 

The first error was about being unable to build the `gevent` package (pinned version 21.12.0) because of incompatibility with Cython (not listed/pinned). I tried downgrading my Cython version, but that didn't fix it. Upgrading my `gevent` version to 23.9.1 did fix the problem, but I also had to upgrade `greenlet` to 2.0.0.

The second error was related to setuptools trying to treat all top-level directories as Python packages. This just required adding `tool.setuptools.packages.find` to `pyproject.toml` with a list of the top-level directories that are actually Python packages.

### Changes in this PR

- Update `gevent` to 23.9.1 and `greenlet` to 2.0.0.
- Add `setuptools.packages.find` with the list of our top-level directories that are Python package.

### `gevent` build error

```zsh
% make uv
uv self update
info: Checking for updates...
success: You're on the latest version of uv (v0.6.14)
uv sync
Using CPython 3.9.22 interpreter at: /Users/beckygilbert/.pyenv/versions/3.9.22/bin/python3.9
Creating virtual environment at: .venv
Resolved 107 packages in 0.58ms
    Updated https://github.com/lookit/django-ace-overlay.git (4b174f78c6d6aaaee961e220c9461cbe74c861da)
    Updated https://github.com/Innovativity/django-revproxy.git (b9fa8375d03fd68747dcb7273a97c19d788aa51b)
      Built fleep==1.0.1
      Built pydenticon==0.3.1
      Built ciso8601==2.1.3
      Built sendgrid-django==4.2.0
      Built psycogreen==1.0.2
      Built greenlet==1.1.2
      Built django-revproxy @ git+https://github.com/Innovativity/django-revproxy.git@b9fa8375d03fd68747dcb7273a97c1
      Built uwsgi==2.0.22
      Built django-ace-overlay @ git+https://github.com/lookit/django-ace-overlay.git@4b174f78c6d6aaaee961e220c9461c                                                                                                              × Failed to build `gevent==21.12.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

      [stdout]
      Compiling src/gevent/resolver/cares.pyx because it changed.
      [1/1] Cythonizing src/gevent/resolver/cares.pyx
[...]

PR #1564, opened 2025-04-17
Update greenlet to 3.0.3

becky-gilbert:
This PR bumps `greenlet` to version 3.0.3. This change allows the use of a precompiled wheel for newer versions of macOS like Sonoma, which @deirdre-k is on.

@deirdre-k and I tried the current `develop` version on her system (macOS Sonoma, M1) and we found that `greenlet` 2.0.0 would not build and there was no precompiled wheel available for her system. Updating `greenlet` to 3.0.3 fixed the problem while maintaining compatibility with the rest of our stack and other developers' systems (macOS Ventura, M1/M2).

Here's a summary of the compatibility for the change here and with the `gevent` update from #1562.

- greenlet 3.0.3
  - supports Python 3.7–3.12
  - stable release
- gevent 23.9.1 
  - supports Python 3.7–3.12
  - stable release
  - compatible with greenlet v3 (requires greenlet>=2.0.0, but greenlet v3 is required on Python 3.11/3.12 and v3 is [recommended on all platforms](https://www.gevent.org/changelog.html#post1-2023-09-02))

Both of these package versions provide wheels for:
  - our developers' machines: macOS 11+ universal (M1/M2 arm64/aarch64, and Intel/AMD 64-bit)
  - our remote build environments (Debian GNU/Linux 10): manylinux glibc 2.17+/2.28+ x86-64
  - Windows x86-64

I've tested these package versions on the Django 4.0 branch and did not encounter any problems.

## To do:

After merging this PR:
- [ ] test on staging (same tests as in #1562)
- [x] give installation instructions to @bleonar5 and make sure he can build successfully --> success ✅ 
- [ ] continue testing on subsequent Django update branches to make sure these package versions don't cause compatibility issues

## References

- `greenlet` [v3.0.3 precompiled wheels](https://pypi.org/project/greenlet/3.0.3/#files:~:text=greenlet%2D3.0.3%2Dcp39,ARM64%2C%20x86%2D64) for Python 3.9
- `gevent` [v23.9.1 precompiled wheels](https://pypi.org/project/gevent/23.9.1/#files:~:text=gevent%2D23.9.1%2Dcp39,ARM64%2C%20x86%2D64) for Python 3.9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "fleep==1.0.1",
     "gevent==23.9.1",
     "google-cloud-storage==2.1.0",
-    "greenlet==2.0.0",
+    "greenlet==3.0.3",
     "kombu==4.6.11",
     "lark-parser==0.12.0",
     "more-itertools==8.8.0",
--- a/uv.lock
+++ b/uv.lock
@@ -653,19 +653,20 @@ wheels = [
 
 [[package]]
 name = "greenlet"
-version = "2.0.0"
+version = "3.0.3"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/23/6f/f72865c589d0c79f03b51520a4cdd65647de0513e9ad107a5731df89c235/greenlet-2.0.0.tar.gz", hash = "sha256:6c66f0da8049ee3c126b762768179820d4c0ae0ca46ae489039e4da2fae39a52", size = 161272 }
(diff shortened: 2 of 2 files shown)


becky-gilbert:
This update broke the staging server, which I believe is because of `gevent`/`greenlet` version compatibility issues. We can either downgrade `greenlet` or try upgrading `gevent`. My preference is to try upgrading `gevent` because: 
1. the reason for updating `greenlet` was to fix build problems on dev machines with newer OS versions, so reverting the `greenlet` version means that we will still have to solve that problem, and
2. we're updating other things anyway (our OS versions, Django, postgres, etc.) so it makes more sense to move everything forward in parallel and future-proof if possible.

If a `gevent` upgrade doesn't fix the problem, we should revert this `greenlet` update to (at least temporarily) fix staging.

PR #1565, opened 2025-04-21
Revert "update greenlet to 3.0.3 (#1564)"

becky-gilbert:
This reverts commit 84437c506508742454e1864cc2e673ce43070d24.

The staging server is broken (500 error) following #1564, so this PR reverts that change.


becky-gilbert:
The staging server was still broken after merging this PR, even though it was working before this (reverted) commit, which means that some part of the cloudbuild process was pulling from cache. I have just tested some changes to the cloudbuild file that seem to fix the problem. I will put in a PR for the cloudbuild changes.

PR #1567, opened 2025-04-22
Django 4.2.20

okaycj:
# Summary 

Bump Django to vers 4.3.20.

### Notes

- JSONDatetime class was heavily altered.  
- When calculating HMAC for videos, the field 'type' must be removed to pass tests.


api/permissions.py
```diff
@@ -78,6 +78,8 @@ def has_permission(self, request, view):
             except KeyError:
                 raise exceptions.ParseError("Missing required relationship: response")
 
+            request_data.pop("type")
+
```
becky-gilbert, line 82: I wonder if [this change](https://github.com/django-json-api/django-rest-framework-json-api/blob/main/CHANGELOG.md#deprecated-1) to `djangorestframework-jsonapi` caused the need for this line. 
> Deprecated support for field name type as it may not be used according to the [JSON:API spec](https://jsonapi.org/format/#document-resource-object-fields).

In any case, it would probably make more sense to remove the type field from the Lamba function code.

api/tests/test_responses.py
becky-gilbert, line 1: I looked into this because I thought it was weird that this ever worked. But indeed, before this update, these paths did have the incorrect "attributes" name instead of "relationships". I'm guessing it was a bug that was fixed, maybe [this one](https://github.com/django-json-api/django-rest-framework-json-api/blob/main/CHANGELOG.md#fixed-5): 
> Fixed invalid relationship pointer in error objects when field naming formatting is used.

becky-gilbert (approved):

PR #1568, opened 2025-04-22
Add `--no-cache` arg to build image steps

becky-gilbert:
This PR adds `--no-cache` to our two build image steps. 

The most recent build/deploy on staging was not clean (it still resulted in server 500 errors despite the fact that the rebuild was successful and I was reverting to a known working version). Adding the `--no-cache` argument seems to have helped - staging worked again after rebuilding with this change to `cloudbuild.yaml`. It's possible that triggering another rebuild without this change would also have worked, but even if so, this shouldn't cause any problems.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -40,6 +40,7 @@ steps:
   name: gcr.io/cloud-builders/docker
   args: [
     'build',
+    '--no-cache',
     '--tag',
     'lookit',
     '--build-arg',
@@ -54,6 +55,7 @@ steps:
   name: gcr.io/cloud-builders/docker
   args: [
     'build',
+    '--no-cache',
     '--tag',
     'lookit-test',
     '--build-arg',


okaycj (approved):

becky-gilbert:
Update: I'm now convinced that adding the '--no-cache' argument did not fix the earlier problem on staging. A similar thing happened again, and in this case it was fixed by running `kubectl rollout restart deployment -n default`. So I think what actually fixed the problem in this first case was just triggering another restart, rather than the changes to the cloudbuild.yaml. 

I'm going to close this because I don't think the no-cache changes are actually needed, and might have other negative consequences (slowing down the build, increasing cloud costs?). We can re-open this if we run into build/deploy problems in the future that are not fixed by a simple restart trigger and might be cache related (e.g. check build logs to see what is actually downloading/building vs coming from cache).

PR #1570, opened 2025-04-24
Python version bump

okaycj:
# Summary

Bumping version of Python to 3.13 before we try to bump the version of Django to 5.*.


becky-gilbert:
@okaycj WRT this CircleCI CSS error:
```
Downloading bootstrap v5.3.3....
Unzip bootstrap file...
Remove downloaded compressed file...
CommandError: Expected directory "bootstrap-5.2.0" was not found after unzip.
```
Does this mean that bootstrap 5.2.0 is no longer supported?

becky-gilbert:
@okaycj Can you say more about this change to `make uv`:
```
uv sync --no-managed-python --no-python-downloads
```
Does it affect how we set up the python version in our local environments?

This is what I have in my draft of the updated installation instructions - let me know if it needs changing:
- Set up the environment to use the correct python version (obviously this would need to be changed to 3.13!)
```
brew install pyenv
pyenv install 3.9.22 
pyenv local 3.9.22
python --version <-- should be Python 3.9.22
```
- install `uv`
```
curl -LsSf https://astral.sh/uv/install.sh | sh
```
- install everything else
```
make uv
```

okaycj:
> Does this mean that bootstrap 5.2.0 is no longer supported? Do we need to just change the directory that we're looking for? And hopefully there are no issues with the version update?

I changed the version of the 'django-bootstrap' package which also bumped the version of bootstrap.  I moved the version back.

okaycj:
> uv sync --no-managed-python --no-python-downloads

UV can (and will if the necessary version of python isn't available) download python to use in a virtual environment.  Unfortunately, the managed version of python doesn't compile uwsgi.  The additional arguments prevents the managed version of python from being used or downloaded.

becky-gilbert:
> I changed the version of the 'django-bootstrap' package which also bumped the version of bootstrap. I moved the version back.

Oh ok, thanks! Bumping bootstrap is fine too. I just wanted to understand the error and make sure I know what to test.

becky-gilbert:
@okaycj I believe the CircleCI test failed because of this issue: #1582 I have a fix that I will put in a separate PR.

becky-gilbert (approved):

PR #1581, opened 2025-04-29
Django 5.2

okaycj:
# Summary

Bump Django to version 5.2 

## Notes

- Remove pytz dependency


project/fields/datetime_aware_jsonfield.py
becky-gilbert, line 1: @okaycj so it looks like this is no longer used, except in migrations. Do you have any thoughts on leaving it in vs trying to take it out?

becky-gilbert: Sorry, I just saw your comment on this PR: #1567

Issue #1600, opened 2025-05-05
Celery beat tasks not sent on staging


becky-gilbert:
## Summary

There are some Celery beat tasks that are not running on staging. These tasks are supposed to run nightly but were last sent several days ago:

- cleanup_builds: 2025-05-01 22:00 EDT
- cleanup_docker_images: 2025-05-01 23:00 EDT
- cleanup_docker_containers: 2025-05-01 23:30 EDT
- cleanup_incomplete_video_uploads: 2025-05-04 01:00 EDT

I assume that this was caused by one of the updates that we made around May 2nd, such as the switch to DEBUG=True, python 3.13 (https://github.com/lookit/lookit-api/pull/1570) or Django 5.2 (https://github.com/lookit/lookit-api/pull/1581) updates. 

These tasks do not seem to have been affected:

- cleanup_checkouts: 2025-05-06 00:00 EDT
- send_announcement_emails: 2025-05-06 10:13 EDT (this should go out at 10am UTC / 6am EDT)
- backend_cleanup: 2025-05-06 04:00 EDT

At 5/6/25 around 3:32 PM EDT / 7:32 PM UTC the staging server was rebuilt with DEBUG=False and the beat container is still just reporting these 2 tasks as the schedule.
- studies.tasks.cleanup_checkouts: 4:00 UTC
- celery.backend_cleanup: 4:00 America/New York

It leaves out the other 5 periodic tasks. Deleting the pod did not fix the problem.

It seems to be picking up changes to the database (e.g. edits to django_celery_beat_crontabschedule and django_celery_beat_periodictask tables will result in "schedule changed" and "fetching database schedule" in the logs) but then it always reports the same incorrect subset of tasks, and changes to the crontab times are not reflected in the schedule it reports.

There are some reports of timezone related issues with celery beat, e.g. if `USE_TZ = False` then you also need `DJANGO_CELERY_BEAT_TZ_AWARE = False`, but that shouldn't apply to us because we have `USE_TZ = True`.

becky-gilbert:
Resolved via https://github.com/lookit/lookit-api/pull/1607

PR #1605, opened 2025-05-08
Add local beat and recreate crontabs

becky-gilbert:
This branch is to help with debugging celery-beat issues locally (see #1600). We may not actually need to do the migration step to fix the problem - I've added it here for testing.
- Adds the beat service to `docker-compose.yml` so that we can reproduce the celery-beat task scheduling issues locally (thanks @okaycj!)
- Adds a migration file that can be used to delete and re-create crontab entries with timezone="America/New_York"


becky-gilbert:
@okaycj This PR no longer contains the original changes that I added for debugging period task issues (local beat service and database migration for recreating the crontabs). Do you think the remaining changes here to Dockerfile/makefile/docker-compose are worth merging? Or should I close this?

PR #1608, opened 2025-05-14
Update django.po

baharsener:
Added some additional Turkish translations


baharsener:
Hello, I just added some Turkish translations to messages parents would see on the site. I do not have a good translation for the word 'toddler' in Turkish, for which I currently used 'babies (aged 1-2)' instead. Otherwise, I believe all other translations (e.g., how buttons are named) should be consistent. I am happy to add more in the future

okaycj (approved):
🥇 Thank you.  This is amazing.

PR #1617, opened 2025-05-19
Move emergency fix to production

okaycj:
# Summary

Until we have a more permanent fix, we'll do the following:

- Force image names to be lower-case when they're hard-coded in the DB.
- Skip checking image size on study ad form submission. 

These changes are to get things working.  A replacement fix will be up in the next day or so.

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -193,16 +193,16 @@ class StudyForm(ModelForm):
 
         return criteria_expression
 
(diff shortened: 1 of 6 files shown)


becky-gilbert (approved):

becky-gilbert:
Tested this on staging and it worked for me.

PR #1623, opened 2025-05-20
Add storages setting

okaycj:
# Summary

Added settings for our storage options.


project/settings.py
```diff
@@ -303,6 +303,17 @@
     GS_PRIVATE_BUCKET_NAME = os.environ.get("GS_PRIVATE_BUCKET_NAME", "")
     GS_QUERYSTRING_AUTH = False
 
+    STORAGES = {
+        "default": {
+            "BACKEND": "storages.backends.gcloud.GoogleCloudStorage",
+            "OPTIONS": {"location": "media"},
```
becky-gilbert, line 309: Aha! I think you're on the right track with this. You've probably already found this, but the options are listed at the bottom of the docs here: https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#settings

becky-gilbert (approved):

PR #1624, opened 2025-05-20
Custom endpoint settings

okaycj:
# Summary

Add custom endpoint setting to support static proxy.


project/storages.py
```diff
@@ -20,14 +19,17 @@ def url(self, name):
...
     # See https://github.com/lookit/lookit-api/issues/570
-    file_overwrite = False
+    # file_overwrite = False
+    pass
 
 
 class LookitExperimentStorage(LookitGoogleCloudStorage):
-    location = settings.EXPERIMENT_LOCATION
+    # location = settings.EXPERIMENT_LOCATION
+    pass
```
becky-gilbert, line 35: @okaycj FYI, it looks like you don't have the experiment location set up in the `STORAGES` object in `project/settings.py`, so if this approach works for static/media then I assume we'll need to add in the experiment location somewhere (I realize you maybe just haven't gotten that far yet!).

PR #1626, opened 2025-05-20
Fix "recipients" dropdown on contact participants page

### LINKED ISSUE #1618: Message recipients not populating correctly on contact page
## Summary

The contact participants feature is not working because of a problem with the recipient selection input element. It looks like something has changed in Django's select multiple form widget, and our select element options are not rendering correctly. 

## Problem

The SelectMultiple form widget expects the choices array to be tuples of `(value, label)`. But we are using a nested structure for our recipient choices so that we can pass some extra information about each participant that we can use in our custom template (e.g. slug, email preferences). Our choices structure looks like this:

```python
[
  (UUID('123'), [
    ('email_next_session', True), 
    ('nickname', 'name1'), 
    ('email_new_studies', True), 
    ('slug', '123-name1'),
    # etc.
  ]),
  (UUID('456'), [
    ('email_next_session', True), 
    ('nickname', 'name2'), 
    ('email_new_studies', True), 
    ('slug', '456-name2'),
    # etc.
  ]),
]
```

The updated version of Django treats this nested structure as a set of _grouped_ choices. If it detects this structure, it lists the first outer tuple element as the group name, and any items in the second element as the options for that group. Here is the current template for the select multiple element (`django/forms/templates/django/forms/widgets/select.html`):

```handlebars
<select name="{{ widget.name }}"{% include "django/forms/widgets/attrs.html" %}>
  {% for group_name, group_choices, group_index in widget.optgroups %}
    {% if group_name %}
      <optgroup label="{{ group_name }}">
    {% endif %}
     {% for option in group_choices %}
        {% include option.template_name with widget=option %}
      {% endfor %}{% if group_name %}
      </optgroup>
    {% endif %}
  {% endfor %}
</select>
```

And it is meant to handle this grouped structure for SelectMultiple choices, where the first element in the outer tuple is the group name (Group 1, Group 2) and the second is the actual choices that should be piped into their own `option` elements:

```python
choices=(
    ('Group 1', (
        ('item1', 'Item 1'),
        ('item2', 'Item 2'),
    )),
    ('Group 2', (
        ('item3', 'Item 3'),
        ('item4', 'Item 4'),
    )),
)
```

becky-gilbert:
Fixes #1618

## Description

This PR fixes a problem with the Contact Participants page "recipients" drop-down choices. The participant choices are rendering incorrectly and it is not possible to select them, which means that researchers cannot send messages. 

The PR restructures the choices and data that we use for the email/message "recipients" field on the Contact Participants page. We used to use a nested tuple for each participant/recipient option, where the first element was the UUID and the second contained all of the other info we needed to create the dropdown option element. However, after our package updates, this nested tuple structure is being treated as a set of _grouped_ choices, where the first element is the group name and the second contains the choices in that group (see #1618). Because of this rendering problem and the way the page is set up, it is not possible to select any participants for the "recipients" field.

## Solution

I think the "right" way to do this (i.e. the way that was intended by Django devs) is to:
1. Move the dynamic injection of the recipient field `choices` out of `get_context_data` and into `get_form`.
2. In `get_form`, set the recipient `choices` (previously done in `get_context_data`) and pass in a mapping of each choice (participant) and associated HTML data attributes.
3. Customize `create_option` in our custom `EmailRecipientSelectMultiple` widget class so that it takes the choice-attributes dictionary passed in through the form class, and attaches the appropriate set of data attributes to each `option` object that is returned from `create_option`. 
4. In the option template, we can now access the data attributes in `option.attrs`.

Other changes:
- Fixed failing test. 
   - The test was trying to get the choices from the context, but they're no longer there so I updated this to get them from the form itself. 
   - I moved the `participant_slug` method out of the `StudyParticipantContactView` class so that I could import it and use it in tests. 
   - I modified the `participant_slug `method to take a User object instead of a dict.
   - The changes to `participant_slug` meant that I also had to change how this method is called within the form view (previously in `get_context_data`, now in `get_form`).

## Manual tests

So far I've manually tested the following:
- Recipients options render correctly.
- Recipients option is disabled when the data attribute is absent for the selected email filter.
- Choice disabled/enabled status is updated when a new email type is selected.
- None of the recipient options are disabled when the 'transaction' email type is selected.
- Can email a participant if they allow the selected email type.
- Cannot select a participant as a recipient if they do not allow the selected email type.
- Link from Individual Responses page still works (automatically adds the participant in the recipients field).
- Link from Individual Responses page works when the participant does not allow the default email type (warning banner message is shown).

## Unit tests

I still need to add tests to cover this choice/data attributes rendering issue. I can do that in this PR or a separate one.

### DIFF
--- a/exp/views/contact.py
+++ b/exp/views/contact.py
@@ -58,27 +58,6 @@ class StudyParticipantContactView(
     def get_context_data(self, **kwargs):
         """Gets the required data for emailing participants."""
         ctx = super().get_context_data(**kwargs)
-        study = ctx["study"]
-        participants = (
-            study.participants.order_by(
-                "-email_next_session"
-            )  # Just to get the grouping in the right order
-            .all()
-            .values(
-                "email_next_session",
-                "uuid",
-                "username",
-                "nickname",
-                "password",
-                "email_new_studies",
-                "email_study_updates",
-                "email_response_questions",
-            )
-        )
-        for par in participants:
-            par["hashed_id"] = self.participant_hash(par)
-            par["slug"] = self.participant_slug(par)
-        ctx["participants"] = participants
 
         previous_messages = (
             Message.objects.filter(related_study=self.object)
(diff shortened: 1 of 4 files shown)


exp/views/contact.py
```diff
@@ -39,16 +47,6 @@ def can_contact_participants(self):
...
-        return hash_id(participant["uuid"], self.object.uuid, self.object.salt)
-
-    def participant_slug(self, participant):
-        return (
-            self.participant_hash(participant)
-            + "-"
-            + slugify(participant["nickname"] or "anonymous")
-        )
-
     def slug_from_user_object(self, user):
```
becky-gilbert, line 50: @okaycj I didn't realize this until pushing my changes, but I think this `slug_from_user_object` is doing the same thing as `participant_hash`/`participant_slug` now that I've modified those to take a User object instead of a dict. We should probably just use one of these methods to avoid confusion. I don't have a preference. Do you?

becky-gilbert: I just cleaned this up. I replaced the use of `slug_from_user_object` with `participant_hash`.

becky-gilbert:
I cleaned up the participant hash/slug method for the contact participants view, but there are other places in the code that also use participant slugs and these should be cleaned up as well. I created a separate issue for this: #1633

PR #1641, opened 2025-05-23
Fix url extra slash

### LINKED ISSUE #1639: Extra slash in some URLs
## Summary

We recently added a trailing slash to our BASE_URL to fix issues with file loading from GCP, following some changes to how the GCP storage settings configured. Now we are generating broken URLs in a few places due to this additional slash after the domain. 

- All URLs added to emails, including study URLs (announcement emails) and unsubscribe links.
- URLs generated for experimenter study pages e.g. (embedded in `DISCOVERABILITY_HELP_TEXT`, `default_exit_url`)

## Dev environments

I think devs need to add trailing slashes to the `BASE_URL` and `EXPERIMENT_BASE_URL` in local .env files to match remote environments. I will test this and update `env_dist`.

becky-gilbert:
Fixes #1639

## Description

This PR fixes some broken links that were caused by an additional slash that occurs in some places between the base URL (e.g. childrenhelpingscience.com) and the rest of the URL. This problem was caused by adding a trailing slash to the `BASE_URL` in the project settings, which was needed to fix some issues with the GCP static/media/experiment URLs.

## Solution

I left the trailing slash in the `BASE_URL` value and added a helper function to remove it where necessary in other places in the code. For instance, the trailing slash is removed from the `base_url` value that is used in email context, because that value is used in email to dynamically generate full URLs by concatenating the base URL with the path from a `url` template tag, e.g. `{{ base_url }}{% url ... %}`. 

## List of changes

- remove trailing slash from base URL in the following places:
   - `send_announcement_email` context
   - `send_as_email` and `send_mail` context (this covers all other email types)
   - `email_headers` (generates the unsubscribe link)
   - link to study listing in 'discoverability help text' (researcher study details page)
   - default exit URL used when creating a new study
   - EFP redirection link for study preview/participate button in local/dev environments
 - add/update tests:
    - study announcement email should contain valid URLs
    - study announcement email URLs should not contain double slashes (after the https:// part)
    - basic tests for `email_headers` method in `Message` class, including URL double slash validation
    - test case for the new `get_url_without_trailing_slash` helper function


okaycj (approved):
Looks like we're moving in the right direction.

exp/tests/test_email_templates.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class TestEmailTemplates(TestCase):
+    def test_notify_researcher_of_lab_permissions_html_lab_url(self):
+        context = {"lab_id": 4321}
+        html = render_to_string(
+            "emails/notify_researcher_of_lab_permissions.html", context
+        )
+        bs = BeautifulSoup(html, "html.parser")
+        self.assertEqual(bs.a["href"], "https://localhost:8000/exp/labs/4321/")
```
becky-gilbert, line 20: Is it a problem to hard code the BASE_URL value here (https://localhost:8000/)? Do we know whether this will work in other test environments? I wonder if we need to import this from project settings, or maybe set it explicitly the way you did for `TestGetAbsoluteURLTestCase` (`@override_settings(BASE_URL="https://fakedomain.asdf/")`).

becky-gilbert (commented):
@okaycj I'm not allowed to approve my own pull request, but this is amazing. SO MUCH BETTER than what I had before! 🌟 🏆 💯 
I left you one tiny question in the tests. 
I've reviewed everything but still need to do some manual testing. I can do that locally and let you know when I'm done. Or I can test this on staging, if that would help to keep things moving forward.

becky-gilbert:
I've just tested the URLs in all of these places locally and am happy that everything is working.
- send_announcement_email
- send_as_email and send_mail 
- email_headers (unsubscribe link)
- link to study listing in 'discoverability help text' (researcher study details page)
- default exit URL used when creating a new study
- EFP redirection link for study preview/participate button in local/dev environments

exp/tests/test_email_templates.py
okaycj: That's valid. Gimme 30 mins and I'll push an update.

PR #1644, opened 2025-05-29
Fix pipe webhook

### LINKED ISSUE #1640: Pipe webhook is failing
## Summary

The pipe webhook is returning 500 errors. This is why a number of researchers have found that they are consistently not seeing consent video files. The first 500 response was on 2025-05-16 15:41:12 (UTC-04:00) and it is still occurring (the most recent I can see is 2025-05-23 14:43:05 (UTC-04:00)). 

The good news is that it looks like the files were saved to S3 successfully, and the problem is just that we need to get the information into the database. Once this is resolved, we will need to import data for 250+ videos.

This only affects the older EFP (pipe) versions - any issues that researchers might be having on newer (RecordRTC) versions are unrelated to this.

becky-gilbert:
Fixes #1640

This PR fixes an issue with the pipe webhook returning 500 errors. 

## Changes

In the `RenameVideoView` API endpoint (`exp/views/video.py`) that we provide to Pipe:

- Get the POST payload from `request.POST` instead of `request.body`, since the former provides the data as a dictionary so no further parsing/conversion is needed.
- Remove the bytes to ints conversion, the use of `urlib.parse.parse_qs`, and the `[0]` indexing on the result because these no longer seem necessary
- 
  ```python
   rawbody = list(request.body)  # bytes to ints
   # rawbody = [45 if i==59 else i for i in rawbody] # semicolons -> dashes
   postbody = urllib.parse.parse_qs(
       bytes(rawbody).decode("utf-8")
   )  # NOW parse the request string
   ...
   message = thisURL + postbody["payload"][0]
  ```

In the `Video` class method `from_pipe_payload` (`studies/models.py`):

- Instead of specifying a UTC timezone via `timezone.utc`(from `django.utils`), we need to use `datetime.timezone.utc` from `datetime`. The django utils version has been [deprecated](https://github.com/django/django/commit/59ab3fd0e9e606d7f0f7ca26609c06ee679ece97). This fixes the "module 'django.utils.timezone' has no attribute 'utc'" error seen in this [Sentry issue](https://massachusetts-institute-of-technology.sentry.io/issues/6610561367/events/1e92de51872841dc8a04f9c88a651723/).

## To do

- Test on staging. It works locally, but there might be other issues, e.g. with the slash at the end of the `BASE_URL` and the way that we build the URL with this line: `thisURL = request.scheme + "://" + request.headers["host"] + request.path`. The URL is part of the data that is hashed for an authentication signature, so it needs to match the webhook URL that we give to Pipe.
- Add tests (we are skipping tests for this view - see [here](https://github.com/lookit/lookit-api/blob/b0c3a3b6de63502e8998b80d6b87fe6484c23cee/exp/tests/test_video_views.py#L49-L59))
- After fixing the endpoint, we can re-run all of the failed Pipe webhook calls (either via a script or individually re-running each one from the Pipe GUI).

### DIFF
--- a/exp/views/video.py
+++ b/exp/views/video.py
@@ -1,8 +1,7 @@
-import ast
 import base64
 import hashlib
 import hmac
-import urllib.parse
+import json
 
 from botocore.exceptions import ClientError
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


okaycj (approved):
💯

PR #1647, opened 2025-05-29
Fix `timezone.utc` error in pipe webhook and add tests

becky-gilbert:
This PR is a follow up to #1640 and #1644. It does the following:

- Fixes a `timezone.utc` error caused by naming conflicts in imports. This _should_ fix the Pipe webhook 500 response on staging.
- Moves some of the code in `RenameVideoView` into a `try`/`catch` to handle missing attributes / key errors in the payload data.
- Adds tests for the `RenameVideoView` (pipe webhook endpoint).

### DIFF
--- a/exp/tests/test_video_views.py
+++ b/exp/tests/test_video_views.py
@@ -1,7 +1,14 @@
+import base64
+import hashlib
+import hmac
+import json
 import logging
-import urllib.parse
-from unittest import skip
+import math
+import random
+import time
+from unittest.mock import MagicMock, patch
 
+from django.conf import settings
 from django.test import TestCase
 from django.urls import reverse
 from django_dynamic_fixture import G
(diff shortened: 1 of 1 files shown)


studies/models.py
```diff
@@ -1470,9 +1470,7 @@ def from_pipe_payload(cls, pipe_response_dict: dict):
         return cls.objects.create(
             pipe_name=old_pipe_name,
             pipe_numeric_id=data["id"],
-            s3_timestamp=datetime.fromtimestamp(
-                int(timestamp) / 1000, tz=datetime.timezone.utc
-            ),
+            s3_timestamp=datetime.fromtimestamp(int(timestamp) / 1000, tz=timezone.utc),
```
becky-gilbert, line 1473: I think this is the reason that the webhook is still failing on the staging site.

exp/views/video.py
becky-gilbert, line 1: The diff here makes it hard to see what's changed. I haven't made any substantial changes to how this works, I just added a `try`/`catch` to catch key / attribute errors and respond with `HttpResponseBadRequest`.

exp/views/video.py
```diff
@@ -28,43 +33,52 @@ def post(self, request):
...
-        # Compare to the one in the header
-        signatureSent = bytes(request.headers["x-pipe-signature"], "UTF-8")
-        authenticated = signatureComputed == signatureSent
 
-        # Convert data from string representation of dict to dict
-        payload_data = json.loads(payload)
+        try:
+            # Authenticate the webhook (see https://addpipe.com/docs#authenticating-webhooks)
+            # Append the JSON POST data received via webhook to the URL string.
+            thisURL = request.scheme + "://" + request.headers["host"] + request.path
```
becky-gilbert, line 41: @okaycj should this URL construction be updated to use the methods you're using in #1641? If so we can merge that first and I'll update it here.

okaycj (approved):
Nice work! This at least resolves the issue that sentry found.  Fingers crossed that this is "one and done".

exp/views/video.py
okaycj: Maybe you could use a simpler version of [this](https://stackoverflow.com/a/15799706).

PR #1649, opened 2025-05-30
Convert study image filenames to lowercase

### LINKED ISSUE #1646: Study images with uppercase don't load
## Summary

There are study images in the database with uppercase letters, but the filenames were converted to lowercase in GCP, so now these images are not loading. This issue has been resolved for newly-uploaded images going forward, so it's just a problem for a subset of previously-uploaded images.

## Possible solutions

- Run a data migration to convert study images to lowercase. 
- Use the 'lower' filter to change filenames to lowercase in templates.

becky-gilbert:
Fixes #1646

This PR fixes a problem with some study images not loading from GCP. The files that are failing to load have uppercase characters in the database, and were converted to lowercase in GCP storage. This addresses the problem by converting the filenames to lowercase in the database so that they match GCP. (We are already doing this with newly uploaded files going forward, so this fix is just to address previously-uploaded files).

I've tested this locally, but it would help to get more testing from other devs. 

Also, there is no way to roll this back unless we store the existing filenames somewhere in the database (e.g. in a new column in the studies_study table).


okaycj (changes requested):

okaycj (approved):
Looks good.

PR #1654, opened 2025-06-04
Add pipe video processing command

becky-gilbert:
This PR adds a management command to the 'studies' app. This script only runs when called manually via `manage.py`, and it must be given a CSV file that contains the exported Pipe recordings for processing. The CSV should be located at the project root and deleted after processing.

This can be used to add Pipe videos to the database while bypassing the AWS S3 actions (renaming the video with the Pipe name to our custom name). This is useful for adding Pipe videos to the database when the Pipe webhook processing failed _after_ the S3 naming step (which means that re-running the webhook no longer works because the file with the original Pipe name no longer exists in S3).

## How to test locally

Here's how to run the command from docker:
```
docker-compose exec web uv run ./manage.py add_pipe_video_objects_without_S3 develop_test_payloads.csv
```

With dry-run flag (no video object creation):
```
docker-compose exec web uv run ./manage.py add_pipe_video_objects_without_S3 develop_test_payloads.csv --dry-run
```

The result will depend on the contents of your local database vs the contents of the CSV file. For each row in the CSV file:
- If the video does not exist in your database, but the study/response UUIDs do exist, then it will create the new video object.
- If the video does exist in your database (matching full_name), then it will log and skip that row.
- If any of the required information is missing from the row, then it will log and skip that row.
- If the `Video.from_pipe_payload` call throws an error, e.g. because the database does not contain the study or response UUID, it will skip the video and provide a 'failed to process' log.

You can test the video creation process by finding a video that was skipped because it already exists in the database, delete the row, and re-run the command. A new video object should be created that matches the original (except for new Video object ID/UUID values).

Example logs:

Created video: videoStream_...mp4
Skipped existing video: videoStream_...mp4
Skipped existing video: videoStream_....mp4
Study with uuid 1e9157cd-b898-4098-9429-a599720d0c0a does not exist. Study matching query does not exist.
Failed to process row 4270581: Study matching query does not exist.
Study with uuid 1e9157cd-b898-4098-9429-a599720d0c0a does not exist. Study matching query does not exist.
Failed to process row 4270598: Study matching query does not exist.


okaycj (approved):
Nice.  The script looks good and the local test works as expected.

PR #1675, opened 2025-06-17
Batch video download

okaycj:
# Summary

With the new disk installed, batch downloading of video files has been re-enabled. Additionally, we've added logging to aid in future debugging if needed.


becky-gilbert (approved):
Great!

Issue #1719, opened 2025-08-01
Study history page: update invalid consent text


becky-gilbert:
## Summary

On the participant-facing study history page, families can see whether the consent videos from their previous sessions have been approved, rejected, or are still pending. When a video has been rejected, the message currently is:

> Consent status: Invalid. There was a technical problem with your consent video, or it did not show you reading the consent statement out loud. Your other data from this session will not be viewed or used by the study researchers.

We should consider the family experience when they have completed a study and the consent has been rejected, but there was nothing obviously wrong with the consent video. This "invalid" message can be confusing to families who have a seemingly-valid consent video, because it fails to capture other reasons why their consent might have been rejected, e.g.
- The researcher determined that they were not eligible (e.g. child not present or clearly out of age range, using a mobile phone/tablet)
- The researcher approved the consent but then realized that there was something wrong with the data, and changed the consent status to "rejected" so that they no longer have access to data that they do not plan to use.

Another thing to keep in mind is that, for the "not eligible" case, there are a number of reasons why the family might've done the study:
- Bad faith: ignored or did not read the eligibility criteria that was clearly stated, hoping to get compensated but probably not surprised if they are rejected
- Good faith: missed the eligibility criteria that would exclude them and are expecting to be compensated
- For fun: they know they are ineligible and simply wanted to do the study anyway - not expecting compensation

Ideally we would have language that is appropriate for all cases.

## Proposed solutions

Here are a number of different possible solutions. These are not all mutually exclusive.

1. Modify the "invalid" statement to include ineligibility:

> There was a technical problem with your consent video, it did not show you reading the consent statement out loud, or **the researcher determined that you did not meet the eligibility criteria**. Your other data from this session will not be viewed or used by the study researchers

2. Change the "Invalid" consent status name to something a little nicer and more general, e.g. "Data not used". (This might also mean changing the "Consent status" label to "Session status" or just "Status").

3. Keep the "Invalid" consent status label/text, but only use it for videos that are actually invalid for the reasons stated, and create a new status label for "Data not used"/"Ineligible".

4. On the consent manager page, add an easy way for researchers to communicate consent video issues to families. E.g. a feedback textbox or checkboxes for common issues. This info would then show up on the study history page so that families can see why a given consent video was rejected.

## Questions

For this case:

> The researcher approved the consent but then realized that there was something wrong with the data, and changed the consent status to "rejected" so that they no longer have access to data that they do not plan to use.

[...]

PR #1723, opened 2025-08-06
Update invalid consent text

becky-gilbert:
Partially addresses #1719 (but shouldn't close that issue).

This PR makes a minor change to the text that is shown on the family-facing study history page when the consent is "invalid".

Because the text is translated, this also updates the translation PO files with the new invalid message, which will need to be translated into other languages. (The other changes to these PO files were due to the files being out-of-date).

Finally, this PR removes the duplicate "Date" (date created) on the study history page.

## Screenshot 

Before/current:

Screenshot 2025-08-06 at 10 21 55 AM

After:

Screenshot 2025-08-06 at 10 15 01 AM

### DIFF
--- a/web/templates/web/studies-history.html
+++ b/web/templates/web/studies-history.html
@@ -105,10 +105,6 @@
                                 <strong>{% trans "Date" %}:</strong>
                                 {{ response.date_created }}
                             </div>
-                            <div>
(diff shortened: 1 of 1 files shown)


bleonar5 (approved):
Looks reasonable to me!

Issue #1727, opened 2025-08-08
High memory usage with download all videos


becky-gilbert:
## TL;DR

We're still having a problem with large 'download all video' jobs potentially disrupting the site. This time it's a RAM issue (rather than disk space, as before). 

## Description

Triggering a 'download all videos' task on a study with a very large number of videos (14k) continuously eats up memory and can cause pod evictions (site downtime). What's worse is that the task will remain in the task queue when the pods restart, so it will be triggered again, causing a sawtooth pattern of linearly-increasing memory usage as the pod continues to consume memory and then crash.

<img width="433" height="297" alt="Image" src="https://github.com/user-attachments/assets/2a2709b2-8b8c-4feb-aec3-59f9c7a35f50" />

## Possible causes

I think the problem is related to multiple caching layers and buffering:
- Python's zip file buffering: the video data is buffered while being compressed
- Overhead from GCS FUSE layer: even though we're using the streaming option with small chunks, FUSE is doing its own read/write buffering and caching in the pod’s RAM
- Zipping into a file on FUSE: each zf.write() is going through multiple buffering layers (Python -> zipfile -> zlib -> FUSE cache -> GCS). These layers could be reduced by either zipping into non-FUSE local storage or streaming into a GCS zip.

## Possible solutions

- [x] Remove the file compression. The video files are already compressed so this probably doesn't do much to reduce the file size anyway. #1728
- [x] Stream the zip file creation directly to GCS (no zip or temp files in local memory). This would be a bigger change to the task and would require the use of a zip streaming library. #1737 #1739 #1740 #1742 

And separate to the above:
- [x] Limit the resources available on the builder to 50% of RAM (not an actual solution to the problem, but this will hopefully stop any large video download jobs from taking down the web pod). https://github.com/lookit/lookit-orchestrator/pull/61

We've decided against these options for now, but could come back to these if the issue persists:
- Tweak GCS FUSE settings (`buffer-size` and maybe others?)
- Switch from GCS FUSE to ephemeral SSD for `/code/scratch` and upload the final zip to GCS. This avoids the double buffering of FUSE while writing the zip file and then moving it to GCS.

The two items above are not necessary, since we're now streaming directly to a zip in GCS and therefore aren't using the GCS FUSE mount or an SSD for zip storage.

- Change the retry settings for the celery task so that any large jobs that bring down the pods do not start up again after pod restarts.

[...]

PR #1728, opened 2025-08-08
Download all videos: remove compression to reduce RAM usage

becky-gilbert:
This is one attempt at addressing #1727 (but should not close the issue).

This removes the file compression step from the download all videos task. This should reduce some of the task's memory usage, with the trade off that the resulting downloadable file will be a little larger (but not much, because the videos are already compressed). We will need to test this version to see if it actually makes much difference to the task's memory profile.


becky-gilbert:
The Sentry issue should've been fixed when I changed a permission setting on the GCS bucket. I'll leave it open until I'm sure it's been handled.

I'm going to merge this without review so that I can do a quick test on staging.

PR #1737, opened 2025-08-19
Use stream_zip for 'download all videos' task

becky-gilbert:
This PR is an attempt to help with #1727 (but should not automatically close that issue).

This is to see whether we can use `stream_zip` to stream files directly into a zip on GCS, rather than storing a temp file. Even though our temp file is stored in the GCS fuse mount rather than in the container, this solution might help prevent the constant RAM usage increase by preventing layered caching from the fuse mount.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "requests==2.32.4",
     "sendgrid-django==4.2.0",
     "sentry-sdk==2.27.0",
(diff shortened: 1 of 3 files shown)


becky-gilbert:
I'm going to merge this for testing, but will request reviews if it works on staging.

Also, the issue that sentry flagged has been fixed (it was on the GCS side rather than in the code).

PR #1831, opened 2026-02-05
Add max responses field to study and auto-pause

### LINKED ISSUE #1829: SCOPING: cap number of participants, condition counterbalancing, cap by criteria
## Summary

This is a multi-phase plan to add a related set of researcher usability features:
- Cap the total number of responses. The study should pause automatically when it has this number of responses.
- Condition counterbalancing. Allow researchers to define a number of conditions, and we counterbalance assignment to those conditions. Researchers have access to the condition ID in the experiment.
- Cap the number of responses within conditions. Keep a running total of valid/complete responses within each condition and stop assigning a condition when the cap is reached.
- Cap the number of responses within a given criteria expression (e.g. age range).

These features are all related and so it would be more efficient to get them done all together. However, I'm grouping them into releaseable phases, so that we can decide at each phase whether to continue on to the next phase, release and continue, or release and pivot to other priorities.

## Capping total number of participants

Related issue: #772

### Question: Which responses count toward the total? 

We need to decide how we will consider a response to be countable toward the study's total response count. There are a few straightforward ways to do this, but there are some issues with each of these.
- Participant must be **eligible**
  - Problem: When a study specifies that the child must not have participated in that same study, valid responses can be labelled ineligible because the participant _started_ the study (even just the first trial) and then went back to do the whole thing. The result is that our response count might underestimate the number of valid responses because it incorrectly treats these responses as ineligible.
- Session must be **complete**
  - Problem: Sometimes participants get all the way through the study but don't do the exit survey, in which case they are marked as 'incomplete' but their data is still useable. The result is that our response count might underestimate the number of valid responses because it incorrectly treats these responses as incomplete.
  - Problem: In hybrid studies, participants do the consent and then are redirected to an external study. These studies are treated as 'internal' in our system, but the study flow usually does not include an exit frame, which means the responses are always marked as 'incomplete'. The result is that our response count might underestimate the number of valid responses.
  - Problem: For external studies, we do not have any information about whether the participant completed the study or not.
- Must be marked as **complete** by the researcher
  - Problem: This requires the researcher to continuously monitor the responses and actively mark each one before it is be counted towards the study total. If the researcher doesn't do this (e.g. because a lot of responses come in over an evening/weekend) then the study may collect additional responses beyond the cap. With this requirement, our response count would underestimate the number of valid responses at any given time.

### Possible solutions

@mekline please let me know what you think about the question above and the solutions outlined here

- Session must not be a preview
- Participant must be eligible
- For internal studies, response must also be complete
- Responses that meet the criteria above are automatically counted toward the study's total, BUT the researcher can edit the include/exclude boolean associated with each response. This way, we can keep an automated running count of the response count without researcher input, but the researcher can correct this count as necessary.
- Responses for external studies are always treated as complete unless changed by researcher

With this solution, there is still a problem with hybrid studies because these are 'internal' but will always produce incomplete responses. Possible solutions:
- Add a 'hybrid' flag (boolean, defaults to false) to internal studies so that we know to ignore the response's completion status.
- Require researchers to include an exit frame in hybrid studies. (This would cause confusion since the exit survey asks for feedback on a study that the family hasn't done yet, so it would probably require new parameters for modifying the exit survey).

For external studies (and hybrid studies, if we add a flag for them), we will need to consider all responses as 'complete' until the researcher changes the status. We could address this limitation by adding an API endpoint that the researcher puts into the end of their external study, which sends the session ID back to our system with a completion code. This would prevent these studies from automatically pausing prematurely because incomplete responses have been counted toward its current response total.

### To do

- [x] In the study model/table, add a field for the response limit. The default value is empty/null, in which case there is no limit (same as status quo). #1831
- [x] Add the response limit field to the study create/edit form. This value can be edited at any time, and can be changed without causing the study to be rejected (if already approved). #1831
- [x] When a response comes in for that study and is marked as complete, check the study's response count and pause the study if the limit has been reached. #1831
- [x] If the study is active and the researcher's edits cause the response count to reach/exceed the limit, then the site should:
  - Show a warning/pop-up stating that the response limit has been reached and the study will now be paused #1832
  - Automatically pause the study #1832
- [x] If the study is inactive and the researcher tries to start it, but the response count has reached/exceeded the limit, then the site should:
  - Show a warning/pop-up stating that the response limit has been reached, and "please modify the response limit if you want to collect more responses" #1832
  - Do not allow the study to start #1832
- [x] Whenever the study changes from active to paused due to reaching the response limit, a notification email should be sent to the researcher. #1832

UI updates for studies with a response limit:
- [x] Show the current valid response count and limit on the study details page #1831
- [x] In the Individual Responses page, flag each response in the left-hand column of the table as one of: Preview, Valid/Accepted, or Invalid/Rejected/Disregard #1846
- [x] In the Individual Responses page, indicate the number of responses associated with each type.
- [x] For the latter two categories, allow the researcher to change the status (~~click an icon or checkbox~~ via button and pop-up confirmation box). This action should update all counts on the page.
- [x] Allow researchers to filter the table by response status (valid/invalid/preview). #1846

@mekline please let me know what you think about the above specs

## Balancing condition assignment

Related issue: #677

- In the study create/edit form, add field called 'add conditions for counterbalancing'. This starts off empty/null. The researcher can add a value indicating the number of conditions that they want counterbalanced assignment for. A list will appear with one ID (number or letter?) for each condition, along with a text box where the researcher can enter their own corresponding condition label.
- Each response is assigned one of these conditions, and the condition value is passed into the experiment runner (just jsPsych to start with?)
- As a new response is created, the site will assign a condition by cycling through the condition list (either in order or randomly without replacement).
- In the experiment code, the researcher is responsible for checking the condition assignment for that session (e.g. `const condition = chsData.getCondition()`) and modifying the experiment as necessary for that condition.

The approach above is a reasonable solution for counterbalanced assignment and would likely produce similar response counts per condition. But this just counterbalances at the start of the session, and it does not take into account the true condition counts that are based on the completed/valid responses.

## Tracking/capping response counts based on criteria/conditions

Related issue: #772

- In the study create/edit form, add field called 'add a response limit/target based on criteria or conditions'. This starts off empty. The researcher clicks to add a new condition and/or criteria expression (e.g. age range), along with a response limit.
- If the researcher adds a single criteria-based response limit, then the study's total response limit will automatically set to the same number. If they add additional criteria-based limits, then the study total response limit will be the sum of all criteria-based limits.
- The researcher can edit the condition/criteria/limit values at any point, add new ones, or delete existing ones.
- As a new completed response comes in for the study, the site checks to see if any of the condition and/or criteria-based response limits have been reached. If so, that condition will not be assigned and/or any child who meets this criteria expression becomes ineligible.
- Whenever the site needs to determine a child's eligibility (e.g. on the participation page, when sending announcement emails, on the study list page), it will take any criteria groups/limits into account.
- One a participant clicks 'participate now' and a new response is created, the site will look at the set of conditions (that this child is eligible for) and associated counts, and assign the condition with the smallest number of valid responses. If there are multiple conditions with the smallest value, it will select at random.

Questions:
- What if the criteria expressions overlap, in which case a single response might be counted toward both totals? And if the response limit is reached for one criteria expression, what happens if a participant meets that criteria and another one that is still under the limit? 
- The implementation of dynamic study eligibility might be tricky. For instance, if a study has some kind of global criteria expression (e.g. age 4-6), as well as limits on other subsets of that criteria (age 4-4.5), and the limit is reached for the subset, then the negation of the subset eligibility would have to be appended to the global eligibility, e.g. "is 4-6 years old" AND  "is NOT 4-4.5 years old". This will probably work but needs clear documentation, as it could cause unexpected behavior if, for instance, a researcher includes global criteria in a subset (e.g. "age 4-4.5 AND speaks English", in which case all English-speaking children would be excluded when the 4-4.5 age range response limit is reached).

becky-gilbert:
Addresses part of #1829 but should not close the issue.

This PR does the following:

- Changes to Study model: 
  - Adds an optional `max_responses` field. Default is None.
  - Adds the helper property `valid_response_count`. For all study types, a valid response must:
    - not be a preview
    - be eligible
  And for internal studies, the response must also:
    - have a completed consent frame
    - be complete
    - not have rejected consent (only pending or accepted are counted)
  - Adds the helper property `has_reached_max_responses`. This returns `True` if the valid response count has reached/exceeded the limit, otherwise it returns `False`. If the study does not have are response limit, then this returns `False`.
  - Adds the stub method `check_and_pause_if_at_max_responses` to the Study model, which is called when a response object is saved/updated and the response is completed.

- Changes to Study create/edit form - see screenshots below: 
  - Adds a "Set a response limit" checkbox question, which is unchecked by default. 
  - Adds the `max_responses` field to the study create/edit form. This field is disabled by default, and becomes enabled if the "Set a response limit" checkbox is checked. Default is None.
  - If the study has a max_responses value, then the valid response count / limit is shown on the study details page, along with a progress bar (blue: up to 75%, yellow: 75-99%, green: >=100%). If the limit has been reached/exceeded, then "Response limit reached" appears below the progress bar.

- Adds tests for the new Study model properties `valid_response_count` and `has_reached_max_responses`.

To do (in separate PRs):
- Implement the `check_and_pause_if_at_max_responses` method.
- For external studies, call `check_and_pause_if_at_max_responses` when the response is first created.
- Changes to individual responses table

### Screenshots for UI updates

"Set a response limit" checkbox (default: unchecked) and "Maximum responses" text box (default: none) on study create/edit pages, below the participant eligibility criteria section.

response_limit_unticked

response_limit_ticked

Valid response counts, response limits, and progress bars on the study details page.

resp_count_0

resp_count_1

resp_count_2

resp_count_3

resp_count_4

resp_count_5


becky-gilbert:
> Text on the max responses field - add a link to the (new) documentation page about how this works? Potentially also add note on extra participants, e.g.
"The study will automatically pause when the number of valid responses reaches this limit. Note that participant sessions running when the limit is reached are permitted to continue. See [the documentation] for more information on response limits and tracking valid responses."

Yes, sounds great. I'll write up a draft of this in the docs so that I have a link to work with.

> Not sure how important it is to distinguish 5/5 vs. 6/5 on the green bar, but it sort of feels weird to have no UI cue to that - ideas?

Agree, I'll think about it. Would a different color be too much? And/or I could update the text underneath the progress bar to "Response limit exceeded".

> Oh! Will the study automatically un pause if you un-validate a few responses, putting you back under the (overall or condition) limit? Probably it shouldn't, but make sure to model that user journey and make sure that that user doesn't feel stymied when they create some open slots and then want to re-open their study.

I wasn't planning on automatically re-starting the study - just automatically pausing. 

If the study is paused and then drops below the response limit because the researcher has un-validated some responses, then we could show a banner message at the top of the page (e.g. "The valid response count is below the limit for this study. Re-start the study if you want to collect more responses."). But I don't think we should automatically re-start the study for data collection without some kind of action by the researcher.

In the opposite case - if the study reaches the limit because the researcher marks some previously invalid responses as valid - then this WILL automatically pause the study, with a banner message at the top of the page (e.g. "Your study has reached the valid response limit and has been paused. If you want to collect more responses, increase the limit for this study, and then re-start it.").

Let me know if this sounds ok to you and if you have suggestions for wording. And we should definitely walk through these user stories to make sure everything feels clear and predictable.

studies/models.py
```diff
@@ -548,6 +554,68 @@ def videos_for_consented_responses(self):
...
+        A response is counted as valid if:
+        - is_preview is False
+        - eligibility is "Eligible" or blank/empty (backwards compatibility)
+
+        And for internal studies, responses must also meet the following conditions:
+        - completed is True
+        - completed_consent_frame is True
+        - the consent has not been rejected (must be either pending or accepted)
+
+        For external studies, the completed, completed_consent_frame, and consent requirements are ignored.
```
becky-gilbert: @mekline FYI, I've updated the valid response count so that, for internal studies, the response must have a consent frame, and consent must not be rejected.

PR #1832, opened 2026-02-11
Pause study when max responses limit is reached

becky-gilbert:
This addresses part of #1829 but should not close the issue.

This PR adds logic to the Study model's `check_and_pause_if_at_max_responses` method. This method checks if the study:
- has a max responses value
- has reached its max valid responses value

If the study is active, it automatically pauses the study. 

Depending on the parameters passed, it will also: (1) email a pause notification to study researchers, and/or (2) display a banner message (message content differs depending on whether or not the study is being paused).

The pause transition triggers a notification email sent to site admins. This callback had to be edited to allow the caller (a user) to be an optional argument, since we've now added the ability for study states to change without a user/caller.

The `check_and_pause_if_at_max_responses` method is triggered by the following events: 
  - when a completed response is received for an internal study
  - when a response is created for an external study
  - when the max responses value is updated for a study
  - when a researcher tries to change the study state to active

This PR also adds tests for the `check_and_pause_if_at_max_responses` method, updates to the `StudyUpdateView`, and the max responses check on the activate/start study state transition.

To do:
- The `check_and_pause_if_at_max_responses` method should also be called when a response's validation status is manually changed on the individual responses page. I will put this in a separate PR that also implements the valid/invalid status flag and other changes to the individual responses table.

## Screenshots

When a user edits the max responses value on the Study Edit form, the site checks to see if the valid response count is equal to or exceeds the new max responses value. If so, it displays a banner message. If the study is active, it will be automatically paused.

https://github.com/user-attachments/assets/82c77011-419f-4b1e-ab5a-b05cd8125fcb

https://github.com/user-attachments/assets/148e8d8c-00fd-4798-b451-21c57a49602c

When the valid response count reaches/exceeds the limit, the study is automatically paused and the researchers receive an email.

external_study_paused

external_study_pause_email

If a study has reached its response limit, researchers cannot start/activate the study. They will see a warning message saying that the response limit has been reached, and the study will remain in the same state.

https://github.com/user-attachments/assets/21c574e1-9608-41a5-b18a-22c9e1dbf68e

### DIFF
--- a/scss/base.scss
+++ b/scss/base.scss
@@ -12,6 +12,7 @@
 
 // import study-responses after custom variables because it uses a color variable
 @import "study-responses";
+@import "study-detail-progress-bar";
 
 // Add all bootstrap features
 @import "bootstrap-5.2.0/scss/bootstrap";
--- a/scss/study-detail-progress-bar.scss
+++ b/scss/study-detail-progress-bar.scss
@@ -1,0 +1,54 @@
+progress {
+    &.study-progress {
+        --study-progress-color: var(--bs-info);
+        appearance: none;
+        -webkit-appearance: none;
+        border: none;
+        border-radius: .375rem;
+        height: 20px;
+        width: 100%;
+        overflow: hidden;
+        background-color: #e9ecef;
+    }
+
+    &.study-progress::-webkit-progress-bar {
+        background-color: #e9ecef;
+        border-radius: .375rem;
+    }
+
+    &.study-progress::-webkit-progress-value {
+        background-color: var(--study-progress-color);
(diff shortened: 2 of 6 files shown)


mekline:
Just checking for understanding!!

2) display a banner message (message content differs depending on whether or not the study is being paused).

- So this banner message always shows for a study that has a max limit, e.g. showing "you have 30 of 50 responses" or similar? (I don't actually care if the number shows)

Re: the "todo" - This would be another case where the max can be violated (e.g. user previously said no to some responses, but updates them to yes, suddenly there are 63 out of 60 valid responses)

PR #1846, opened 2026-03-03
UI for valid/invalid status on IR page

### LINKED ISSUE #1829: SCOPING: cap number of participants, condition counterbalancing, cap by criteria

becky-gilbert:
This PR makes the UI updates on the Individual Responses page to allow researchers to change the valid/invalid status of each (non-preview) response. This adds a column to the left side of the table, which shows the valid/invalid status of each response, along with the "P" for preview responses. It also adds the Valid/Invalid/Preview response counts above the table. The Response validity column and response type counts are only shown when the Study has a `max_responses` value.

## Questions

1. Researchers that visit the Individual Responses page can have different permission levels: some can access real participant data (READ_STUDY_RESPONSE_DATA) while others can only see preview data (READ_STUDY_PREVIEW_DATA). Should these Valid/Invalid/Preview counts be the same for all researchers, or should they only relate to the data that each researcher can access (i.e. no valid/invalid counts for researchers with preview-only permissions)? -- **All researchers should see the same counts, regardless of their permissions. But when permissions are restricted, we should note that these numbers include responses that may not be visible/available to the user.**
2. For internal studies, a response can be automatically marked as valid if it has a consent frame and the consent has not been rejected (i.e. consent is either accepted or pending). Should the Valid/Invalid/Preview counts on the Individual Responses page include the Valid/Invalid status of Responses with _pending_ consent? -- **Yes, counts should include those with pending consent. We will need to make this clear on the IR page, or on a different page (and refer to that from the IR page).**


becky-gilbert:
Some additional changes that I'll be making in this PR:
- Add a 'is_valid_researcher_override' field to the response model/table, which stores whether the automatic valid/invalid categorization has been overridden by the researcher. 
- Update the Individual Responses page for external studies to remove irrelevant info
- Add a new 'response status' card to display specific info about the valid/invalid status (all of the relevant fields, based on study type, and researcher override value)
- Add a new 'response count summary' page, which displays more details about the response types and consent status. This will also display conditions and counts, once we introduce that feature.
- Response pages navigation tabs: add the new 'response count summary' tab, and reorder/rename existing tabs: Response Counts, Individual Responses, Download Responses, Download Demographics, Download Videos.

mekline:
Love the pop-up confirmation for researcher changes, note that we probably want it to be different text for the external researchers - e.g. whether it's okay to un-tally a participant who schedules and then no-shows (TBD) should be stated explicitly. 

Should researchers be able to change Tally status always, or should it get 'locked' at some point? I can anticipate future scenarios where we might want that, but I think for now it's fine?

mekline:
Extremely minor - with the return to text links instead of buttons, the red and green icons are reading as very "heavy" to me.  What about colored letters to match the "P" for preview? 

(Rough mockup of the whole thing)

Screenshot 2026-04-24 at 1 13 58 PM

becky-gilbert:
# Updated screenshots

## Internal study with a response limit:

Screenshot 2026-05-27 at 4 35 50 PM
Screenshot 2026-05-27 at 4 36 18 PM
Screenshot 2026-05-27 at 4 36 36 PM
Screenshot 2026-05-27 at 4 38 48 PM


## External study with a response limit:

Screenshot 2026-05-27 at 4 37 22 PM
Screenshot 2026-05-27 at 4 37 39 PM
Screenshot 2026-05-27 at 4 37 59 PM
Screenshot 2026-05-27 at 4 38 20 PM

mekline:
This all looks fantastic! Just one question and a couple of minor text changes: 

Q: You mention that these are screenshots when a max cap is present.  What's the behavior when this has never been set, or has been set and then removed? Is the tab absent or is it present with some kind of placeholder text?  (I might minorly favor the latter to help surface this new feature to people, but no need to delay release if not)

# response count pages

On both of the response count summary pages, let's add some language to help people understand that this is how tallying works *for the current study type*: 

```
Because this is an [internal][external] study, a response is automatically categorized as tallied when...
```
Also, I'd like us to remain strategically silent about eventual no-shows for external synchronous for now, because this is going to get nailed down and will make more sense in context of the payment rollout I think: 

New final paragraph for external response count summary page: 

```
You can override the automatic tallied/untallied categorization on the Individual Responses page. Responses are categorized with the assumption that, when a family clicks the button to participate or schedule a time to participate, they will complete the study. You may need to change tallied responses to untallied when a family clicks this button but does not actually complete the experiment or schedule an appointment. Responses should be tallied if they meet the criteria above and the family participated in good faith. 
```

# Pop-ups

I actually really like how the bulleting gets the Tallied classification across in those paragraphs, so let's use it in the pop up windows too! E.g. 

```
Responses should be tallied if: 
- The correct child participated
- The child was eligible for the study
- The family participated in good faith (not a confirmed scammer)
This is true even if you are not able to use this response.... 
```

In the external version, there's a double "not" in the final list that makes interpretation a little unclear.  How about the following: 

```
This response is currently tallied. 
Are you sure you want to untally it?

Responses should be tallied if: 

- The correct child participated
- The child was eligible for the study
- The family participated in good faith (not a scammer)

This is true even if you are not able to use this response, for instance because the child was fussy or there were technical problems. 

For external studies, you may un-tally a response if, in this response, the family:

- Did not complete the study during this session (asynchronous studies)
- Did not schedule a time to participate (synchronous studies)

```

becky-gilbert:
@mekline 

> Q: You mention that these are screenshots when a max cap is present. What's the behavior when this has never been set, or has been set and then removed? Is the tab absent or is it present with some kind of placeholder text? (I might minorly favor the latter to help surface this new feature to people, but no need to delay release if not)

Right now that page is just absent when there's no max responses set for the study - screenshots at the very bottom. It's very easy to add it, I just need to know what placeholder text to use!

>  there's a double "not" in the final list that makes interpretation a little unclear

Yes! This was bothering me and I couldn't figure out how to fix it. Thanks!

## Updated screenshots

Internal study - response count summary page. Added the "Because this is an internal study,..."

internal_rcs

Internal study - pop up confirmation. Switched to bulleted list.

internal_popup

External study - response count summary page. Added the "Because this is an external study,..." and used your text from the comment above (no more mention of no-shows).

external_rcs

External study - pop up confirmation. Switched to bulleted list and got rid of double-negative wording.

external_popup

Internal and external studies with no max responses value. Currently no "Response count summary" page/placeholder. And in the IR table, no ❌ / ✅  in non-preview rows, and no "Response XXX: Tallied/Untallied" box below the table.

internal_no_max

external_no_max

mekline:
This all looks good! I think this is fine as-is, just wanted to check that setting/altering a study cap will *not* automatically punt a study back to review?

becky-gilbert:
Yes that's right - researchers can modify the response limit (turn it on/off, change the number) anytime they want, and that will NOT trigger an automatic study rejection if the study has been approved.

PR #1863, opened 2026-04-03
Fix large data downloads

### LINKED ISSUE #1599: Response overview download times out
## Summary

On production, it looks like the 'response overview' CSV download can timeout. I assume this is happening because (a) it's a regular GET request (not an offline celery task) and (b) there is a relatively large number of responses for this particular study.

The relevant code is here: https://github.com/lookit/lookit-api/blob/55079df6c302f8e0e53318ec13b98be4f2407809/exp/views/responses.py#L1453-L1476

## Steps to reproduce

- On production, find study 3266 -> Responses: https://childrenhelpingscience.com/exp/studies/3266/responses/all/
- Click the "Data (CSV)" button in the "Response overview" section.

becky-gilbert:
Fixes #1599

We've had a number of problems with data downloads timing out recently. This PR attempts to fix those issues by streaming to/from temp files, rather than concatenating very large strings that are fully stored in memory. 

The temp files created with `tempfile.NamedTemporaryFile` will be deleted automatically. There a few places that use the `delete=False` argument because the data needs to be read back in later - these are cleaned up via a `try`/`finally` block, but we may need to check periodically to make sure that they do not persist if/when a download request fails.

Also, these changes reduce the peak memory usage, but there might still be issues with the response taking too long and hitting the response timeout. So if the problems continue, then we may need to turn some of the large/slow downloads into async tasks.

This PR also adds tests for the PsychDS zip file contents.

### DIFF
--- a/exp/tests/test_response_views.py
+++ b/exp/tests/test_response_views.py
@@ -3,6 +3,7 @@ import datetime
 import io
 import json
 import re
+import zipfile
 
 from django.test import Client, TestCase, override_settings
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


exp/tests/test_response_views.py
```diff
@@ -1373,6 +1380,95 @@ def test_get_appropriate_individual_responses_as_previewer(self):
         # Assumes n_previews fit on one page
         self.assertEqual(n_matches, self.n_previews)
 
+    def _get_psychds_zip(self, query_string=""):
```
becky-gilbert, line 1383: @bleonar5 I was worried about modifying the content in psychds downloads, so I added some tests (starting here) to check that it's still correct. Can you let me know what you think, and if I should add any others? I want to make sure that I'm not missing any edge cases.

PR #1900, opened 2026-07-08
Update and load jsPsych plugins dynamically

### LINKED ISSUE #1484: Load jsPsych plugins dynamically
## Summary

Currently we're loading a static list of plugins in all jsPsych studies (regardless of whether or not the study code actually needs those plugins). It would be better to allow users to select the plugins they need from a checklist, and then load dynamically from that list for each experiment.

## Description

There are currently ~50 ["official" jsPsych plugins](https://www.jspsych.org/latest/plugins/list-of-plugins/) that researchers might want to use in their experiments, and our jsPsych study template loads 5 of these. We also _always_ load all 5 plugins, regardless of whether or not the study needs them. Each plugin that is loaded results in an additional request to the jsPsych package CDN, which increases the study's initial load time (which we've already noticed is slow: https://github.com/lookit/lookit-jspsych/issues/87), so automatically loading all plugins for every study is not a great solution.

Rather than loading the same set of plugins for all studies, we should provide a multiple-selection list on the jsPsych study edit form so that users can select all of the plugins that they need for their study. In the jsPsych study template, we could loop through the plugin list for that particular study and constructs the `script` elements accordingly. The CDN source URLs follow a predictable pattern of: `https://unpkg.com/@jspsych/PLUGIN-NAME@VERSION`, making it very easy to construct the URLs using only the plugin name and version.

This way, each study would only be loading the plugins they need, and users would have access to the full set of jsPsych plugins. 

This would require the following changes:
- [ ] Update the jsPsych study edit form with a multiple-select checklist, populated with the full set of plugins and associated versions
- [ ] Add the plugin list to the study model/context and make it available to the jsPsych study template
- [ ] Add a for loop to the study template that loops through plugin list and adds the `script` elements
- [ ] Possibly store the integrity SHAs that are associated with each plugin so that these can be added to the `script` elements

## Questions

How should we handle plugin versions? 
- We could simplify things by leaving out the plugin version, which would mean always linking to the 'latest' version, but this means that a researcher's code could change unexpectedly whenever jsPsych puts out releases.
- Offer a static set of versions (e.g. current latest version for each plugin) and update these manually as new jsPsych releases come out.
- Add an associated 'version' option in the plugin form question where users can specify a valid version. This would be more complicated to set up, because the study table would need to store both the plugin name and version (rather than just name), plus we would either need to get rid of the integrity SHAs or store them for all versions. But it would mean that researchers could use different plugin versions and optionally update whenever a jsPsych release comes out.

becky-gilbert:
Fixes #1484

This PR implements the dynamic loading of jsPsych plugins from the database. The plugins are stored in a table with metadata: name, URL, integrity SHA, category, file type, whether or not it should be auto-loaded, whether or not it should be displayed on the jsPsych study edit form, documentation URL. This metadata allows us to auto-load the packages that are always necessary or that we decide to add for convenience, and to display the optional plugins according to category (CHS-jsPsych, core jsPsych, jsPsych-contrib) along with their version numbers and documentation links.

This change allows us to more easily add/update the available jsPsych plugins for researchers through the admin interface, without requiring any changes to the site code (which requires build/deploy and downtime). Plugins/extensions can be added/modified/removed at any time, which makes it much easier to add newly-requested plugins and get bug fixes out quickly.

This change also addresses the slow initial load issue as studies can opt-in to plugins, which means they're only loading the packages that they actually use.

The updated jsPsych study edit page now has two sections underneath the code editor:
- Automatically-loaded files: These are the files (packages/plugins) that are always available
- More plugins and extensions: These are optional plugins/extensions that researchers can select if they want to use them.

These sections are automatically collapsed, so as not to be overwhelming, but can be expanded to see the details/options in each section.

For backwards-compatibility, existing jsPsych studies will automatically include all optional plugins/extensions. This ensures that existing studies don't break because they've lost access to an optional plugin/extension that was automatically available before. But new studies will not have any of the optional packages selected by default.

## Screenshots

jsPsych study edit page

Screenshot 2026-07-15 at 2 58 55 PM

jsPsych study edit page - expanded "Automatically-Loaded Files" section

Screenshot 2026-07-15 at 2 59 10 PM

jsPsych study edit page - expanded "Select More Plugins and Extensions" section

Screenshot 2026-07-15 at 2 59 34 PM


mekline:
The organization and hiding of options all looks great! Just some potential changes to the UI structure, which I think/hope only affects html tags:  

- Add an additional header, "jsPsych Options" which is the same weight/hierarchy as "jsPsych Experiment Code"

- Naming issue in general. The new text references files, plugins, extensions, libraries, and packages.  I'm going to just say THINGS for the rest but can you pick just one (max 2) to use throughout here?  
  - I tend to think that "files" is confusing (people may wonder how to add their stimulus jpgs etc. to the list).  
  - "Plugins/Extensions" is my best guess at the idiomatic jsPsych so we can go with that if there's not a better umbrella term or it's confusing to just say "plugins"! 
 
Overall proposed layout and new text: 

# jsPsych Experiment Code

(Big box of code) 

# jsPsych Options

Studies created with jsPsych rely on some underlying THINGS which have to be loaded in order for your study to work properly. Some of these are needed for almost every study, but in general you want to minimize how many you include, because they increase website loading times. 

If you aren't sure yet which ones you'll need, leave these options alone for now. 

If you have a THING you want to use that you don't see listed below, please [standard contact method used in this part of the site?]. We can probably add it for you!

### See Automatically-Loaded THINGS >

### Select Optional THINGS >

(those are smaller headers, probably left justified if I remember what the website looks like? And the same drop-down triangle & contents)

mekline (changes requested):
Formatting/wording changes requested!

becky-gilbert:
Thanks for the feedback @mekline! I see what you mean about the formatting and language. 

I know that "plugins/extensions" is clunky, but there's an important distinction from the researcher's point-of-view and the terms are not used interchangeably in jsPsych docs, so I decided to stick with that. We could also use "packages" or "files" for everything, but I think that's too vague, including for seasoned jsPsych users.

How's this version?

Both collapsed: 

Screenshot 2026-07-30 at 3 15 43 PM

Auto-loaded file section expanded:

Screenshot 2026-07-30 at 3 15 57 PM

Optional file section expanded:

Screenshot 2026-07-30 at 3 16 12 PM

mekline (approved):
All great, thank you!

